PR #7, opened 2023-03-15
rename mpm and moveos-stdlib to mos and mos-stdlib

jolestar:


lshoo:
Add const value and fn for get `0x1` address

lshoo (approved):
LGTM

PR #42, opened 2023-03-30
[moveos] Rename mos to moveos, mos-framework to rooch-framework

jolestar:
* Rename `mos` to `moveos`
* Rename `mos-framework` to `rooch-framework`


moveos/moveos-stdlib/rooch-framework/sources/account.move
```diff
@@ -1,4 +1,4 @@
-module mos_framework::account{
+module rooch_framework::account{
```
baichuan3, line 1: after rename the new directory structure is:

-moveos-stdlib
| --moveos-stdlib
| --rooch-stdlib

The positioning of moveos is an independent library. Maybe it should be that rooch-framework depends on moveos-stdlib. Now moveos-stdlib directly includes rooch-framework, and the dependency relationship is a bit strange.

Think of two optimization methods, according to the existing directory structure, rename to
-framework
|--moveos-stdlib
|--rooch-framework

Or the rooch-framework is raised one level up, and it is at the same level as moveos.

jolestar: I will move out the `rooch-framework` from `moveos`, but need to refactor the framework build.

moveos/moveos-stdlib/src/natives/mod.rs
```diff
@@ -1,25 +1,25 @@
...
     table_extension: move_table_extension::GasParameters,
-    type_info: mos_stdlib::type_info::GasParameters,
-    rlp: mos_stdlib::rlp::GasParameters,
-    account: mos_framework::account::GasParameters,
-    bcd: mos_stdlib::bcd::GasParameters,
-    tx_context: mos_stdlib::tx_context::GasParameters,
-    object: mos_stdlib::object::GasParameters,
+    type_info: moveos_stdlib::type_info::GasParameters,
+    rlp: moveos_stdlib::rlp::GasParameters,
+    account: rooch_framework::account::GasParameters,
```
lshoo, line 19: If moveos split to an independent library，the `GasParameters` in moveos can‘t depend rooch, do we need to move `GasParamerters` into rooch_framework?

lshoo:
The words `moveos` and `MoveOS` in `CONTRIBUTING.md` should be replaced with `rooch` or `Rooch`.
And `moveos` in `.github/workflows/rust.yml` should be replaced with `rooch` for CI.

lshoo (approved):

Issue #21, opened 2023-03-17
[Move] Hybrid Object and Global Storage in Move


jolestar:
In core Move, global storage is part of the programming model and can be accessed through special operations, such as move_to, move_from and many more[ global storage operators](https://move-language.github.io/move/global-storage-operators.html). Both resources and modules are stored in the core Move global storage.

[Sui](https://docs.sui.io) introduce [Object model](https://docs.sui.io/learn/sui-move-diffs#object-centric-global-storage) in Move, and disable all global storage operators.

We try to hybridize the two features and make the Move language more expressive.

The developers can use global storage to save the Object index and use Objects to express Resources.

- [x] Implement a PoC Object on Core Move
- [ ] Object Reference
- [ ] Process Object input argument for a transaction
- [ ] Define Object ownership and lifetime.
- [ ] Write an example for the hybrid solution, and give the best practice when using Object and when using Global Storage

pause125:
What are the advantages of Object and global storage respectively?

jolestar:
> What are the advantages of Object and global storage respectively?

Give an example. We use Move to write a Blog system.

Where to save the Article? If we save the Article in the author's storage, How do we aggregate the author's articles for the list?

A solution uses the Object and Global Storage:

1. We define the Article as an Object and save the Article to the object store.
2. We reference articles by the ObjectID, make an index for every list, and save the index to the global storage.

jolestar:
## The Object in StateDB

<img width="1249" alt="moveos statedb" src="https://user-images.githubusercontent.com/77268/228173501-e837fa36-a80b-4031-95ed-811746906ffc.png">

1. Anything is an Object, including the Table.
2. A user has two tables, one for store Resources and another for store Modules.
3. Object in the global state tree, the key is ObjectID.
4. Every Table is an SMT tree. The Table Object stores the SMT's state root.
5. Normal resource struct can ref an object by the ObjectRef(TODO).

jolestar:
An idea about ResourceTable

We can only save ObjectRef in the ResourceTable, requiring the struct with the `key` ability to be an Object, like Sui. 

* `move_to<T>(signer)` = save T to object store and save a reference in ResourceTable.
* `borrow_global<T>(address)` = get the T object reference and borrow it from the object store.

jolestar:
Some ideas about ObjectRef

We need two types of ObjectRef:

* Ownership Reference: an Object only have one Ownership Reference, and represents the Object's ownership.
[...]

PR #48, opened 2023-04-09
[object] A draft implementation of box style object

jolestar:
## Box Style Move Object 

This PR is part of #21 

In the Sui Move, Object is a Move struct which has some additional requirements:

1. The struct must have the `key` ability.
2. The struct's first field must be `UID` type.

In this PR, I try to introduce a Box-style Move Object. The Object is like a Box and has an ID field. We can put a Move resource to the Box and save it to the Object Store.

```rust
struct Object<T: store> {
        id: UID,
        value: T,
}

public fun new<T: store>(ctx: &mut TxContext, value: T): Object<T>{
    let id = UID {
        id: ID { bytes: tx_context::new_object(ctx) },
    };
    Object<T>{id, value}
}

public fun borrow_value<T: store>(this: &Object<T>): &T{
    &this.value
}

public fun borrow_value_mut<T: store>(this: &mut Object<T>): &mut T{
    &mut this.value
}

public fun transfer<T: store>(obj: Object<T>, recipient: address);
```

The ObjectRef:

```rust
struct ObjectRef< phantom T: store> has store, drop{
    id:  ID,
}

public fun as_object_ref<T: store>(this: &Object<T>): ObjectRef<T>{
    ObjectRef<T>{
        id: this.id.id,
    }
}

public fun borrow<T: store>(this: &ObjectRef<T>): &Object<T>{
        &borrow_object(this.id)
}

native fun borrow_object<T: store>(id: ID): &Object<T>;

```

## Pros

1. Box-style Move Object can be an extension to the Move VM like the Table extension, which does not require modifying the bytecode verifier.

## Cons

1. Developer can not directly define an Object type. (`Article` vs `Object<Article>`).

## TODO

1. Implement the ObjectRef(this will implement in the next PR)


moveos/moveos-stdlib/moveos-stdlib/sources/object.move
```diff
@@ -141,20 +141,73 @@ module moveos_std::object {
...
+
+    public fun borrow_mut<T: store>(this: &mut Object<T>): &mut T{
+        &mut this.value
+    }
+
+    public fun borrow_by_ref<T: store>(this: &ObjectRef<T>): &T{
+        &borrow_object(this.id).value
+    }
+
+    public fun borrow_mut_by_ref<T: store>(this: &mut ObjectRef<T>): &mut T{
```
baichuan3, line 195: How to get &mut ObjectRef<T>? Is there any security problem to modify T through ObjectRef reference?

moveos/moveos-stdlib/moveos-stdlib/sources/object.move
```diff
@@ -141,20 +141,73 @@ module moveos_std::object {
...
     /// Turn the given object into a mutable shared object that everyone
     /// can access and mutate. This is irreversible, i.e. once an object
     /// is shared, it will stay shared forever.
-    public native fun share_object<T: key>(obj: T);
+    //public native fun share_object<T: store>(obj: Object<T>);
+
+    native fun transfer_internal<T: store>(obj: Object<T>, recipient: address);
+
+    // Box style object
+     struct Object<T: store> {
```
baichuan3, line 161: In both core Move and Sui Move, the [key ability] denotes a type that can appear as a key in global storage. Does box style Object need the [key ability]  , and the [store ability]?

moveos/moveos-types/src/object.rs
```diff
@@ -129,98 +132,120 @@ pub type SequenceNumber = u64;
...
+    pub fn new_move_object(value: MoveObjectData) -> Self {
         Self {
-            data: ObjectData::MoveObject(object),
+            data: ObjectData::Move(value),
             //TODO: set the owner
             owner: Owner::Immutable,
         }
     }
+
+    pub fn as_move_resource(&self, id: ObjectID) -> Result<Vec<u8>> {
```
lshoo, line 159: I think the move_resource will be a struct(type), so the funtions name`as_move_resource`  will like  `as_resource_bytes`

moveos/moveos-stdlib/moveos-stdlib/sources/object.move
jolestar: Yes, I also think about this. Maybe we need to put a field `mutable` to ObjectRef? or define two types of reference: `ObjectRef` and `ObjectMutRef`.

The `ObjectMutRef` can be created by the mut Object.

moveos/moveos-stdlib/moveos-stdlib/sources/object.move
jolestar: No. `Object` has no `key` or `store`, so the developer can not put Object to global storage(move_to), or embed to another struct. only the `transfer` function can consume the Object.

moveos/moveos-types/src/object.rs
jolestar: This is a draft method. I try to find a way to map the Move Object struct to Rust Object. Because the Move Object has a generic type parameter `T`, it is hard to implement in Rust.

pause125: Why need this map( Move object to Rust object)?

jolestar: We need parse the Move object in Rust code.

jolestar: I rename this function to `as_object_argument`

jolestar: > I think there's no need to parse the whole Object. Maybe parse the Object::id field is enough?

Yes.

Rust Move Object
```
struct MoveObject{
  id: ObjectID,
  value: Vec<u8>,
}
```

Move Object
```
struct Object<T>{
   id: address,
   value: T,
}
```

I can not find a simple method to make rust `bcs::to_bytes(&move_object)` == move `bcs::to_bytes(&object)`

jolestar:
I remove the ObjectRef, please help review @baichuan3 @pause125

baichuan3:
> I remove the ObjectRef, please help review @baichuan3 @pause125

Is object ref implemented later, or does this feature no longer exist?

jolestar:
> > I remove the ObjectRef, please help review @baichuan3 @pause125
> 
> Is object ref implemented later, or does this feature no longer exist?

The ObjectRef design #51

moveos/moveos-stdlib/src/natives/moveos_stdlib/object.rs
```diff
@@ -153,6 +153,23 @@ pub fn native_share_object(
     Ok(NativeResult::ok(cost, smallvec![]))
 }
 
+// ObjectBox keep same layout with Object struct in Move
+pub(crate) struct ObjectBox {
+    pub id: AccountAddress,
```
baichuan3, line 158: If the account address is used directly for the object id, it means that if the recipient address passed in when transferring is the object id, because it cannot be verified, the owner will become the object id and the object will never be accessible again.

moveos/moveos-stdlib/src/natives/moveos_stdlib/any_table/mod.rs
```diff
@@ -0,0 +1,665 @@
+// Copyright (c) RoochNetwork
+// Copyright (c) The Diem Core Contributors
+// Copyright (c) The Move Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+/// A native Table implementation for save any type of value.
+/// Refactor from https://github.com/rooch-network/move/blob/5b413d009515ad1144042ded27cbe9bd702aaad7/language/extensions/move-table-extension/src/lib.rs#L4
```
baichuan3, line 7: any table is used to implement storage abstraction?

baichuan3:
wait for further implementation of object and then continuing the review

moveos/moveos-stdlib/src/natives/moveos_stdlib/any_table/mod.rs
jolestar: yes

moveos/moveos-stdlib/moveos-stdlib/sources/object.move
```diff
@@ -2,159 +2,88 @@
...
+        //immutable: bool,
     }
 
-    /// Make an `ID` from raw bytes.
-    public fun id_from_bytes(bytes: vector<u8>): ID {
-        id_from_address(address::from_bytes(bytes))
+    #[private_generic(T)]
+    /// Create a new object, the object is owned by `owner`
+    /// The private generic is indicate the T should be defined in the same module as the caller. This is ensured by the verifier.
+    public fun new<T: key>(ctx: &mut TxContext, owner: address, value: T): Object<T>{
```
baichuan3, line 30: When the address corresponding to the signer in the owner and tx_context is inconsistent, it is equivalent to implicitly operating the transfer after creating the object, and the owner passes the parameter to become the owner of the object?

moveos/moveos-stdlib/moveos-stdlib/sources/object.move
```diff
@@ -2,159 +2,88 @@
...
+        (id, value, owner)
     }
 
-    /// Get the raw bytes for the underlying `ID` of `obj`
-    public fun id_bytes<T: key>(obj: &T): vector<u8> {
-        bcs::to_bytes(&borrow_uid(obj).id)
+    
+    #[private_generic(T)]
+    /// Add object to object store
+    public fun add<T: key>(this: &mut ObjectStore, obj: Object<T>) {
```
baichuan3, line 81: why T need ability `key`? Does object stroage is independent of global storage ?

moveos/moveos-stdlib/moveos-stdlib/sources/object.move
jolestar: The object owner can be a resource account, so let the caller decide who is owner.

moveos/moveos-stdlib/moveos-stdlib/sources/object.move
jolestar: I want to reuse the `key` ability to distinguish the object type and other types.

In Sui object, the object must have the `key` ability, and if have both `key` and `store`, the object can be public transfer.

baichuan3 (approved):

moveos/moveos-stdlib/src/natives/moveos_stdlib/object.rs
jolestar: It is like transferring Token to a no exists account.

jolestar:
1. implements raw_table (formerly any_table)
2. implements Table, TypeTable, ObjectStorage based on raw_table
3. Implemented AccountStorage based on TypeTable, Table and ObjectStorage, and implemented alternative methods for instructions such as move_to|move_from|borrow_global in AccountStorage.
4. Unify the global storage command and table with raw_table.

baichuan3:
LGTM

jolestar:
TODO:

1. Do some cleanup
2. Disable move_to|move_from|borrow_global or use same GlobalValue with table.

Issue #39, opened 2023-03-30
System architecture


jolestar:
Design components and how the components communicate with each other.

- [x] #36 
- [x] #11 
- [x] #10

PR #49, opened 2023-04-10
[moveos-server] MoveOS server as a standalone service

lshoo:
## Description
Moveos server is a standalone service based on [MoveOS](https://github.com/rooch-network/rooch/blob/moveos_server/moveos/moveos-server/moveos), which includes move vm, statedb, moveos framework, etc.
Moveos server is also standard MoveVM runtime, you can publish the Move module to Moveos server, and call the entry functions(execute transaction) on Moveos server .

1. Moveos server listens to and receives requests from the client through the grpc service, 
2. and then delegates the requests to the proxy to process, the proxy including moveos(state and runtime) and actor (executor), 
3. and then the actor performs the specific tasks and returns the results to the client. 
If parallel processing is required, the actor can provide good parallelism.

Moveos server provides a default configuration file in the current directory, including ip and port, 
or you  can specify a different configuration file if you need to use a different configuration.

### TODO:
1. moveos server receives publish module and handle.
2. moveos server receives entry function from cli and execute transaction.
3. publish and entry function protocol between cli and server.

## Build and run
1. Build the crate, run `cargo build -p moveos-server` for standalone, or `cargo build` for (rooch) 

2. Start the MoveOS server, run `moveos-server start` with default config, or `ROOCH_CONFIG=./rooch.yml ./target/debug/moveos-server start` if you can specific the config path or use the default value, can print the text like Listening on 0.0.0.0:50051.
   or run `rooch server start` with default config, or `ROOCH_CONFIG=./rooch.yml ./target/debug/rooch server start`


jolestar (approved):
We can rush code faster and do refactor after the first milestone. I think we should change the `moveos-server` to `rooch-server`, because the `server` need to dependency rooch's crates. But we can make this TBD.

moveos/moveos-server/protos/server.proto
```diff
@@ -0,0 +1,438 @@
+
+// Copyright (c) RoochNetwork
+// SPDX-License-Identifier: Apache-2.0
+
+syntax = "proto3";
+package moveoss;
+
+import "google/protobuf/timestamp.proto";
+
+message Transaction {
```
baichuan3, line 10: Is decided to use protobuf for serialization and deserialization?

lshoo: yes，the first version of server and cli uses grpc, so choose proto, and later versions can be updated as needed

baichuan3 (approved):

lshoo:
link #39

PR #57, opened 2023-04-22
[W.I.P] add publish module and run function in both client and server

pause125:
A  very primitive implementation for:

1. Publish modules to MoveOS;
2. Call contracts function.


Cargo.toml
```diff
@@ -39,6 +41,8 @@ moveos-stdlib = { path = "moveos/moveos-stdlib" }
 moveos = { path = "moveos/moveos" }
 moveos-cli = { path = "moveos/moveos-cli" }
 moveos-server = { path = "moveos/moveos-server" }
+moveos-client = { path = "moveos/moveos-client" }
```
lshoo, line 44: `moveos-client` not found

pause125: fixed

jolestar (approved):
Great!

PR #59, opened 2023-04-26
[object] Fix named table id bug and enable more test

jolestar:
Part of #21 , Continue #48 

1. Use the NamedTableID to ensure the resource and module table id are determined. 
2. Fix some bugs, and enable the account/basic test.

### DIFF
--- a/moveos/moveos-stdlib/moveos-stdlib/sources/account_storage.move
+++ b/moveos/moveos-stdlib/moveos-stdlib/sources/account_storage.move
@@ -5,11 +5,13 @@ module moveos_std::account_storage {
 
     use std::string::String;
     use std::signer;
+    use std::bcs;
     use moveos_std::type_table::{Self, TypeTable};
     use moveos_std::table::{Self, Table};
     use moveos_std::object::{Self, Object};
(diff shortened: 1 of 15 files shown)


moveos/moveos-stdlib/moveos-stdlib/sources/account_storage.move
```diff
@@ -5,30 +5,39 @@ module moveos_std::account_storage {
...
+    const NamedTableResource: u64 = 0;
+    const NamedTableModule: u64 = 1;
+
     struct AccountStorage has key {
         resources: TypeTable,
         modules: Table<String, vector<u8>>,
     }
 
+    //Ensure the NamedTableID generate use same method with Rust code
+    fun named_table_id(account: address, table_type: u64): address{
```
baichuan3, line 31: Is the account storage contract transparent to application developers? The same as same table_type.

baichuan3 (approved):

moveos/moveos-stdlib/moveos-stdlib/sources/account_storage.move
jolestar: yes, it is transparent

PR #68, opened 2023-05-04
implement several account api and cli 

baichuan3:
This PR contains the following implementations:

1, merge jsonrpc implementation for jsonrpsee RPC framework
2, support register jsonrpc methods for multi modules
3, implements several account api and cli


jolestar:
@wow-sven @baichuan3  We need to think about how to keep the account. 

1. Implemention the account manager on the node server side, cli call the account API to manage the account, but the account API only be accessed through the local IP by default. Like ethereum and starcoin, this approach is suited for the P2P node network.
2. Implemention the account manager on the cli client side, and maintain the account in a local config file. 

I prefer the first approach because the Rooch node needs an account do some operation.

crates/rooch/src/account/mod.rs
jolestar, line 1: This file should be at `src/commands/account/mod.rs` ?

jolestar (approved):

baichuan3:
> @wow-sven @baichuan3 We need to think about how to keep the account.
> 
> 1. Implemention the account manager on the node server side, cli call the account API to manage the account, but the account API only be accessed through the local IP by default. Like ethereum and starcoin, this approach is suited for the P2P node network.
> 2. Implemention the account manager on the cli client side, and maintain the account in a local config file.
> 
> I prefer the first approach because the Rooch node needs an account do some operation.

The current implementation is based on the first approach, but it needs to add local IP restrictions

crates/rooch/src/account/mod.rs
baichuan3: fixed

PR #70, opened 2023-05-04
[docs] Add document about storage abstraction

jolestar:
The design document for Storage Abstraction 

* RawTable
* Table
* TypeTable
* Object
* ObjectStorage
* AccountStorage

TBD: 
1. Whether to provide a method that allows Owners to directly operate on Objects, similar to Sui's `public_transfer`. #80 
2. Whether to completely abandon the global storage instructions in Move, or to simultaneously provide both methods of operation.


baichuan3 (approved):

docs/design/storage_abstraction.md
```diff
@@ -0,0 +1,275 @@
...
+### Rooch Object
+
+Objects in Rooch adopt a Box mode where they are equivalent to a unique ID box that can encapsulate a type `T` to create an Object.
+
+```move
+module moveos_std::object{
+    struct Object<T: key>{
+        id: ObjectID,
+        owner: address,
+        value: T,
```
Mine77, line 142: So, what could be the value here? Primitive types like integers and addresses? Can it be another object?

docs/design/storage_abstraction.md
```diff
@@ -0,0 +1,275 @@
...
+
+Currently, all methods in RawTable are `friend` methods, and developers cannot directly operate on them. They need to operate through the upper-level data structures that encapsulate RawTable, such as `Table<K,V>`, `TypeTable`, and `ObjectStorage`.
+
+### Rooch Object
+
+Objects in Rooch adopt a Box mode where they are equivalent to a unique ID box that can encapsulate a type `T` to create an Object.
+
+```move
+module moveos_std::object{
+    struct Object<T: key>{
```
Mine77, line 139: Do we need Versioned ID?
https://docs.sui.io/learn/objects#referring-to-objects

docs/design/storage_abstraction.md
```diff
@@ -0,0 +1,275 @@
...
+    public fun exists_module(ctx: &StorageContext, account: address, name: String): bool;
+
+    /// Publish modules to the account's storage
+    public fun publish_modules(ctx: &mut StorageContext, account: &signer, modules: vector<vector<u8>>);
+}
+```
+
+The above methods can replace the global storage instructions in Move, so that all storage-related operations in Move are ultimately unified to Operations on Tables. 
+
+At the same time,this approach also solves a reference problem in Move whereby developers cannot return a reference borrowed through the `borrow_global<T>` instruction in their methods, since this reference is created out of thin air and cannot be returned as a method return value. However, if the reference is borrowed through `StorageContext`, then that goal can be achieved, and the lifetime of the reference is bound to the lifetime of `StorageContext`. This eliminates the need for developers to manually track reference lifetimes and ensures the safety of their usage in the Move program.
```
Mine77, line 248: So does it mean I don't need to use the `acquires` annotation anymore? and is it safe?
https://move-language.github.io/move/global-storage-operators.html#reference-safety-for-global-resources

docs/design/storage_abstraction.md
```diff
@@ -0,0 +1,275 @@
...
+
+In addition, AccountStorage also provides module-related methods, making it easy to deploy Move contracts in Move and allowing developers to use contracts to define contract deployment rules, such as upgrading the contract through DAO governance.
```
Mine77, line 250: what's the method for updating a module? (i.e. what happens if I publish a module with an existed name) and how exactly do I define the deployment rules?
Maybe adding an example here?

docs/design/storage_abstraction.md
```diff
@@ -0,0 +1,275 @@
...
+
+> TBD: Whether to completely abandon the global storage instructions in Move, or to simultaneously provide both methods of operation, needs further study to decide.
```
Mine77, line 252: I can imagine it will be very confusing when having two sets of methods.

docs/design/storage_abstraction.md
```diff
@@ -0,0 +1,275 @@
...
+    /// Add object to object store
+    public fun add<T: key>(this: &mut ObjectStorage, obj: Object<T>);
+
+    public fun contains<T: key>(this: &mut ObjectStorage, object_id: ObjectID): bool;
+}
+```
+
+The above methods are ensured to only be called directly by Modules containing `T` through [private_generics](https://github.com/rooch-network/rooch/issues/64), and the safety model follow to the constraints of the Move global storage instructions.
+
+>TBD: Whether to provide a method that allows Owners to directly operate on Objects, similar to Sui's `public_transfer`, needs further study.
```
Mine77, line 179: I think yes cause on of the benefit of objects is that I can use a native function to fastly and inexpensively transfers it.
If I have objects and I still need to implement a rather complicated methods to transfer it, it might cost me more gas fee?

docs/design/storage_abstraction.md
jolestar: `T` must have `key` ability, so it can not be Primitive types and also can not be another Object.

docs/design/storage_abstraction.md
jolestar: We do not need a Versioned ID or Sequencer Number to Object. The Object can not be transaction arguments, and the Object can only get by ObjectID in the Move contract, so it is always the latest version.

docs/design/storage_abstraction.md
jolestar: It is a good question. We don't need the `acquires`, but the safety need to be re-evaluated. I record an issue #82

docs/design/storage_abstraction.md
Mine77: so can an object own another object?

docs/design/storage_abstraction.md
```diff
@@ -0,0 +1,275 @@
...
+
+It mainly provides the following API:
+
+```
+module moveos_std::account_storage{
+
+    #[private_generics(T)]
+    /// Move a resource to the account's storage
+    /// This function equates to `move_to<T>(&signer, resource)` instruction in Move
+    public fun global_move_to<T: key>(ctx: &mut StorageContext, account: &signer, resource: T);
```
Mine77, line 226: I fee like there are 2 sets of object operation methods? for example if I wanna create a resource under my account, I have 2 options right?
 1. `global_move_to<T: key>(ctx: &mut StorageContext, account: &signer, resource: T)` 
 2. `add<T: key>(this: &mut ObjectStorage, obj: Object<T>)` 

As a dapp dev, will I be able to use both of them? if so, which one should I use under what circumstances.

docs/design/storage_abstraction.md
```diff
@@ -0,0 +1,275 @@
...
+
+In addition, AccountStorage also provides module-related methods, making it easy to deploy Move contracts in Move and allowing developers to use contracts to define contract deployment rules, such as upgrading the contract through DAO governance.
+
+> TBD: Whether to completely abandon the global storage instructions in Move, or to simultaneously provide both methods of operation, needs further study to decide.
+
+### Unified State AccessPath API
+
+Since Rooch's StateDB is a nested SMT, we can provide a unified [access path API](https://github.com/rooch-network/rooch/issues/58).
+
+- `/key`: Access the first-level SMT, where `key` is an `ObjectID`.
```
Mine77, line 258: Which method should I use to query the ObjectID?

docs/design/storage_abstraction.md
```diff
@@ -0,0 +1,275 @@
...
+
+    public fun contains<T: key>(this: &mut ObjectStorage, object_id: ObjectID): bool;
+}
+```
+
+The above methods are ensured to only be called directly by Modules containing `T` through [private_generics](https://github.com/rooch-network/rooch/issues/64), and the safety model follow to the constraints of the Move global storage instructions.
+
+>TBD: Whether to provide a method that allows Owners to directly operate on Objects, similar to Sui's `public_transfer`, needs further study.
+
+### TypeTable
```
Mine77, line 181: In the diagram it is called `TableInfo`? Is this the same thing or a different type?

docs/design/storage_abstraction.md
jolestar: I provide a `publish_modules` function, but it is incomplete. I write two issues to track this #83 #84

docs/design/storage_abstraction.md
jolestar: Yes, but the tradeoff is to keep compatibility.

Mine77:
I think specifically for this design document, the first priority is to make sure it describes the details precisely, instead of optimizing it for dissemination. 
If we put an intro of Storage Abstraction on our website, we will write another shorter piece without introducing too many details, and link to this doc.

docs/design/storage_abstraction.md
jolestar: This design focuses on the Object's expressive ability rather than execution efficiency, So make it TBD.

docs/design/storage_abstraction.md
jolestar: ObjectStorage is for ObjectID -> Object store, and AccountStorage is for StructType -> Struct store.

docs/design/storage_abstraction.md
jolestar: Object<TableInfo> is the table object for storing the SMT root. `TypeTable` and `Table` have TableInfo.

docs/design/storage_abstraction.md
jolestar: No, we have ResourceAccount to do the same thing. And a struct can put an ObjectStorage as its field, so the Struct can hold many Objects.

```move
struct ChildrenObjects has key {
   children: ObjectStore,
}
```

jolestar:
I merge this PR first. If anyone has more comments, I will fix them in the following PR.

wow-sven (approved):

PR #78, opened 2023-05-05
add a basic_object_example

wubuku:
A example using box style object.


wow-sven:
Can you provide a workflow？

PR #81, opened 2023-05-06
Improving owner model for box style objects

wubuku:
This PR can be used for discussion purposes only, it does not necessarily mean it should not be merged.

### DIFF
--- a/moveos/moveos-stdlib/moveos-stdlib/sources/object.move
+++ b/moveos/moveos-stdlib/moveos-stdlib/sources/object.move
@@ -3,9 +3,13 @@
 /// Move object identifiers
 module moveos_std::object {
     use moveos_std::tx_context::{Self, TxContext};
+    use std::option::{Self, Option};
+    use moveos_std::object_owner::{Self, Owner};
 
     friend moveos_std::account_storage;
-    
+    friend moveos_std::object_storage;
+
+
     /// Invalid access of object, the object is not owned by the signer or the object is not shared or immutable
     const EInvalidAccess: u64 = 0;
    
@@ -22,7 +26,8 @@ module moveos_std::object {
     /// The object can not be copied, droped, only can be consumed by `add`
     struct Object<T> {
         id: ObjectID,
-        owner: address,
+        //owner: address,
+        owner: Option<Owner>,
         //TODO define shared and immutable
         //shared: bool,
         //immutable: bool,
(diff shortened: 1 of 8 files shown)


moveos/moveos-stdlib/moveos-stdlib/sources/object_storage.move
```diff
@@ -33,13 +34,31 @@ module moveos_std::object_storage {
...
 
+    #[private_generic(T)]
+    public fun transfer<T: key>(this: &mut ObjectStorage, obj: Object<T>, recipient: address) {
+        object::set_address_owner(&mut obj, recipient);
+        add(this, obj);
+    }
+
+    #[private_generic(T)]
+    public fun freeze_object<T: key>(this: &mut ObjectStorage, obj: Object<T>) {
+        add(this, obj);
```
baichuan3, line 54: Is the operation of setting the Owner flag missing here? And also in the below fun ```share_object```.

moveos/moveos-stdlib/moveos-stdlib/sources/object.move
```diff
@@ -22,24 +26,43 @@ module moveos_std::object {
...
+        //owner: address,
+        owner: Option<Owner>,
         //TODO define shared and immutable
         //shared: bool,
         //immutable: bool,
         // The value must be the last field
         value: T,
     }
 
+    public(friend) fun set_address_owner<T>(obj: &mut Object<T>, owner: address) {
```
baichuan3, line 38: verification is required when setting owner, for example, transfer is not allowed after freeze and share

moveos/moveos-stdlib/moveos-stdlib/sources/object_storage.move
```diff
@@ -33,13 +34,31 @@ module moveos_std::object_storage {
...
     #[private_generic(T)]
     /// Add object to object store
-    public fun add<T: key>(this: &mut ObjectStorage, obj: Object<T>) {
+    public(friend) fun add<T: key>(this: &mut ObjectStorage, obj: Object<T>) {
+        assert!(option::is_some(&object::owner(&obj)), EObjectOwnerIsEmpty);
         raw_table::add<ObjectID, Object<T>>(*&this.handle, object::id(&obj), obj);
     }
 
+    #[private_generic(T)]
+    public fun transfer<T: key>(this: &mut ObjectStorage, obj: Object<T>, recipient: address) {
```
jolestar, line 47: I think the `transfer`,`freeze_object`,`share_object` should be a function in object module

```move
public fun transfer<T:key>(this: &mut Object<T>, recipient: address)
```

jolestar:
I didn't think clearly about the application scenario of the `shared` and `immutable` Object.

In the previous version, I put two fields `shared`,`immutable` in the Object, and then commented out the fields.

PR #98, opened 2023-05-08
[types] Define rooch-types and use H256 to replace HashValue

jolestar:
1. Define rooch-types and draft the rooch transaction types.
2. Migrate the moveos::types to moveos-types.
3. Use H256 to replace HashValue.

TODO
1. How to support multi chain transaction
2. Transaction validation


crates/rooch-types/src/transaction/mod.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) RoochNetwork
+// SPDX-License-Identifier: Apache-2.0
+
+use ethereum::{EthereumSignature, EthereumTransactionData};
+
+pub mod ethereum;
+
+pub enum TransactionType {
+    Ethereum,
+    Move,
```
baichuan3, line 10: Need to distinguish transactions from different Move chains?

moveos/moveos-types/src/h256.rs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) RoochNetwork
+// SPDX-License-Identifier: Apache-2.0
+
+pub use ethereum_types::H256;
+use tiny_keccak::{Hasher, Sha3};
+
+pub const LENGTH: usize = 32;
```
baichuan3, line 7: The difference of this method is hash[.32], the old method is hash[..], is it for compatibility with the Ethereum chain?

jolestar: Old HashValue also hash[.32].

crates/rooch-types/src/transaction/mod.rs
jolestar: Just use an example for support multichain transactions, it is not finished yet.

baichuan3 (approved):

jolestar:
part of #72

Issue #66, opened 2023-05-04
[Framework] refactor the framework based on Move Object and Storage Abstraction


baichuan3:
Refactor the framework based on Move Object and Storage Abstraction, including instructions such as move_to, move_from, borrow_global, borrow_global_mut and exist

baichuan3:
close as complete https://github.com/rooch-network/rooch/pull/102

PR #102, opened 2023-05-09
Refactor framework by using new storage context instruction

baichuan3:
this pr is an implementation of https://github.com/rooch-network/rooch/issues/66

Include several changes:
1, Refactor the framework based on new Storage Abstraction, including instructions such as move_to, move_from, borrow_global, borrow_global_mut and exist
2, Provide native implementation, you can discard any Move Resource, even if the Resource has no drop ability, it is only used for test scenarios
3, Fixed a bug when raw table is in drop or destroy


baichuan3:
The unit test encountered an exception thrown by the raw table. After troubleshooting, it was determined that the table handler was the same when creating and borrowing. It feels like a raw table bug was triggered. @jolestar have a look?

```
Failures in 0x1::chain_id:

┌── test_get ──────
│ error[E11001]: test failure
│     ┌─ ./../moveos-stdlib/sources/raw_table.move:105:16
│     │
│ 105 │     native fun borrow_box<K: copy + drop, V, B>(table_handle: address, key: K): &Box<V>;
│     │                ^^^^^^^^^^
│     │                │
│     │                Test was not expected to error, but it aborted with code 25863 originating in the module 0000000000000000000000000000000000000000000000000000000000000001::raw_table rooted here
│     │                In this function in 0x1::raw_table
│
│
└──────────────────
```

moveos/moveos-stdlib/moveos-stdlib/sources/storage_context.move
```diff
@@ -33,4 +40,20 @@ module moveos_std::storage_context {
         &mut this.object_storage
     }
 
+    #[test_only]
+    /// Create a StorageContext for unit test
+    public fun test_context(sender: address): StorageContext {
+        let tx_context = tx_context::test_context(sender);
+        let object_storage = object_storage::new(&mut tx_context);
```
jolestar, line 47: The global object storage's table handle should be 0x0. We can export a friend function in object_storage, like `new_with_id` in TypeTable.

moveos/moveos-stdlib/moveos-stdlib/sources/table.move
```diff
@@ -110,7 +110,9 @@ module moveos_std::table {
         upsert(&mut t, key, 23);
         assert!(*borrow(&t, key) == 23, error_code);
 
-        move_to(&account, TableHolder { t });
+        // drop_unchecked(t)
```
jolestar, line 113: The `drop_unchecked` can not drop the table?

moveos/moveos-stdlib/moveos-stdlib/sources/raw_table.move
```diff
@@ -110,7 +110,7 @@ module moveos_std::raw_table {
 
     native fun remove_box<K: copy + drop, V, B>(table_handle: address, key: K): Box<V>;
 
-    native fun destroy_empty_box(table_handle: address);
+    native fun destroy_empty_box<K>(table_handle: address);
 
-    native fun drop_unchecked_box(table_handle: address);
+    native fun drop_unchecked_box<K>(table_handle: address);
```
jolestar, line 115: I think `destroy_empty_box` and `drop_unchecked_box` do not require the `K` Type. You need to remove the ty_args length check: `assert_eq!(ty_args.len(), 1)`

https://github.com/rooch-network/rooch/blob/5ac0b5b9b534952f20059869a3da1d1c76dea42c/moveos/moveos-stdlib/src/natives/moveos_stdlib/raw_table/mod.rs#L547-L566

baichuan3: has remove the `K` type for destroy_empty_box and drop_unchecked_box

jolestar:
The view function does not auto-resolve the StorageContext argument

https://github.com/rooch-network/rooch/blob/5ac0b5b9b534952f20059869a3da1d1c76dea42c/moveos/moveos/src/tests/mod.rs#L35-L57

You can disable the test and record an issue.

moveos/moveos-stdlib/moveos-stdlib/sources/storage_context.move
```diff
@@ -33,4 +40,21 @@ module moveos_std::storage_context {
         &mut this.object_storage
     }
 
+    #[test_only]
+    /// Create a StorageContext and AccountStorage for unit test
+    public fun new_test_context(sender: address): StorageContext {
+        let tx_context = tx_context::test_context(sender);
+        // let object_storage = object_storage::new(&mut tx_context);
+        let object_storage = object_storage::new_with_id(@moveos_std_package);
```
jolestar, line 48: This address "0x0" is not `moveos_std_package`, I think we should hard code it, and named it `global_object_storage_handle`.

jolestar (approved):

PR #103, opened 2023-05-09
Feature/account

wow-sven:
There's still work to be done,
Partial create is currently implemented
image


moveos/moveos-types/src/rooch_serde.rs
jolestar, line 1: If this file is at moveos-types, it should be `moveos_serde`

jolestar (approved):

PR #115, opened 2023-05-12
fix dependency order when package publish

baichuan3:
An error will be reported when deploying multiple contracts, resolve https://github.com/rooch-network/rooch/issues/116

`
VMError with status LINKER_ERROR at location UNDEFINED and message Cannot find ModuleId { address: 0000000000000000000000000000000000000000000000000000000000000123, name: Identifier("something_do_logic") } in data cache

`

Ω

### DIFF
--- a/moveos/moveos/src/vm/dependency_order.rs
+++ b/moveos/moveos/src/vm/dependency_order.rs
@@ -1,0 +1,71 @@
+// Copyright (c) The Diem Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use anyhow::{bail, ensure, Result};
+use move_binary_format::{access::ModuleAccess, file_format::CompiledModule};
+use petgraph::graphmap::DiGraphMap;
+use std::collections::BTreeMap;
+
+/// Given an array of compiled modules, sort the modules by their dependency orders.
+pub fn sort_by_dependency_order<'a>(
+    modules: impl IntoIterator<Item = &'a CompiledModule>,
+) -> Result<Vec<CompiledModule>> {
+    let graph = DependencyGraph::new(modules)?;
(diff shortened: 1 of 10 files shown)


moveos/moveos/src/vm/dependency_order.rs
jolestar, line 1: I think we can directly use the Move's `dependency_order.rs`, you can see the Framework build.

jolestar: The framework build package all dependency into one package. For DApp, We just package the project's module, call `package.root_modules_map()`

https://github.com/rooch-network/rooch/blob/78a263912d0def756a27c4567d7b70c6a979f98a/moveos/moveos-stdlib/src/lib.rs#L81-L89

baichuan3: yes, use package the project's module, call package.root_modules_map()

you can use this branch for test ,  https://github.com/rooch-network/rooch/tree/bugfix_publish_order_tmp

Step1:
` rooch server start`

Step2:
`rooch move publish -p examples/basic_object_example --sender-account 0x123`

then occur error,
`thread 'main' panicked at 'Missing dependency 000000000000000000000000000000000000000000000000000000000000001::object',~/.cargo /git/checkouts/move-d59e49550a684528 /5b413d0/language/tools/move-bytecode-utils/src/dependency_graph.rs:46:44`

call entry is here, https://github.com/rooch-network/rooch/blob/5b11359c4844dbe825178b843792e4f344de854a/moveos/moveos-cli/src/commands/publish.rs#L49

jolestar:
Maybe this is a bug of Move's `dependency_order.rs`. We can merge first, then try to fix it on upstream. #124

jolestar (approved):

Issue #63, opened 2023-05-04
[moveos] Provides a module initialization method


jolestar:
When the developer publishes a module, usually needs to call initializer method.

There are some solutions:

1. Starcoin: Provide an init function combined with the modules transaction.

https://github.com/starcoinorg/starcoin/blob/de940468329edb4fbc884e1c2381281c2afd7e49/vm/types/src/transaction/package.rs#L19-L25

```
pub struct Package {
    ///Package's all Module must at same address.
    #[schemars(with = "String")]
    package_address: AccountAddress,
    modules: Vec<Module>,
    init_script: Option<ScriptFunction>,
}
```

2. Sui: Provide a module initializers function spec: https://docs.sui.io/learn/sui-move-diffs#module-initializers

> As described in [Object-centric global storage](https://docs.sui.io/learn/sui-move-diffs#object-centric-global-storage), Sui Move modules are published into Sui storage. A special initializer function optionally defined in a module is executed (once) at the time of module publication by the Sui runtime for the purpose of pre-initializing module-specific data (e.g., creating singleton objects). The initializer function must have the following properties in order to be executed at publication:
> * Name init
> * Single parameter of &mut TxContext type
> * No return values
> * Private

3. Aptos

https://aptos.dev/move/move-on-aptos/modules-on-aptos

> When publishing a module for the first time, The VM will search for and execute an init_module(account: &signer) function.

jolestar:
Option 2

baichuan3:
close via https://github.com/rooch-network/rooch/pull/142

PR #142, opened 2023-05-16
provides a module initialization method

baichuan3:
reslove https://github.com/rooch-network/rooch/issues/63

since the Move runtime does not to know about new packages created,
the first deployment contract and the upgrade contract need to be handled separately

The initializer function must have the following properties in order to be executed at publication:
- Name init
- Single parameter of &mut TxContext type
- No return values
- Private & no Entry

In addition, the Ocject information is finally output through the event, and the event has not been realized yet, so the log output is added first, so that the process can be connected through the cli.


moveos/moveos/src/moveos.rs
```diff
@@ -175,6 +193,144 @@ impl MoveOS {
...
+            assert!(
+                result.return_values.is_empty(),
+                "init should not have return values"
+            )
+        }
+
+        Ok(())
+    }
+
+    fn execute_move_call_bypass_visibility<S>(
```
jolestar, line 305: this function conflict with #143 , need to resolve

moveos/moveos/src/moveos.rs
```diff
@@ -175,6 +193,144 @@ impl MoveOS {
...
+                &module_id,
+                INIT_FN_NAME,
+                vec![],
+                args,
+            )?;
+
+            assert!(
+                result.return_values.is_empty(),
+                "init should not have return values"
+            )
```
jolestar, line 299: use bail! to raise the error, assert will panic the VM.

moveos/moveos/src/moveos.rs
```diff
@@ -175,6 +193,144 @@ impl MoveOS {
...
+            !modules.is_empty(),
+            "input checker ensures package is not empty"
+        );
+
+        Ok(modules)
+    }
+
+    /// The initializer function must have the following properties in order to be executed at publication:
+    /// - Name init
+    /// - Single parameter of &mut TxContext type
```
jolestar, line 213: Is support `signer`? if not, can not call `global_move_to`

moveos/moveos/src/moveos.rs
```diff
@@ -175,6 +193,144 @@ impl MoveOS {
...
+            if !self
+                .check_module_init_permission(
+                    session,
+                    tx_context,
+                    &module_id,
+                    INIT_FN_NAME,
+                    vec![],
+                    vec![],
+                )
+                .unwrap()
```
jolestar, line 278: do not `unwrap` at here

moveos/moveos/src/moveos.rs
```diff
@@ -175,6 +193,144 @@ impl MoveOS {
...
+            module_id,
+            function_name,
+            ty_args,
+            args,
+            gas_meter,
+        )?;
+        assert!(
+            result.return_values.is_empty(),
+            "Entry function should not return values"
+        );
```
jolestar, line 330: This is not an entry function that can return values.

wow-sven (approved):

jolestar (commented):
Add a TODO issue to check the module compatibility and ensure the `init` function only call on the first deployment.

moveos/moveos/src/moveos.rs
baichuan3: `init` function doesn't support `signer`, so can not `global_move_to `

moveos/moveos/src/moveos.rs
```diff
@@ -209,12 +249,100 @@ impl MoveOS {
...
+                vec![],
+                vec![],
+            )? {
+                continue;
+            };
+
+            let function =
+                Function::new(module_id, INIT_FN_NAME_IDENTIFIER.clone(), vec![], vec![]);
+            let _result =
+                Self::execute_function_bypass_visibility(session, tx_context, gas_meter, function)?;
```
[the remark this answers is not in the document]
baichuan3: > I think it's better to break immediately once found the init function.

support multi module init

jolestar (approved):

PR #149, opened 2023-05-17
[framework] Define address_mapping and return rooch address when transaction validate

### LINKED ISSUE #106: The view function does not auto-resolve the StorageContext argument
https://github.com/rooch-network/rooch/blob/5ac0b5b9b534952f20059869a3da1d1c76dea42c/moveos/moveos/src/tests/mod.rs#L35-L57

jolestar:
1. Define `address_mapping` and return rooch address when the transaction `validate`， prepare for support the eth transaction in the future.
2. Support `&signer` as the entry function argument.
3. resolve #106 

this is part of #72


moveos/moveos-stdlib/moveos-stdlib/sources/account_storage.move
```diff
@@ -55,6 +55,12 @@ module moveos_std::account_storage {
         object_storage::contains<AccountStorage>(object_storage, object_id)
     }
 
+    public fun ensure_account_storage(ctx: &mut StorageContext, account: address) {
+        if (!exist_account_storage(ctx, account)) {
```
baichuan3, line 59: account storage will create when account create, so after account created, there is no scenario where account storage does not exist, without considering the test scenario

moveos/moveos/src/moveos.rs
```diff
@@ -371,12 +384,19 @@ impl MoveOS {
         let mut session = self.vm.new_session(&self.db);
         //TODO limit the view function max gas usage
         let mut gas_meter = UnmeteredGasMeter;
-        let result = session.execute_function_bypass_visibility(
-            module,
-            function_name,
+        //View function use a fix address and fix hash
+        let tx_context = TxContext::new(AccountAddress::ZERO, H256::zero());
```
baichuan3, line 388: using the default mock TxContext will have problems in some scenarios, such as account_storage::exist_account_storage depends on the real user address.

Therefore, the view function needs to pass the real address as a parameter.

```
    /// check if account storage eixst
    public fun exist_account_storage(ctx: &mut StorageContext, account: address): bool {
        let object_id = object::address_to_object_id(account);
        let object_storage = storage_context::object_storage_mut(ctx);
        object_storage::contains<AccountStorage>(object_storage, object_id)
    }
```

baichuan3 (approved):

PR #153, opened 2023-05-18
Refactor CLI output types

pause125:
Unify output types of each commands.

To fix: 

The output of transaction is not pretty enough like this:
```
"{\n  \"code\": \"Ok\",\n  \"message\": null,\n  \"data\": {\n    \"state_root\": \"0xe30674dab1f57e29849b74d470c76d21ce41871e13974677f8e8f8ec87e1ba28\",\n    \"status\": \"Executed\"\n  }\n}"
```
Try to figure out why and fix it.

### DIFF
--- a/crates/rooch-types/src/cli.rs
+++ b/crates/rooch-types/src/cli.rs
@@ -45,8 +45,8 @@ pub enum CliError {
     IOError(String),
     #[error("Sign message error: {0}")]
     SignMessageError(String),
-    #[error("Submit TX error: {0}")]
-    SubmitTxError(String),
+    #[error("Transaction error: {0}")]
+    TransactionError(String),
     #[error("View function error: {0}")]
     ViewFunctionError(String),
 }
--- a/crates/rooch-types/src/cli.rs
+++ b/crates/rooch-types/src/cli.rs
@@ -4,6 +4,7 @@
 use std::io;
 
 use anyhow::Result;
+use async_trait::async_trait;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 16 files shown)


crates/rooch-types/src/cli.rs
```diff
@@ -61,3 +68,17 @@ impl From<io::Error> for CliError {
...
+#[async_trait]
+pub trait CommandAction<T: Serialize + Send>: Sized + Send {
+    /// Executes the command, returning a command specific type
+    async fn execute(self) -> CliResult<T>;
+
+    /// Executes the command, and serializes it to the common JSON output type
+    async fn execute_serialized(self) -> CliResult<String> {
+        match self.execute().await {
+            Ok(result) => Ok(serde_json::to_string_pretty(&result).unwrap()),
+            Err(e) => Err(e),
```
jolestar, line 81: Make the error also a json? For convenience parse the cli output.

jolestar (approved):

crates/rooch-types/src/cli.rs
pause125: Command with error will exit with code 1. And the error is a enum type will print a single `String` output.

You mean define a `struct` type for error like:
```
struct CliError {
  code: int,
  kind: ErrorKind,
  message: String,
```

jolestar: Or wrap the whole result? like

```
{"ok": {xxx}}
{"error": "message"}
```

I merge this PR first, let this TBD.

Issue #41, opened 2023-03-30
Basic Sequencer Interfaces


jolestar:
- [ ] Ordering: User-facing Sequencer Interface (+ order witness)
- [ ] Execution: Query Interface 
- [ ] I/O: Persistence Interface

steelgeek091:
## Ordering: User-facing Sequencer Interface (+ ordering witness)

**1. SendTransaction to the sequencer and get return ordering witness response from the sequencer.**

**URL:**
```txt
/v1/sequencer/send_tx
```

**Method:**

```
POST
```

**Request:**

```rust
struct {
	sender_address: String,
	public_key: Vec<u8>,
	program: Vec<u8>,
	sequencer_number: u64,
	expiration_time: u64,
	signature: Vec<u8>
}
```

**Response:**

```rust
struct {
	transaction_hash: String,
	ordering_witness: Vec<u8>
}
```

**2. The user verifies that their transaction conforms to the order guaranteed by the sequencer.**

**URL:**
```txt
/v1/sequencer/verify_tx
```

**Method:**

```
POST
```

**Request:**

```rust
struct {
	transaction_hash: String,
	ordering_witness: Vec<u8>
}
```

**Response:**

```rust
struct {
	verify_result: Vec<u8>
}
```

jolestar:
I think we need to define the Rust trait first and design a transaction flow between components.

templexxx:
some suggestions @steelgeek091 

1. transaction -> txn
2. sequencer_signature -> txn_order_signature: means sign the txn order
3. order_witness -> txn_order_witness: same strucuture as order_signature

p.s.

In computer science, and cryptographic, using witness as proof or part of proof is common, so this word is okay. The original name "sequence proof" may conflict with "sequence" in other places.

jolestar:
Sequencer basic implementation in #158 

Transaction RPC tracked by #229 
Sequence Proof and RPC tracked by #228 (v0.2)
Transaction storage tracked by #157 

Close this issue.

PR #158, opened 2023-05-19
[rooch-server] Refactor rooch-server, define rooch executor sequencer

jolestar:
part of #41 

1. Define TransactionInfo
2. Move executor to rooch-executor crate.
3. Define rooch-sequencer crate.

Transaction flow:

```
client --- submit tx --> Rpc Service
                                 ---- validate tx ---> Executor
                                 ---- sequence tx ----> Sequencer
                                 ---- execute tx ---> Executor
client <---- response -----
```

Next:
1. Define Proposer and State commitment Chain. The State commitment Chain is a chain of TransactionInfo.


crates/rooch-sequencer/src/actor/messages.rs
```diff
@@ -0,0 +1,25 @@
...
+    pub tx: TypedTransaction,
+}
+
+#[derive(Debug)]
+pub struct TransactionSequenceResult {
+    /// The tx order
+    pub order: u128,
+    /// The tx order witness, usually it is the signature of the sequencer to prove the order
+    //TODO confirm the type.
+    pub order_witness: Authenticator,
```
templexxx, line 20: order_witness is for verifying tx & its order is correct in sequencer when someone wants to chanllege sequencer.

For normal tx request is just a signature for telling users that sequencer has confirmed that the tx has this order.

@jolestar @steelgeek091

jolestar: I put the signature here and called `order_witness`, the sequencer witness the tx order. In the challenge scenario, it should be `sequence_proof`, which proves the witness is correct.

templexxx: > I put the signature here and called `order_witness`, the sequencer witness the tx order. In the challenge scenario, it should be `sequence_proof`, which proves the witness is correct.

@jolestar 

membership/non-membership witness is proper noun for checking element existed or not in accumulator. So I think witness is not a good idea name here.

And for sequence_proof: Will it make people confused that this sequence and other sequence(it's widely used in other blockchain interface)?

jolestar: I change the `order_witness` to `tx_order_signature`.

jolestar: Upon reading numerous articles, I have observed the mix use of "witness" and "proof" which let me confused.

I ask the ChatGPT, and its answer is:
--------------------------------------------
In the accumulator scenario, the terms "membership witness" and "membership proof" are usually used to describe proving whether an element belongs to a specified accumulator. In this scenario, both terms can represent narrowing down the likelihood of "the element belongs to the accumulator".

Generally, "membership witness" refers to the proof that contains the specified element (as in Merkle tree), while "membership proof" refers to the complete proof path.

Therefore, if you only need to prove whether an element belongs to an accumulator, the term "membership witness" may be more intuitive. If you need the complete proof path, the term "membership proof" is more appropriate.

baichuan3:
The role of the current rooch-executor craet, in addition to providing the actor mode, is similar to the block executor in Sui and Aptos, and finally executes the contract transaction, right? @jolestar

jolestar:
> The role of the current rooch-executor crate, in addition to providing the actor mode, is similar to the block executor in Sui and Aptos, and finally executes the contract transaction, right? @jolestar

yes.

baichuan3 (approved):

Issue #160, opened 2023-05-22
Introduce fuzzing proptest for test struct serialize and deserialize.


pause125:

jolestar:
Finished

PR #164, opened 2023-05-22
Introduce fuzzing proptest for test struct serialize and deserialize

pause125:
part of #160

To fix:

Encountering problem when impl Arbitrary trait for FileBasedKeystore struct.
```
pub struct FileBasedKeystore {
    keys: BTreeMap<RoochAddress, RoochKeyPair>,
    path: Option<PathBuf>,
}
```
When using ` btree_map(any::<RoochAddress>(), any::<RoochKeyPair>(), 1..100)` to generate a Strategy, it shows "the trait `Arbitrary` is not implemented for `RoochAddress`" and "the trait `Arbitrary` is not implemented for `RoochKeyPair`" but they are actually implemented.

### DIFF
--- a/crates/rooch-key/src/keystore.rs
+++ b/crates/rooch-key/src/keystore.rs
@@ -6,6 +6,10 @@ use anyhow::anyhow;
 use bip32::DerivationPath;
 use bip39::{Language, Mnemonic, Seed};
 use enum_dispatch::enum_dispatch;
+// #[cfg(any(test, feature = "fuzzing"))]
+// use proptest::{collection::btree_map, prelude::*};
+// #[cfg(any(test, feature = "fuzzing"))]
+// use proptest_derive::Arbitrary;
 use rand::{rngs::StdRng, SeedableRng};
 use rooch_types::account::{
     get_key_pair_from_rng, EncodeDecodeBase64, PublicKey, RoochKeyPair, SignatureScheme,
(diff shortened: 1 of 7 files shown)


crates/rooch-types/src/transaction/authenticator.rs
```diff
@@ -627,4 +656,31 @@ mod tests {
         let auth_key2 = AuthenticationKey::multi_ed25519(&multi_pubkey);
         assert_eq!(auth_key, auth_key2);
     }
+
+    #[test]
+    fn test_temp() {
```
pause125, line 661: @jolestar Please help checking this test case. The `serialized` len is 145 instead of 65, the value of `b` is {r:0, s:0, v:0}.

jolestar: The length because serialized is json.

The `assert_eq!(a.signature, b.signature);` failed because the `v` is u8 when serialized, 256 is overflow.

jolestar: Screen Shot 2023-05-23 at 09 11 32

I have no idea why they use u64 as the v's type.

crates/rooch-key/src/keystore.rs
```diff
@@ -116,6 +120,28 @@ impl<'de> Deserialize<'de> for FileBasedKeystore {
     }
 }
 
+// #[cfg(any(test, feature = "fuzzing"))]
```
jolestar, line 123: should set the features = ["fuzzing"] on the rooch-types dependency "rooch-types = { workspace = true }"

jolestar (approved):
LGTM

jolestar:
I merge this first; @pause125 can fix the features set in the following PR.

Issue #152, opened 2023-05-17
[MoveVerifier] entry function verifier


jolestar:
Implement verifier extension for entry function.

verifier the entry function parameter type and return type

### Ref

#148

steelgeek091:
Should we put this issue and its code solution together with the resolution for issue #64?

jolestar:
It is a verifier requirement, but the entry function does not need an annotation.

steelgeek091:
In this link, https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/src/extended_checks.rs#L116, the logic for checking the entry function is exactly what we need (I think), because it also checks the parameter types for the entry function to see if they are the specified types (without checking the entry function's annotations).

The code and analysis for the annotation and metadata generation of the entry function are placed together in the above code section, https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/src/extended_checks.rs#L69. My point is, if we want to quickly solve the problem with the entry function, can't we use this method first, instead of adding a new verifier to the Move codebase, as adding a new verifier would require modifications to the Move code.

PR #172, opened 2023-05-25
Private generics attribute and other extended checks

### LINKED ISSUE #64: [Move] Implement private_generics verifier annotation
First, implement this feature on the Rooch side, and then try to push to Move upstream.

* [Introduce verifier annotation, such as private_generics](https://github.com/move-language/move/issues/1043)

steelgeek091:
According to the content of this issue, https://github.com/move-language/move/issues/1043, the `#[private_generics(T1, T2)]` function attribute feature has been implemented in the Rooch project, which checks that `T1` and `T2` are defined as types in the current module.

resolve #64 

Here is an example code:

```move
module rooch_examples::Data {
    struct Box<T: store> has key, store, drop {
        value: T
    }

    struct Data1 has key, store, drop, copy {
        v: u64,
    }

    struct Data2 has key, store, drop, copy {
        v: u64,
    }

    #[private_generics(T3, T2)]
    public fun create_box<T1: store, T2, T3>(value: T1): Box<T1> {
        Box<T1> { value: value }
    }

    public fun box_value<T: copy+store>(box: &Box<T>): T {
        *&box.value
    }

    public fun run() {
        let data = Data1{ v: 123 };
        let box_val = create_box<Data1, Data2, Box<u32>>(data);
        let _ = box_value(&box_val);
    }
}
```

In the above code, we are checking the following items:

- `private_generics` needs to define at least one type `T` and cannot be empty.
- `T2` and `T3` in `#[private_generics(T3, T2)]` must exist in the generic parameters of the `create_box` function.
- When calling the function `create_box<Data1, Data2, Box<u32>>(data)`, according to the attribute `#[private_generics(T3, T2)]` checks that `Data2` and `Box<u32>` must be defined as types in the current module.

There is still have unfinished issue that will be completed at a later time:

- entry function verifier https://github.com/rooch-network/rooch/issues/152
- initial function of a module

### DIFF
--- a/crates/rooch/src/commands/move_cli/commands/new.rs
+++ b/crates/rooch/src/commands/move_cli/commands/new.rs
@@ -22,7 +22,7 @@ pub struct New {
     pub new: new::New,
 
     /// Path of the new package to create.
-    #[clap(long = "path", short = 'p', global = true, parse(from_os_str))]
+    #[clap(long = "mpath", short = 'm', global = true, parse(from_os_str))]
     pub path: Option<PathBuf>,
 }
 
--- a/crates/rooch/src/extended_checks/metadata.rs
+++ b/crates/rooch/src/extended_checks/metadata.rs
@@ -1,0 +1,274 @@
+use move_binary_format::binary_views::BinaryIndexedView;
+use move_binary_format::file_format::{Bytecode, FunctionInstantiation, SignatureToken};
+use move_core_types::language_storage::ModuleId;
+use move_model::ast::Attribute;
+use move_model::model::{FunctionEnv, GlobalEnv, ModuleEnv};
+use serde::{Deserialize, Serialize};
+use std::collections::BTreeMap;
+
+const PRIVATE_GENERICS_ATTRIBUTE: &str = "private_generics";
+
+/// Enumeration of potentially known attributes
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct KnownAttribute {
+    kind: u8,
+    args: Vec<String>,
+}
+
+/// V1 of Aptos specific metadata attached to the metadata section of file_format.
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+pub struct RuntimeModuleMetadataV1 {
+    /// Attributes attached to structs.
+    pub struct_attributes: BTreeMap<String, Vec<KnownAttribute>>,
+
+    /// Attributes attached to functions, by definition index.
+    pub fun_attributes: BTreeMap<String, Vec<KnownAttribute>>,
+
+    /// The correspondence between private generics and their type parameters.
+    pub private_generics_indices: BTreeMap<String, Vec<usize>>,
+}
+
+impl RuntimeModuleMetadataV1 {
+    pub fn is_empty(&self) -> bool {
+        self.fun_attributes.is_empty()
+            && self.struct_attributes.is_empty()
+            && self.private_generics_indices.is_empty()
+    }
+}
+
+/// Run the extended context checker on target modules in the environment and returns a map
+/// from module to extended runtime metadata. Any errors during context checking are reported to
+/// `env`. This is invoked after general build succeeds.
+pub fn run_extended_checks(env: &GlobalEnv) -> BTreeMap<ModuleId, RuntimeModuleMetadataV1> {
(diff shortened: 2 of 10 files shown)


crates/rooch/src/extended_checks/metadata.rs
```diff
@@ -0,0 +1,280 @@
...
+
+                    if private_generics_types.is_some() {
+                        let private_generics_types_indices = private_generics_types.unwrap();
+
+                        for generic_type_index in private_generics_types_indices {
+                            let type_arg = type_arguments.get(*generic_type_index).unwrap();
+                            let (defined_in_current_module, struct_name) =
+                                is_defined_in_current_module(&view, type_arg);
+
+                            if !defined_in_current_module {
```
steelgeek091, line 207: The error should be thrown in the following way instead of panicking directly:

```
self.env.error(&fun.get_loc(), "error message");
```

crates/rooch/src/commands/move_cli/commands/new.rs
```diff
@@ -22,7 +22,7 @@ pub struct New {
     pub new: new::New,
 
     /// Path of the new package to create.
-    #[clap(long = "path", short = 'p', global = true, parse(from_os_str))]
+    #[clap(long = "mpath", short = 'm', global = true, parse(from_os_str))]
```
steelgeek091, line 25: This is just a temporary fix because there were two conflicting `--path` reported before. We should find the real reason for the conflict.

crates/rooch/src/extended_checks/metadata.rs
```diff
@@ -0,0 +1,280 @@
...
+                                panic!(
+                                    "{}",
+                                    format!(
+                                        "resource struct {:?} not defined in current module",
+                                        struct_name
```
steelgeek091, line 212: The error should be thrown in the following way instead of panicking directly:

```
self.env.error(&fun.get_loc(), "error message");
```

jolestar (approved):
LGTM

1. We need to put the `extended_checks` in a lower-level crate. Because the framework build and integration-test also need this.
2. We need to write more integration-test cases to test the private_generic.

We can merge this PR and do this in the following PR.

crates/rooch/src/extended_checks/metadata.rs
```diff
@@ -0,0 +1,280 @@
...
+        | SignatureToken::U8
+        | SignatureToken::U16
+        | SignatureToken::U32
+        | SignatureToken::U64
+        | SignatureToken::U128
+        | SignatureToken::U256
+        | SignatureToken::Address
+        | SignatureToken::Vector(_)
+        | SignatureToken::Signer
+        | SignatureToken::Reference(_)
```
jolestar, line 277: Do we not support `Reference`? Is the code as follows worked?

```
#[private_generics(T)]
public fun create_box<T>(value: &T): Box<T1> {
    Box<T1> { value: value }
}
```

jolestar:
Would placing this code in a crate called `moveos/moveos-verifier` be more appropriate?

crates/rooch/src/commands/move_cli/commands/new.rs
steelgeek091: Okay, I have raised an issue: https://github.com/rooch-network/rooch/issues/173

crates/rooch/src/extended_checks/metadata.rs
steelgeek091: I have tested the following code and it works properly:

```move
#[private_generics(T1, T3)]
public fun create_box<T1: store+copy, T2, T3>(value: &T1): Box<T1> {
    Box<T1> { value: *value }
} 

public fun run() {
    let data = &Data1{ v: 123 };
    let box_val = create_box<Data1, Data2, Box<u32>>(data);
    let _ = box_value(&box_val);
}   
```

Because we are checking the generic parameter `Data1` (which is an ordinary struct) passed when calling `create_box<Data1, Data2, Box<u32>>(data);` and not the type of the `value: &T1` parameter in the create_box function (which is a reference type), the code works properly.
Unless we not only check the type of the generic parameter, but also the type of the actual parameter, the code will not work properly.

steelgeek091:
> Would placing this code in a crate called `moveos/moveos-verifier` be more appropriate?

I think it is reasonable to call it 'moveos-verifier' because, essentially, 'extended_checks' also verify 'Metadata'.

Issue #155, opened 2023-05-18
[moveos_std] Implement the signature native function


jolestar:
- [ ] Ed25519
- [ ] Ethereum ECDSA

### Ref:

* https://github.com/starcoinorg/starcoin-framework/blob/main/sources/Signature.move

PR #178, opened 2023-05-26
https://github.com/rooch-network/rooch/issues/155

wow-sven:


jolestar (approved):

jolestar:
The signature and hash function should be part of moveos_std or rooch_framework?

wow-sven:
> The signature and hash function should be part of moveos_std or rooch_framework?

I'm not sure only the Authenticator will use it, so I brought it up. I can put it in Authenticator if I need to

PR #185, opened 2023-05-28
fix cli

### LINKED ISSUE #173: The path flag of the 'rooch move new' command should inherit from the move arguments.
The path flag of the 'rooch move new' command should inherit from the move arguments to avoid conflicts with the --path of Move.

```rust
#[derive(Parser)]
pub struct New {
    #[clap(flatten)]
    pub new: new::New,

    /// Path of the new package to create.
    #[clap(long = "path", short = 'p', global = true, parse(from_os_str))]
    pub path: Option<PathBuf>,
}
```

The following error will be thrown.

`
thread 'main' panicked at 'Command new: Long option names must be unique for each argument, but '--path' is in use by both 'path' and 'package-path'',
`

pause125:
1. Fix option conflicts in `rooch move new`. Resolve #173 .
2. Compatible windows system in rooch server.

### DIFF
--- a/crates/rooch/src/commands/move_cli/mod.rs
+++ b/crates/rooch/src/commands/move_cli/mod.rs
@@ -82,7 +82,7 @@ async fn run_cli(move_cli: MoveCli) -> CliResult<String> {
             .map_err(CliError::from),
         MoveCommand::New(c) => c
             .execute(move_args.package_path)
-            .map(|_| "success".to_string())
+            .map(|_| "Success".to_string())
             .map_err(CliError::from),
         MoveCommand::Prove(c) => c
             .execute(move_args.package_path, move_args.build_config)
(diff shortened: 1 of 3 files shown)


crates/rooch/src/commands/move_cli/mod.rs
```diff
@@ -80,7 +80,10 @@ async fn run_cli(move_cli: MoveCli) -> CliResult<String> {
             .execute(move_args.package_path, move_args.build_config)
             .map(|_| "Success".to_string())
             .map_err(CliError::from),
-        MoveCommand::New(c) => c.execute_serialized().await,
+        MoveCommand::New(c) => c
+            .execute(move_args.package_path)
+            .map(|_| "success".to_string())
```
jolestar, line 85: Be consistent with others to `Success`?

jolestar (approved):

PR #187, opened 2023-05-28
Unified crypto lib && Implement transaction signature verification

wow-sven:


crates/rooch-types/src/transaction/rooch.rs
```diff
@@ -126,7 +124,7 @@ impl AbstractTransaction for RoochTransaction {
     }
 
     fn encode(&self) -> Vec<u8> {
-        bcs::to_bytes(self).expect("encode transaction should success")
+        bcs::to_bytes(&self).expect("encode transaction should success")
```
jolestar, line 127: `self` is already a reference and does not need to `&self`

crates/rooch-types/src/transaction/rooch.rs
```diff
@@ -137,7 +135,8 @@ impl AuthenticatableTransaction for RoochTransaction {
     //TODO unify the hash function
     fn tx_hash(&self) -> H256 {
         //TODO cache the hash
-        moveos_types::h256::sha3_256_of(self.encode().as_slice())
+        let data = bcs::to_bytes(&self.data).expect("encode transaction should success");
```
jolestar, line 138: There are two hash:

1. the RoochTransaction's hash which includes the signature data and it is the ID of the transaction.
2. the RoochTransactionData's hash which is used for the signature. 

We should put the second one in the RoochTransactionData.

moveos/moveos-stdlib/rooch-framework/sources/authenticator.move
```diff
@@ -1,76 +1,108 @@
...
+      let AuthenticatorInfo { sender, sequence_number, authenticator } = info;
+      (sender, sequence_number, authenticator)
+   }
+
+   public fun decode_ed25519_authenticator(authenticator: Authenticator): Ed25519Authenticator {
+      assert!(authenticator.scheme == SCHEME_ED25519, EUnsupportedScheme);
+      moveos_std::bcd::from_bytes<Ed25519Authenticator>(authenticator.payload)
+   }
+
+   public fun ed25519Publick(this: &Ed25519Authenticator): vector<u8> {
```
jolestar, line 69: Publick -> Public

crates/testsuite/tests/integration.rs
```diff
@@ -23,9 +28,34 @@ async fn start_server(w: &mut World) {
...
+    let default_addr = match config {
+        Ok(v) => {
+            if v.exists() {
+                let config: RoochConfig =
+                    PersistedConfig::read(rooch_config_path().unwrap().as_path()).unwrap();
+                let config: PersistedConfig<RoochConfig> =
+                    config.persisted(rooch_config_dir().unwrap().join(ROOCH_CONFIG).as_path());
+
+                match config.active_address {
+                    Some(addr) => format!("0x{:x}", AccountAddress::from(addr)),
```
jolestar, line 48: account_address.to_hex_literal() == format!("0x{:x}", AccountAddress::from(addr))

crates/rooch-types/src/transaction/authenticator.rs
```diff
@@ -382,342 +209,10 @@ impl fmt::Display for Authenticator {
...
-
-impl fmt::Display for AuthenticationKey {
-    fn fmt(&self, f: &mut fmt::Formatter) -> std::fmt::Result {
-        // Forward to the LowerHex impl with a "0x" prepended (the # flag).
-        write!(f, "0x{:#x}", self)
-    }
-}
-
-#[derive(Clone, Debug, Hash, PartialEq, Eq, DeserializeKey, SerializeKey)]
-pub enum AccountPublicKey {
```
jolestar, line 528: The `AccountPublicKey` and `AccountPrivateKey` are suitable wrappers for supporting multi schemes. We can consider keeping them or introducing them again in the future PR.

wow-sven: Later, when implementing multi-schemes, I will consider bringing it back.

jolestar (approved):

PR #196, opened 2023-05-29
Implement the parameter and return value checking for the entry function.

steelgeek091:
According to issue https://github.com/rooch-network/rooch/issues/152, the function to check the parameters of the entry function has been implemented.

If the function is an entry function, the following checks will be performed:
- Check the type of the parameters to see if they meet the requirements.
- The function cannot have a return value.

Currently, the type of function parameters can only be one of the following types (more types can be added if necessary):

1. `PrimitiveType`
```rust
pub enum PrimitiveType {
    Bool,
    U8,
    U16,
    U32,
    U64,
    U128,
    U256,
    Address,
    Signer,
    Num,
    Range,
    EventStore,
}
```

2. `Vector<PrimitiveType>`
3. `0x1::string::String`
4. `0x1::object_id::ObjectID`

### DIFF
--- a/moveos/moveos-verifier/src/metadata.rs
+++ b/moveos/moveos-verifier/src/metadata.rs
@@ -2,7 +2,9 @@ use move_binary_format::binary_views::BinaryIndexedView;
 use move_binary_format::file_format::{Bytecode, FunctionInstantiation, SignatureToken};
 use move_core_types::language_storage::ModuleId;
 use move_model::ast::Attribute;
-use move_model::model::{FunctionEnv, GlobalEnv, ModuleEnv};
+use move_model::model::{FunctionEnv, GlobalEnv, Loc, ModuleEnv, QualifiedId, StructId};
+use move_model::ty::PrimitiveType;
+use move_model::ty::Type;
 use serde::{Deserialize, Serialize};
 use std::collections::BTreeMap;
 
@@ -64,7 +66,7 @@ impl<'a> ExtendedChecker<'a> {
         for ref module in self.env.get_modules() {
             if module.is_target() {
                 self.check_private_generics_functions(module);
-                //self.check_entry_functions(module);
+                self.check_entry_functions(module);
                 //self.check_init_module(module);
             }
         }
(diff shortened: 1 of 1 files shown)


steelgeek091:
During the compilation of `rooch-framework`, the following error was reported:

```
INCLUDING DEPENDENCY MoveNursery
INCLUDING DEPENDENCY MoveStdlib
INCLUDING DEPENDENCY MoveosStdLib
BUILDING RoochFramework
error: type `&mut storage_context::StorageContext` is not supported as a parameter type
   ┌─ /home/roy/coding/rooch/moveos/moveos-stdlib/rooch-framework/sources/account.move:80:4
   │  
80 │ ╭    public entry fun create_account_entry(ctx: &mut StorageContext, new_address: address){
81 │ │       Self::create_account(ctx, new_address);
82 │ │    }
   │ ╰────^

extended checks failed
```

The parameter type of this entry function is a reference type of `0x1::storage_context::StorageContext`. Do we really need this type of parameter?
If we need it, I will add this type to the type check list.

moveos/moveos-verifier/src/metadata.rs
```diff
@@ -230,6 +231,65 @@ impl<'a> ExtendedChecker<'a> {
...
+        }
+    }
+
+    fn is_allowed_input_struct(&self, qid: QualifiedId<StructId>) -> bool {
+        let name = self.env.get_struct(qid).get_full_name_with_address();
+        matches!(
+            name.as_str(),
+            "0x1::string::String"
+                | "0x1::object_id::ObjectID"
+                | "0x1::storage_context::StorageContext"
```
jolestar, line 288: 0x1::tx_context::TxContext

moveos/moveos-verifier/src/metadata.rs
```diff
@@ -230,6 +231,65 @@ impl<'a> ExtendedChecker<'a> {
...
+            }
+            Vector(ety) => {
+                // Vectors are allowed if element type is allowed
+                self.check_transaction_input_type(loc, ety)
+            }
+            Struct(mid, sid, _) if self.is_allowed_input_struct(mid.qualified(*sid)) => {
+                // Specific struct types are allowed
+            }
+            _ => {
+                // Everything else is disallowed.
```
jolestar, line 270: the `Type::Signer` and `Type::Reference(Type::Signer)` also need to support

jolestar (approved):

moveos/moveos-verifier/src/metadata.rs
steelgeek091: Fixed. Type::Signer was already a supported type in PrimitiveType, and Type::Reference(Type::Signer) type has been added.

moveos/moveos-verifier/src/metadata.rs
steelgeek091: Fixed.

PR #170, opened 2023-05-23
Implement event memory store and define event data structure

baichuan3:
this is part of https://github.com/rooch-network/rooch/issues/91

Implemented the following parts: 
- event memory store implement
- event data structure define

TODO：
1、event native implement
2、event filter
3、event RPC API

And also depend on https://github.com/rooch-network/rooch/issues/169

Issue #91, opened 2023-05-06
[Framework] event implementation


baichuan3:
Design and implement the event mechanism of rooch framework.
  
**1、aptos and starcoin event style**
```
    struct EventHandle<phantom T: drop + store> has store {
        /// Total number of events emitted to this event stream.
        counter: u64,
        /// A globally unique ID for this event stream.
        guid: GUID,
    }
```
[https://aptos.dev/concepts/events/](url)

Aptos event storage structure K-V
```
K: txn_info_id: HashValue,
V: events: Vec<Event>,
```

Starcoin event storage structure K-V, using an EventKey (sender address + counter) to express a global uuid.
```
K: txn_info_id: HashValue,
V: events: Vec<Event>,
```

**2、sui event style**
```
module sui::event {
    /// Emit a custom Move event, sending the data offchain.
    ///
    /// Used for creating custom indexes and tracking onchain
    /// activity in a way that suits a specific application the most.
    ///
    /// The type T is the main way to index the event, and can contain
    /// phantom parameters, eg emit(MyEvent<phantom T>).
    public native fun emit<T: copy + drop>(event: T);
}
```

```
/// Every MoveEvent has the following properties:
///  - sender
///  - type signature (`T`)
///  - event data (the value of `T`)
///  - timestamp (local to a node)
///  - transaction digest
```

https://docs.sui.io/build/event_api

Sui event storage structure K-V, 
```
DBMap<(TransactionEventsDigest, usize), Event>
```

And using an ```EventID {
                tx_digest,
                event_seq,
            } ``` to express a global uuid.

Rooch event data structure:
/// Every MoveEvent has the following properties:
///  - sender
///  - type signature (`T`)
///  - event data (the value of `T`)
///  - timestamp (local to a node)

TODO: 
1. Is it necessary to maintain the global event uuid in the contract? and certificate is necessary?
    where to storage  event root?  
    every transaction has a merkle tree
    ethereum event implement
3. event filter dimension, by sender? by event key/event id? by MoveEvent?
4. use Object as Storage

wubuku:
If we adopt the Aptos/Starcoin-style event mechanism implementation, I think we can also consider making some improvements, such as removing the counter field from EventHandle and making it an immutable object right after creation.

It may be necessary to give each event a unique ID. So, can we consider having Rooch provide a native method to obtain [UUIDs](https://en.wikipedia.org/wiki/Universally_unique_identifier)?

[...]

PR #197, opened 2023-05-29
implements native event, event filter and event RPC API

baichuan3:
Compared with the implementation, the event scheme of the design has been adjusted. The original idea is to implement it based on Sui Event. The advantage is that the entire event is implemented based on native. There is no need to provide mechanisms such as guid and counter in the framework, which is relatively simple. However, during the implementation process, it was discovered that in order to implement an event mechanism similar to Sui, it is necessary to implement a layer of runtime extensions on the existing Move VM virtual machine to implement transactions and event adapters. Sui achieves similar effects based on ObjectRuntime, and the cost Higher, so the function is implemented based on the Aptos/Starcoin style first, and then refactored and optimized.

this is part of https://github.com/rooch-network/rooch/issues/91 and  is a follow-up to this issue https://github.com/rooch-network/rooch/pull/170

Implemented the following parts:

- native event
- event filter
- event RPC API
- provide unified storage into MoveOSDB to prepare for the next storage implementation.

TODO：
1、event proof
2、test case

Refactor TOOD：
1、event RPC API, data flow: event -> MoveOSEvent -> EventView，There are multiple conversions, and when converting FromIterator<Vec<T>> to Vec<T>, the existing method is multiple Vec::new(), and then consider implementing the corresponding FromIterator se/de method to achieve a more elegant convert.
2、unify trait TableResolver/ModuleResolver/ResourceResolver for MoveOSDB and StateDB


moveos/moveos-store/src/state_store/mod.rs
```diff
@@ -322,6 +323,46 @@ impl ResourceResolver for StateDB {
...
+            .map(|s| {
+                ensure!(
+                    s.match_struct_type(tag),
+                    "Resource type mismatch, expected: {:?}, actual: {:?}",
+                    tag,
+                    s.value_type
+                );
+                Ok(s.value)
+            })
+            .transpose()
```
jolestar, line 347: We should directly call the StateDB::get_resource here, do not repeat.

moveos/moveos-store/src/state_store/mod.rs
```diff
@@ -322,6 +323,46 @@ impl ResourceResolver for StateDB {
...
+
+    fn get_module(&self, module_id: &ModuleId) -> Result<Option<Vec<u8>>, Self::Error> {
+        let module_table_id = NamedTableID::Module(*module_id.address()).to_object_id();
+        let key = module_name_to_key(module_id.name());
+        //We wrap the modules byte codes to `MoveModule` type when store the module.
+        //So we need unwrap the MoveModule type.
+        self.state_store
+            .get_with_key(module_table_id, key)?
+            .map(|s| Ok(s.as_move_state::<MoveModule>()?.byte_codes))
+            .transpose()
```
jolestar, line 362: Same

moveos/moveos-store/src/state_store/mod.rs
```diff
@@ -346,6 +387,28 @@ fn module_name_to_key(name: &IdentStr) -> Vec<u8> {
     bcs::to_bytes(&name.to_string()).expect("bcs to_bytes String must success.")
 }
 
+impl TableResolver for MoveOSDB {
+    fn resolve_table_entry(
+        &self,
+        handle: &TableHandle,
+        key: &[u8],
+    ) -> std::result::Result<Option<TableValueBox>, anyhow::Error> {
+        let state = if handle.0 == storage_context::GLOBAL_OBJECT_STORAGE_HANDLE {
```
jolestar, line 396: Same

moveos/moveos-stdlib/rooch-framework/sources/events.move
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    /// A handle for an event such that:
+    /// 1. Other modules can emit events to this handle.
+    /// 2. Storage can use this handle to prove the total number of events that happened in the past.
+    struct EventHandle<phantom T: drop + store> has store {
+        /// Total number of events emitted to this event stream.
+        counter: u64,
+        /// A globally unique ID for this event stream.
+        guid: vector<u8>,
```
jolestar, line 28: Use the ObjectID as `guid` type? like the TableHandle

moveos/moveos-stdlib/rooch-framework/sources/events.move
```diff
@@ -0,0 +1,87 @@
...
+    fun fresh_guid(counter: &mut EventHandleGenerator): vector<u8> {
+        let sender_bytes = bcs::to_bytes(&counter.addr);
+        let count_bytes = bcs::to_bytes(&counter.counter);
+        counter.counter = counter.counter + 1;
+
+        // EventHandleGenerator goes first just in case we want to extend address in the future.
+        vector::append(&mut count_bytes, sender_bytes);
+
+        count_bytes
+    }
```
jolestar, line 59: How about we use the TxContxt::fresh_object_id() to generate the event guid?

moveos/moveos-stdlib/rooch-framework/sources/events.move
```diff
@@ -0,0 +1,87 @@
...
+    /// Publishs a new event handle generator.
+    public fun publish_generator(ctx: &mut StorageContext, account: &signer) {
+        let addr = signer::address_of(account);
+        assert!(!account_storage::global_exists<EventHandleGenerator>(ctx, addr), error::already_exists(EEventGenerator));
+        account_storage::global_move_to(
+            ctx,
+            account,
+            EventHandleGenerator{ counter: 0, addr }
+        )
+    }
```
jolestar, line 43: We do not need the `EventHandleGenerator` if we use the ObjectID as the event stream `guid`. This makes the event API simpler.

moveos/moveos-types/src/filter.rs
```diff
@@ -0,0 +1,142 @@
...
+use crate::contract_event::ContractEvent;
+use crate::event::EventKey;
+use crate::language_storage::TypeTag;
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Filter {
+    /// Blockchain will be searched from this block.
+    pub from_block: BlockNumber,
+    /// Till this block.
+    pub to_block: BlockNumber,
```
jolestar, line 16: The executor does not know the block, so this should be `from_tx` `to_tx`, But we can wait for the tx API to ensure this.

moveos/moveos-stdlib/src/natives/rooch_framework/event.rs
jolestar, line 1: The `event-store` in MoveOS, so the `event` native and `events.move` should be in the moveos_stdlib

moveos/moveos-stdlib/rooch-framework/sources/events.move
baichuan3: I tried. Originally designed according to the sui event style, the contract does not contain guid at all, which is the simplest and more intuitive for developers.

However, when implementing native event, it is found that to implement an event mechanism similar to Sui, it is necessary to expand the existing Move VM virtual machine to implement a runtime event adapter. 

The current transaction is based on the native Move VM, and the transaction and event are bound, which means that to expand the Event, it needs to be connected with the transaction, and the cost of implementation is very high.

Sui achieves a similar effect based on ObjectRuntime,  Its transactions and events are generated based on ObjectRuntime.

Events are related to the following two interfaces of the Move VM core:

1. Save event in native function: 
```
context.save_event(guid, seq_num, ty, msg)
```

2. Derived events after the transaction is executed:
```
let (change_set, raw_events, mut extensions) = session.finish_with_extensions()?;
```

moveos/moveos-stdlib/rooch-framework/sources/events.move
baichuan3: Maybe i can use the Vec[] corresponding to the ObjectID, and then can use the Move VM related interface.

moveos/moveos-stdlib/moveos-stdlib/sources/events.move
```diff
@@ -0,0 +1,23 @@
...
+
+    /// Emit a custom Move event, sending the data offchain.
+    ///
+    /// Used for creating custom indexes and tracking onchain
+    /// activity in a way that suits a specific application the most.
+    ///
+    /// The type T is the main way to index the event, and can contain
+    /// phantom parameters, eg emit(MyEvent<phantom T>).
+    public fun emit_event<T: drop + store>(ctx: &mut TxContext, event: T) {
+        let guid = tx_context::fresh_object_id(ctx);
```
jolestar, line 14: The original GUID is an event stream id representing the Event Type. the GUID + event sequence_number = EventID(or EventKey).

If we generate a Guid for everyone event, EventKey must also be changed.

baichuan3:
already use object_id as the guid value of the event @jolestar

moveos/moveos-types/src/event_filter.rs
```diff
@@ -0,0 +1,202 @@
...
+        let Event {
+            key,
+            sequence_number,
+            type_tag,
+            event_data,
+            event_index,
+        } = event;
+
+        //TODO how to store and derive sender address ?
+        let sender = AccountAddress::ZERO;
```
jolestar, line 72: The current implemention does not store the sender in the GUID, so we can not know the sender. If we want to know the sender, we need to store then sender to the Event.

jolestar:
The discord discussion: https://discord.com/channels/1078938449974935592/1112801688785535117

jolestar (approved):
According to the discussion on Discord, merge first and then refactor.

Issue #108, opened 2023-05-11
[RPC]Implementing ETH-Compatible RPC Proxy for Rooch


jolestar:
## Motivation

To enable ETH ecosystem users to use Rooch conveniently, we need to implement an ETH-compatible RPC proxy on top of an existing RPC service used by Rooch. This will enable ETH clients to interact with Rooch and use the features of a compatible Ethereum wallet.

### The ETH-compatible RPC proxy

A proxy service would be developed to provide an interface between the Rooch RPC and the Ethereum client. The proxy would have to support the necessary methods used in ETH protocol such as account balance retrieval, transfers, smart contract calls, and ensure ETH client requests are received.

### Transform the input and output formats.

As input and output formats used by the existing RPC may differ from those used in the ETH protocol, we need to develop a mechanism to transform input and output formats making them compatible with ETH protocol. 

1. In the transaction format, we need to support the original eth transaction because the user signature is for the original transaction. If We convert it to Rooch's transaction, we can not verify it.
2. In the address format, we need to provide a method to let the Rooch users can bind an ETH address, then provide a method to resolve the Eth address to Rooch address.

- [x] #277
- [x] #278
- [x] #279

Ref:

1. https://github.com/starcoinorg/sips/issues/22

jolestar:
I define some ETH API in #174. We need to determine if we want to support Metamask and which APIs are required first. @yubing744

yubing744:
I found the API used by the basic functions by debugging the MetaMask wallet：

| Use Case                        | APIs Used                                                |
| ------------------------------- | ---------------------------------------------------------|
| Query balance                   | `eth_getBalance`                                          |
| Query current Gas price         | `eth_gasPrice`                                            |
| Send transaction                | `eth_sendRawTransaction`, `eth_estimateGas`              |
| Check transaction status        | `eth_getTransactionByHash`, `eth_getTransactionReceipt`  |
| Query smart contract state      | `eth_call`                                                |

jolestar:
Please help to mock these APIs and let MetaMask try to connect Rooch

yubing744:
@jolestar I have mocked some APIs required by MetaMask, now we can configure the rooch server RPC address to MetaMask and initiate transactions.

![image](https://github.com/rooch-network/rooch/assets/1449069/8dda8d2b-f5f1-4f2c-ad48-9083a1b98cff)

jolestar:
[...]

PR #201, opened 2023-05-31
[rooch-server] mock eth compatible rpc

yubing744:
I have mocked some APIs required by MetaMask, now we can configure the rooch server RPC address to MetaMask and initiate transactions.

Ref https://github.com/rooch-network/rooch/issues/108

### DIFF
--- a/crates/rooch-server/src/jsonrpc_types/bytes.rs
+++ b/crates/rooch-server/src/jsonrpc_types/bytes.rs
@@ -23,18 +23,6 @@ impl Bytes {
     }
 }
 
-impl From<Vec<u8>> for Bytes {
-    fn from(bytes: Vec<u8>) -> Bytes {
-        Bytes(bytes)
-    }
-}
-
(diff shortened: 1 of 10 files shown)


jolestar (approved):
LGTM

crates/rooch-server/src/jsonrpc_types/bytes.rs
```diff
@@ -0,0 +1,110 @@
...
+use rustc_hex::{FromHex, ToHex};
+use serde::{
+    de::{Error, Visitor},
+    Deserialize, Deserializer, Serialize, Serializer,
+};
+use std::fmt;
+
+/// Wrapper structure around vector of bytes.
+#[derive(Debug, PartialEq, Eq, Default, Hash, Clone)]
+pub struct Bytes(pub Vec<u8>);
```
jolestar, line 13: How about use the StrView<Vec<u8>> to replace Bytes?

PR #207, opened 2023-05-31
use self hosted runner

yourmoonlight:


jolestar:
The CI cluster config file also needs to commit to the git repo. @yourmoonlight

jolestar (approved):

Issue #204, opened 2023-05-31
build moveos stdlib failed due to move `ObjectID` verifier


baichuan3:
An error is reported when compiling moveos stdlib with the main branch, failed due to 'resource struct "ObjectID" not defined in current module.

reproduce path:

cd the following directory
`rooch-network/rooch/moveos/moveos-stdlib/moveos-stdlib`

then execute:
`rooch move build`

error stacktrace:
```
rooch move build
INCLUDING DEPENDENCY MoveNursery
INCLUDING DEPENDENCY MoveStdlib
BUILDING MoveosStdLib
thread 'main' panicked at 'resource struct "ObjectID" not defined in current module', moveos/moveos-verifier/src/metadata.rs:204:33
```

PR #208, opened 2023-05-31
Fix error message printing when checking private_generics.

steelgeek091:
resolve https://github.com/rooch-network/rooch/issues/204

1. If a function decorated with `#[private_generics(T)]` is called with a generic parameter type that is not allowed, print the code where the `CallGeneric` instruction is located, instead of the called function.
2. In `moveos-stdlib`, some `#[private_generics(T)`] have `T` that is not defined in the current function.

### DIFF
--- a/moveos/moveos-stdlib/moveos-stdlib/sources/bcd.move
+++ b/moveos/moveos-stdlib/moveos-stdlib/sources/bcd.move
@@ -30,7 +30,7 @@ module moveos_std::bcd{
     //1. The caller of from_bytes is the module that defines the `T`. This is ensured by private_generics.
     //2. The fields contained in `T` are either primitive types or are defined by the module that calls from_bytes. 
     //We need to find a solution to this problem. If we cannot solve it, then we cannot set from_bytes to public.
-    #[private_generics(T)]
+    // #[private_generics(MoveValue)]
     /// Function to deserialize a type T.
     /// Note the `private_generics` ensure only the `MoveValue`'s owner module can call this function
     public native fun from_bytes<MoveValue>(bytes: vector<u8>): MoveValue;
--- a/moveos/moveos-verifier/src/metadata.rs
+++ b/moveos/moveos-verifier/src/metadata.rs
@@ -1,6 +1,7 @@
 use move_binary_format::binary_views::BinaryIndexedView;
 use move_binary_format::file_format::{Bytecode, FunctionInstantiation, SignatureToken};
 use move_core_types::language_storage::ModuleId;
+// use move_model::model::ModuleId;
 use move_model::ast::Attribute;
 use move_model::model::{FunctionEnv, GlobalEnv, Loc, ModuleEnv, QualifiedId, StructId};
 use move_model::ty::PrimitiveType;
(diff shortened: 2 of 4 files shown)


moveos/moveos-verifier/src/metadata.rs
```diff
@@ -340,18 +361,20 @@ fn is_defined_in_current_module(
                 view.identifier_at(shandle.name).to_string(),
             )
         }
-        SignatureToken::TypeParameter(_)
-        | SignatureToken::Bool
+        // Primitive types are allowed.
```
jolestar, line 364: Should we support Primitive type and Address?

steelgeek091: In the file moveos-stdlib/sources/bcd.move, there is the following code:

```move
module moveos_std::bcd{
    public fun to_bool(v: vector<u8>): bool {
        from_bytes<bool>(v)
    }

    public fun to_u8(v: vector<u8>): u8 {
        from_bytes<u8>(v)
    }

    public fun to_u64(v: vector<u8>): u64 {
        from_bytes<u64>(v)
    }

    public fun to_u128(v: vector<u8>): u128 {
        from_bytes<u128>(v)
    }

    public fun to_address(v: vector<u8>): address {
        from_bytes<address>(v)
    }

    #[private_generics(MoveValue)]
    /// Function to deserialize a type T.
    /// Note the `private_generics` ensure only the `MoveValue`'s owner module can call this function
    public native fun from_bytes<MoveValue>(bytes: vector<u8>): MoveValue;
}
```

jolestar: This is a specific case, we should not support Primitive type and Address in `private_generics`. We need to refactor the `bcd`:

1. Make the native function `from_bytes` private and without annotation.
2. Provide a public function and add `private_generics`, maybe `public_from_bytes` or `generic_from_bytes`.
3. the `to_xx` function calls the private `from_bytes`.

steelgeek091: Understood. Do we need to prepare to solve these problems?
If we are not ready to solve them yet, I will exclude the `bcd::from_bytes` function from the scope of verification in the verifiter for now.

steelgeek091: `1. Make the native function from_bytes private and without annotation`

To solve this problem, we need to handle it specially in the verifier. We need to implement the same functionality as #[private_generics] without requiring annotations.

jolestar: You can refactor it in this PR or just remove `private_generics` from `from_bytes` and record an issue.

moveos/moveos-stdlib/moveos-stdlib/sources/bcd.move
```diff
@@ -30,7 +30,7 @@ module moveos_std::bcd{
     //1. The caller of from_bytes is the module that defines the `T`. This is ensured by private_generics.
     //2. The fields contained in `T` are either primitive types or are defined by the module that calls from_bytes. 
     //We need to find a solution to this problem. If we cannot solve it, then we cannot set from_bytes to public.
-    #[private_generics(T)]
+    // #[private_generics(MoveValue)]
```
steelgeek091, line 33: I have recorded an issue at https://github.com/rooch-network/rooch/issues/210, and will handle this problem later.

moveos/moveos-verifier/src/metadata.rs
steelgeek091: I have temporarily removed the `#[private_generics]` attribute from `from_bytes<T>` and recorded an issue at https://github.com/rooch-network/rooch/issues/210 and will handle this problem later.

jolestar (approved):

Issue #202, opened 2023-05-31
Before executing the entry function, call the verifier to check its attributes.


steelgeek091:
In the `execute_transaction` function of the `moveos.rs` file, before executing the `entry function`, the verifier should be called to check if its parameters and return values meet the requirements (Aptos also does this.).

https://github.com/rooch-network/rooch/blob/main/moveos/moveos/src/moveos.rs#L197

PR #223, opened 2023-06-03
Check function parameters before executing the function and script.

steelgeek091:
According to issue https://github.com/rooch-network/rooch/issues/202.

When MoveOS executes functions and scripts, the verifier checks the parameters and return values of the functions before execution, in the same way as it checks the entry function during compile time.

### DIFF
--- a/moveos/moveos-verifier/src/metadata.rs
+++ b/moveos/moveos-verifier/src/metadata.rs
@@ -1,11 +1,13 @@
+use crate::build::ROOCH_METADATA_KEY;
 use move_binary_format::binary_views::BinaryIndexedView;
 use move_binary_format::file_format::{Bytecode, FunctionInstantiation, SignatureToken};
 use move_core_types::language_storage::ModuleId;
-// use move_model::model::ModuleId;
+use move_core_types::metadata::Metadata;
 use move_model::ast::Attribute;
-use move_model::model::{FunctionEnv, GlobalEnv, Loc, ModuleEnv, QualifiedId, StructId};
+use move_model::model::{FunctionEnv, GlobalEnv, Loc, ModuleEnv};
 use move_model::ty::PrimitiveType;
 use move_model::ty::Type;
+use move_vm_runtime::move_vm::MoveVM;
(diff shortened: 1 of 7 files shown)


moveos/moveos/src/moveos.rs
```diff
@@ -190,6 +193,11 @@ impl MoveOS {
                 let args = session
                     .resolve_args(&tx_context, loaded_function, function.args)
                     .map_err(|e| e.finish(Location::Undefined))?;
+
+                let func =
+                    session.load_function(&function.function_id, function.ty_args.as_slice())?;
+                moveos_verifier::verifier::verify_entry_function(func, session.runtime_session())?;
+
```
jolestar, line 200: `verify_entry_function` should be before the `resolve_args`.

moveos/moveos-verifier/src/metadata.rs
```diff
@@ -313,25 +333,28 @@ impl<'a> ExtendedChecker<'a> {
...
-                | "0x1::tx_context::TxContext"
-        )
-    }
+pub fn is_allowed_input_struct(name: String) -> bool {
+    matches!(
+        name.as_str(),
+        "0x1::string::String"
+            | "0x1::object_id::ObjectID"
+            | "0x1::storage_context::StorageContext"
+            | "0x1::tx_context::TxContext"
```
jolestar, line 356: 0x1::ascii::String also support.

jolestar (approved):

jolestar:
So the #152 and #202 can be marked as resolved?

moveos/moveos/src/moveos.rs
steelgeek091: Got it. I will submit a pull request to modify it later.

moveos/moveos-verifier/src/metadata.rs
steelgeek091: Okay, I will add it in a new pull request.

steelgeek091:
> So the #152 and #202 can be marked as resolved?

Right, these two issues have been resolved.

Issue #203, opened 2023-05-31
verification code for init function in module


steelgeek091:
In the `moveos.rs` file, when the executed transaction is of type `MoveAction::ModuleBundle`, the module is checked for an `init` function, and if it exists, the code checks its attributes before executing it.

Do we need to move this verification code to moveos_verifier?

https://github.com/rooch-network/rooch/blob/main/moveos/moveos/src/moveos.rs#L301

@jolestar

jolestar:
Yes, we need this.

jolestar:
> I can try to complete this feature

@steelgeek091 Are you currently working on this issue?

steelgeek091:
> > I can try to complete this feature
> 
> @steelgeek091 Are you currently working on this issue?

Yes, some relevant code has been written.

steelgeek091:
> > I can try to complete this feature
> 
> @steelgeek091 Are you currently working on this issue?

I haven't started working on this issue yet, I am currently working on https://github.com/rooch-network/rooch/issues/202.

jolestar:
This issue is resolved. If not, please reopen it.

PR #232, opened 2023-06-05
[WIP]Add verifier for module init function 

WGB5445:
This is the code implemented for #203 

A new verify has been added in verifier.rs_ Init_ Function function and replace or delete some code with the same function in moveos

### DIFF
--- a/moveos/moveos-verifier/src/verifier.rs
+++ b/moveos/moveos-verifier/src/verifier.rs
@@ -1,9 +1,75 @@
 use crate::metadata::is_allowed_input_struct;
-use anyhow::{Error, Result};
-use move_core_types::resolver::MoveResolver;
+use anyhow::{Error, Ok, Result};
+use move_binary_format::file_format::Visibility;
+use move_binary_format::{access::ModuleAccess, CompiledModule};
+use move_core_types::move_resource::MoveStructType;
+use move_core_types::{identifier::Identifier, resolver::MoveResolver};
 use move_vm_runtime::session::{LoadedFunctionInstantiation, Session};
-use move_vm_types::loaded_data::runtime_types::Type;
+use move_vm_types::loaded_data::runtime_types::{StructType, Type};
+use moveos_types::move_types::FunctionId;
+use moveos_types::storage_context::StorageContext;
+use once_cell::sync::Lazy;
 use std::ops::Deref;
+use std::sync::Arc;
+
+pub static INIT_FN_NAME_IDENTIFIER: Lazy<Identifier> =
+    Lazy::new(|| Identifier::new("init").unwrap());
+
+fn as_struct_no_panic<S>(session: &Session<S>, t: &Type) -> Option<Arc<StructType>>
+where
+    S: MoveResolver,
+{
+    match t {
+        Type::Struct(s) | Type::StructInstantiation(s, _) => session.get_struct_type(*s),
+        Type::Reference(r) => as_struct_no_panic(session, r),
+        Type::MutableReference(r) => as_struct_no_panic(session, r),
+        _ => None,
+    }
+}
+
+fn is_storage_context(t: &StructType) -> bool {
+    *t.module.address() == *moveos_types::addresses::MOVEOS_STD_ADDRESS
+        && t.module.name() == StorageContext::module_identifier().as_ident_str()
+        && t.name == StorageContext::struct_identifier()
+}
+
+pub fn verify_init_function<S>(module: &CompiledModule, session: &Session<S>) -> Result<bool>
+where
+    S: MoveResolver,
+{
+    for fdef in &module.function_defs {
+        let fhandle = module.function_handle_at(fdef.function);
+        let fname = module.identifier_at(fhandle.name);
+        if fname == INIT_FN_NAME_IDENTIFIER.clone().as_ident_str() {
+            if Visibility::Private != fdef.visibility {
+                return Err(Error::msg("init function should private".to_string()));
+            } else if !fdef.is_entry {
+                return Err(Error::msg(
+                    "init function should not entry function".to_string(),
+                ));
+            } else {
+                let function_id =
+                    FunctionId::new(module.self_id(), INIT_FN_NAME_IDENTIFIER.clone());
+                let loaded_function = session.load_function(
+                    &module.self_id(),
+                    &function_id.function_name,
+                    vec![].as_slice(),
+                )?;
+                let Some((_i, _t)) = loaded_function.parameters.iter().enumerate().find(|(i, t)| {
+                        let struct_type = as_struct_no_panic(session, t);
+                        (*i as u32 == 0u32) && Option::is_some(&struct_type) && is_storage_context(&(struct_type.unwrap()))
+                    }) else {
+                        return Ok(true)
+                    };
+
+                if !(loaded_function.return_.is_empty()) {
+                    return Ok(true);
+                }
+            }
+        }
+    }
+    return Err(Error::msg("module not have init function".to_string()));
+}
 
 pub fn verify_entry_function<S>(
     func: LoadedFunctionInstantiation,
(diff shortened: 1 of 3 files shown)


moveos/moveos-verifier/src/verifier.rs
```diff
@@ -1,9 +1,75 @@
...
+                let function_id =
+                    FunctionId::new(module.self_id(), INIT_FN_NAME_IDENTIFIER.clone());
+                let loaded_function = session.load_function(
+                    &module.self_id(),
+                    &function_id.function_name,
+                    vec![].as_slice(),
+                )?;
+                let Some((_i, _t)) = loaded_function.parameters.iter().enumerate().find(|(i, t)| {
+                        let struct_type = as_struct_no_panic(session, t);
+                        (*i as u32 == 0u32) && Option::is_some(&struct_type) && is_storage_context(&(struct_type.unwrap()))
```
steelgeek091, line 60: The code logic here needs to be clearer and more readable, and comments should be added.

moveos/moveos-verifier/src/verifier.rs
```diff
@@ -1,9 +1,80 @@
...
```
steelgeek091, line 65: The code logic here needs to be clearer and more readable, and comments should be added.

moveos/moveos-verifier/src/verifier.rs
```diff
@@ -1,9 +1,75 @@
...
+                    }) else {
+                        return Ok(true)
+                    };
+
+                if !(loaded_function.return_.is_empty()) {
```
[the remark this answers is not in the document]
WGB5445: These two issues are indeed my mistakes and will be fixed in the next submission

moveos/moveos-verifier/src/verifier.rs
WGB5445: yes, thank you for your suggestion. I am adding the legality judgment of the init module in metadata.rs and will make corresponding modifications here as well

steelgeek091: I suggest that all checks regarding the init function should be placed in `verifier.rs`. 

`metadata.rs` should contain code related to compile-time checks, while `verifier.rs` should contain code related to runtime checks.

moveos/moveos-verifier/src/metadata.rs
```diff
@@ -270,6 +273,59 @@ impl<'a> ExtendedChecker<'a> {
...
 impl<'a> ExtendedChecker<'a> {
+    fn check_init_module(&mut self, module: &ModuleEnv) {
+        for ref fun in module.get_functions() {
+            if fun.get_identifier().as_ident_str() != INIT_FN_NAME_IDENTIFIER.as_ident_str() {
+                continue;
+            }
+
+            if Visibility::Private != fun.visibility() {
+                self.env
+                    .error(&fun.get_loc(), "module init function should private")
```
baichuan3, line 284: module init verify is different from other verify. It is expected to find the init method that satisfies the conditions and initialize it, instead of throwing an exception

moveos/moveos-verifier/src/verifier.rs
```diff
@@ -1,10 +1,80 @@
...
+                    FunctionId::new(module.self_id(), INIT_FN_NAME_IDENTIFIER.clone());
+                let loaded_function = session.load_function(
+                    &module.self_id(),
+                    &function_id.function_name,
+                    vec![].as_slice(),
+                )?;
+                let parameters_usize = loaded_function.parameters.len();
+                if parameters_usize != 1 {
+                    return Err(Error::msg(
+                        "init function only should have a parameter with storageContext"
```
baichuan3, line 47: as the same

moveos/moveos-verifier/src/metadata.rs
WGB5445: The init function should be verified during execution and compilation

so I understand that any function that does not comply with the init function regulations should have an exception during compilation in order to modify the code

moveos/moveos-verifier/src/verifier.rs
```diff
@@ -1,10 +1,80 @@
...
+                }
+                for ref ty in loaded_function.parameters {
+                    match ty {
+                        Type::Struct(s) | Type::StructInstantiation(s, _) => {
+                            let struct_type = session.get_struct_type(*s).unwrap();
+                            if *struct_type.module.address()
+                                == *moveos_types::addresses::MOVEOS_STD_ADDRESS
+                                && struct_type.module.name()
+                                    == StorageContext::module_identifier().as_ident_str()
+                                && struct_type.name == StorageContext::struct_identifier()
```
steelgeek091, line 59: The checking logic code used here for checking the init function is inconsistent with the logic in the check_init_module function in [metadata.rs](http://metadata.rs/). It should be consistent and abstract a common checking logic.

WGB5445: This problem has been resolved

moveos/moveos-verifier/src/metadata.rs
```diff
@@ -270,6 +275,85 @@ impl<'a> ExtendedChecker<'a> {
...
+                                "module init function only should have two parameter with signer or storageContext",
+                            )
+                        }
+                    }
+                    Type::Primitive(primitive) => {
+                        if let PrimitiveType::Signer = primitive {
+                        } else {
+                            self.env.error(
+                            &fun.get_loc(),
+                            "module init function only should have two parameter with signer or storageContext",
```
steelgeek091, line 343: The full name of the function should be printed in the error message.

moveos/moveos-verifier/src/metadata.rs
```diff
@@ -270,6 +275,85 @@ impl<'a> ExtendedChecker<'a> {
...
+            for ty in arg_tys {
+                match ty {
+                    
+                    Type::Reference(true, bt) => {
+                        
+                        let struct_tag = bt.clone().into_struct_tag(self.env);
+                        if struct_tag.is_none() {
+                            self.env.error(
+                                &fun.get_loc(),
+                                "module init function only should have two parameter with signer or storageContext"
```
steelgeek091, line 317: Similar multiple error messages like the following should be put at the same exit instead of being repeated multiple times.

`"module init function only should have two parameter with signer or storageContext"`

steelgeek091 (changes requested):

moveos/moveos-verifier/src/metadata.rs
WGB5445: Good suggestion
It has been changed to a different error message
Which will help developers have a clearer understanding of their code

examples/counter/sources/counter.move
```diff
@@ -13,7 +13,7 @@ module rooch_examples::counter {
         counter.value = counter.value + 1;
      }
 
-     public entry fun init(account: signer){
+     public entry fun init_entry(account: signer){
```
steelgeek091, line 16: Is this a temporary method, or will there never be a function like `entry fun init()` in the future?

WGB5445: Yes, the init function name cannot be occupied

jolestar: Please help refactor the counter example according to the standard of the init method, or you can start a new PR to implement it.

crates/testsuite/features/cmd.feature
```diff
@@ -13,7 +13,7 @@ Feature: Rooch CLI integration tests
       # TODO split Scenario for every example
       # counter example
       Then cmd: "move publish -p ../../examples/counter --sender-account {default} --named-addresses rooch_examples={default}"
-      Then cmd: "move run --function {default}::counter::init --sender-account {default}"
+      Then cmd: "move run --function {default}::counter::init_entry --sender-account {default}"
```
baichuan3, line 16: there is an example dedicated to verifying module init, use this example `module_init_example `?

steelgeek091 (approved):

PR #236, opened 2023-06-06
rename event  parsed_json to parsed_event_data

baichuan3:


jolestar (approved):
I merged this PR first. Later, we can consider refactoring using Event and AnnotatedEvent.

Issue #227, opened 2023-06-05
When the execution of the contract is failed, the error message shows the specific function name


baichuan3:
The current error message shows function number, not sure whether it is in the natural order of moduleli or in the dictionary order?

```
"ExecutionFailure": {
      "location": {
        "Module": {
          "address": "0000000000000000000000000000000000000000000000000000000000000001",
          "name": "raw_table"
        }
      },
      "function": 9,
      "code_offset": 0
    }
  }
}
```

Expect to display the error function name directly.

steelgeek091:
@baichuan3 Which part of the code triggers the return of these error messages?

baichuan3:
> @baichuan3 Which part of the code triggers the return of these error messages?

This error will be thrown when the rooch move run is executed on the command line and the execution contract method fails.

corresponding to the specific code location
https://github.com/rooch-network/rooch/blob/adce7f294e3bcc6f0374f23acb11e02ce712e661/moveos/moveos/src/moveos.rs#L240-L276

steelgeek091:
Alright, received.

steelgeek091:
@baichuan3 Can we take a look at the related error messages? This code block is too long, and it's not easy to find where the error occurred.

jolestar:
There are usually two types of errors in Move:

1. ExecutionFailure: A native method throws an error, or an instruction throws an error.
2. Abort: The contract explicitly aborts or aborts through assert.

Designing these two scenarios can reproduce the errors.

Starcoin's implemention:

https://github.com/starcoinorg/starcoin/blob/ee3c02fc55cdf99a0d00072d1f2847052641ca73/vm/vm-status-translator/src/lib.rs#L80

PR #237, opened 2023-06-07
Return a customized error message when there is an error during function verification or execution.

steelgeek091:
When there is an error in MoveVM verification or execution function, return a customized error message which includes the module and function name.

resolve https://github.com/rooch-network/rooch/issues/227.

### DIFF
--- a/crates/rooch-types/src/transaction/mod.rs
+++ b/crates/rooch-types/src/transaction/mod.rs
@@ -4,7 +4,8 @@
 use self::authenticator::Authenticator;
 use crate::{address::MultiChainAddress, H256};
 use anyhow::Result;
-use move_core_types::{account_address::AccountAddress, vm_status::KeptVMStatus};
+use move_core_types::account_address::AccountAddress;
+use move_core_types::vm_status::KeptVMStatus;
 use moveos_types::h256;
 use moveos_types::transaction::AuthenticatableTransaction;
 use serde::{Deserialize, Serialize};
--- a/moveos/moveos/src/moveos.rs
+++ b/moveos/moveos/src/moveos.rs
@@ -1,6 +1,7 @@
 // Copyright (c) RoochNetwork
 // SPDX-License-Identifier: Apache-2.0
 
+use crate::vm::vm_status_explainer::{explain_vm_status, VmStatusExplainView};
 use crate::vm::{
     move_vm_ext::{MoveVmExt, SessionExt},
     tx_argument_resolver::{as_struct_no_panic, is_storage_context},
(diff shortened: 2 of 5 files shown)


crates/rooch-types/src/transaction/mod.rs
```diff
@@ -170,7 +171,7 @@ pub struct TransactionInfo {
     /// The vm status. If it is not `Executed`, this will provide the general error class. Execution
     /// failures and Move abort's receive more detailed information. But other errors are generally
     /// categorized with no status code or other information
-    pub status: KeptVMStatus,
+    pub status: VmStatusExplainView,
```
jolestar, line 174: the View type only for the JSON RPC output, the transaction info should keep the original Status

moveos/moveos/src/vm/vm_status_explainer.rs
```diff
@@ -0,0 +1,124 @@
...
+        location: AbortLocation,
+        function: u16,
+        // The function name refers to the function that fails during MoveVM verify or execution,
+        // not the function called by the user.
+        function_name: Option<String>,
+        code_offset: u16,
+    },
+}
+
+pub fn explain_vm_status(state_db: &StateDB, vm_status: VMStatus) -> Result<VmStatusExplainView> {
```
jolestar, line 52: 这里的 StateDB 可以用 `ModuleResolver`

crates/rooch-types/src/transaction/mod.rs
steelgeek091: Okay, it has been fixed.

moveos/moveos/src/vm/vm_status_explainer.rs
steelgeek091: I have tried, but due to borrowing issues with the Session struct, I need to change the code structure in order to solve it.

Here is the function being called:

```rust
let (change_set, raw_events, mut extensions) = session.finish_with_extensions()?;`
```

`session.finish_with_extension(self)` is a move type, which means that session cannot be used afterwards.

```rust
pub fn finish_with_extensions(
        self,
    ) -> VMResult<(ChangeSet, Vec<Event>, NativeContextExtensions<'r>)> {
        self.session.finish_with_extensions()
    }
```

moveos/moveos/src/moveos.rs
```diff
@@ -265,9 +271,13 @@ impl MoveOS {
                         .finish(Location::Undefined)
                 })?;
 
+                let state_db = self.db.get_state_store();
+                let vm_status_explain = explain_vm_status(state_db, vm_status_cloned)?;
+
                 Ok(TransactionOutput {
                     state_root: new_state_root,
                     status,
+                    status_reexplain: vm_status_explain,
```
baichuan3, line 280: vm_status_explain contains status ?

crates/rooch-types/src/transaction/mod.rs
```diff
@@ -171,6 +173,8 @@ pub struct TransactionInfo {
     /// failures and Move abort's receive more detailed information. But other errors are generally
     /// categorized with no status code or other information
     pub status: KeptVMStatus,
+
+    pub status_reexplain: VmStatusExplainView,
```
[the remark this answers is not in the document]
steelgeek091: Okay, the field has been removed.

moveos/moveos/src/moveos.rs
```diff
@@ -39,6 +42,7 @@ pub struct TransactionOutput {
     /// The new state root after the transaction execution.
     pub state_root: H256,
     pub status: KeptVMStatus,
+    pub status_reexplain: VmStatusExplainView,
```
jolestar, line 45: VmStatusExplainView should not be part of TransactionOutput. But we can refactor in #239

jolestar (approved):

Issue #209, opened 2023-05-31
[VM & Framework] Implement the sequence number validate


jolestar:
1. Validate the sequence_number in the transaction validate function.
2. Call a function in MoveVM to increment the sequence_number in rooch_framework::account::Account.
3. Provide RPC API to get the account sequence_number, and fill the sequence_number when constructing the transaction.

jolestar:
@WGB5445 You can also consider taking this issue if @wow-sven has yet to start.

jolestar:
I start to work on this issue

PR #242, opened 2023-06-07
[moveos] Refactor MoveOS VM and Session

jolestar:
1. Refactor MoveOS, MoveVMExt, and SessionExt, to prepare #209
   * Open One Session for One TxContext
   * Extract the transaction finalizer logic, ensuring the failed transaction changes should be discarded, but the finalizer's changes must be kept.
3. Rename TransactionInfo to TransactionExecutionInfo.

#232 has a bug. Verifying the module before publishing raises an error because it tries to load the function from the session. @WGB5445

### DIFF
--- a/crates/rooch-types/src/transaction/mod.rs
+++ b/crates/rooch-types/src/transaction/mod.rs
@@ -150,11 +150,9 @@ impl AuthenticatableTransaction for TypedTransaction {
     }
 }
 
-/// `TransactionInfo` represents the result of executing a transaction.
-//TODO rename to TransactionExecutionInfo?
-// #[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize, CryptoHash, CryptoHasher)]
+/// `TransactionExecutionInfo` represents the result of executing a transaction.
 #[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
-pub struct TransactionInfo {
+pub struct TransactionExecutionInfo {
     /// The hash of this transaction.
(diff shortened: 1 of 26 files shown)


baichuan3 (approved):

crates/rooch-types/src/transaction/mod.rs
```diff
@@ -174,15 +172,15 @@ pub struct TransactionInfo {
     pub status: KeptVMStatus,
 }
 
-impl TransactionInfo {
+impl TransactionExecutionInfo {
```
baichuan3, line 175: transaction storage needs to use TransactionExecutionInfo, but TransactionExecutionInfo belongs to rooch types, consider putting TransactionExecutionInfo into moveos type, otherwise moveos will be reversely dependent on rooch.

jolestar: I also think about this. How to store the TransactionExecutionInfo and TransactionSequenceInfo.

I try to figure out a solution on the next PR.

baichuan3: from my understanding, the entry for calling Sequence and calling Transaction storage should also be in the rooch executor, right?

jolestar: Currently, the RPC service calls executor and sequencer.

jolestar:
The diff of this PR is not easy to read. I merged it first, and @wow-sven @steelgeek091  can directly view the merged code. Please provide feedback if there are any issues.

Issue #249, opened 2023-06-08
[API] The `rooch_getEventsByEventHandle` seems to have a problem


wubuku:
The tested contract project is here:

https://github.com/wubuku/rooch/tree/fe71d73d06e53e7c981bdb41d143bcf5895c222f/examples/basic_object_example

Start the local server first:

```shell
rooch server start
```

Then deploy:

```shell
rooch move publish --named-addresses rooch_examples={ACCOUNT_ADDR}
```

Then call the contract function:

```
rooch move run --function {ACCOUNT_ADDR}::something_aggregate::create_something --args 1u32 2u128 --sender-account {ACCOUNT_ADDR}
```

Can be executed successfully. According to the code logic of the contract, the create_something function will emit an event.

CURL view event handle:

```shell
curl --location --request POST 'http://localhost:50051' \
--header 'Content-Type: application/json' \
--data-raw '{
 "id":101,
 "jsonrpc":"2.0",
 "method":"rooch_executeViewFunction",
 "params":[{"function_id":"0x1::events::get_event_handle","ty_args":["{ACCOUNT_ADDR}::something::SomethingCreated"],"args":[]}]
}'
```

The output is similar to:

```json
{"jsonrpc":"2.0","result":[{"value":{"type_tag":"0x1::object_id::ObjectID","value":"0x53f32af12dc9236eb67f1c064cf55ee8891a90040f71ba17422cfdd91eb7358b"},"move_value":"0x53f32af12dc9236eb67f1c064cf55ee8891a90040f71ba17422cfdd91eb7358b"},{"value":{"type_tag":"address","value":"0x565d5717526aecec1f9d464867f7d92d6eae2dc8ca73a0dc2613dd185d3d7bc7"},"move_value":"0x565d5717526aecec1f9d464867f7d92d6eae2dc8ca73a0dc2613dd185d3d7bc7"},{"value":{"type_tag":"u64","value":"0x0100000000000000"},"move_value":"1"}],"id":101}
```

Then, take the event handle in the output, such as `0x53f32af12dc9236eb67f1c064cf55ee8891a90040f71ba17422cfdd91eb7358b`  output above, and use CURL to request event information:

```shell
curl --location --request POST 'http://localhost:50051' \
--header 'Content-Type: application/json' \
--data-raw '{
 "id":101,
 "jsonrpc":"2.0",
 "method":"rooch_getEventsByEventHandle",
[...]

PR #251, opened 2023-06-09
fix event query bug which result is null

baichuan3:
resolve https://github.com/rooch-network/rooch/issues/249


jolestar (approved):

jolestar:
The Event API needs some test. Add a cli command to test events?

Issue #245, opened 2023-06-08
[move-std] Use different addresses for move-std, moveos-std, rooch-framwork


jolestar:
Use different addresses for move-std, moveos-std, rooch-framwork:

## Motivation

1. Using different addresses can avoid module naming conflicts and also facilitate extensions. For example, `moveos-std::bcs` can wrap all methods of `move-std::bcs` and add extension methods.
2. If the ecosystem partners are willing to build some basic libraries in the future, splitting addresses is convenient for collaborative building together.

## Solution

For example:

1. move-std: 0x1
2. moveos-std: 0x2
3. rooch-framework: 0x3

Other built-in stdlib or framework, continue 0x4, 0x5 .. etc.

## Challenges

1. Multiple addresses will bring complexity to build and init Genesis.
2. Friend methods cannot be shared between addresses, which may pose challenges.

## Ref:

1. Discussion: https://discord.com/channels/1078938449974935592/1116245987598540811

PR #266, opened 2023-06-12
[stdlib] Refactor stdlib build

jolestar:
The first step of #245 

* Copy MoveStdlib to this repo
* Make stdlib build multi genesis transaction

## TODO

1. Add a `moveos-stdlib-builder` crate to hold all stdlib builder codes.
2. Move out the rooch-framework from moveos.
3. Move out the framework test casees to a new crate.
4. Use new address for moveos-stdlib and rooch-framework.


moveos/moveos-stdlib/move-stdlib/Move.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "MoveStdlib"
```
baichuan3, line 2: If move-stdlib is directly moved to moveos-stdlib, if move-stdlib needs to be upgraded later, it can only be replaced by source code?

baichuan3: And corresponding move rust code, or rely on cargo package management dependencies?

wow-sven (approved):

moveos/moveos-stdlib/move-stdlib/Move.toml
jolestar: move-stdlib in 0x1, moveos-stdlib in 0x2, there are two package. We need to find a way to support building remote package.

PR #269, opened 2023-06-12
[Testing] Test cases for runtime entry function verification.

### LINKED ISSUE #252: [Testing] Test cases for runtime entry function verification.
We should write test cases for `runtime verification of the entry function`. 
These test cases can be divided into two categories: `successful verification` and `failed verification`. 
Additionally, multiple test scenarios should be constructed for each category.

Reference the Move test cases at: https://github.com/move-language/move/tree/main/language/move-bytecode-verifier/transactional-tests/tests.

For example:
- entry_function_with_return_value.move
- entry_function_invalid_struct.move
- entry_function_valid_struct.move
- entry_function_invalid_reference.move

In the Rooch project, we can write the test code in this directory: `crates/rooch-integration-test-runner/tests/cases`.

WGB5445:
Part of #252

### DIFF
--- a/crates/rooch-framework-tests/tests/cases/entry_function/entry_function_with_return_value.move
+++ b/crates/rooch-framework-tests/tests/cases/entry_function/entry_function_with_return_value.move
@@ -1,0 +1,54 @@
+//# init --addresses creator=0x42
+
+//# publish
+module creator::test {
+
+    entry public fun test_entry_function_return_value_u8():u8{
+        0
+    }
+
+    entry public fun test_entry_function_return_value_u16():u16{
+        0
+    }
+
+    entry public fun test_entry_function_return_value_u32():u32{
+        0
+    }
+
+    entry public fun test_entry_function_return_value_u64():u64{
+        0
+    }
+
+    entry public fun test_entry_function_return_value_u128():u128{
+        0
+    }
+
+    entry public fun test_entry_function_return_value_u256():u256{
+        0
+    }
+
+    entry public fun test_entry_function_return_value_address():address{
+        @creator
+    }
+
+    entry public fun test_entry_function_return_value_String():std::string::String{
+        std::string::utf8(b"")
+    }
+
+    struct Foo has copy, drop {
+        x: u64,
+    }
+
+    entry public fun test_entry_function_return_value_struct():Foo {
+        Foo { x: 0 }
+    }
+
+    struct FooT<phantom T> has copy, drop {
+        x: u64,
+    }
+
+    entry public fun test_entry_function_return_value_T<T>(): FooT<T>{
+        FooT<T>{ x: 0 }
+    }
+
+}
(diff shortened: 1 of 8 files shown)


crates/rooch-integration-test-runner/tests/cases/entry_function_invalid_reference.move
```diff
@@ -0,0 +1,35 @@
...
+
+    entry public fun test_entry_function_invalid_reference_signer( _: & signer ){
+        
+    }
+
+    entry public fun test_entry_function_invalid_reference_mut_signer( _: &mut signer ){
+        
+    }
+
+    entry public fun test_entry_function_invalid_reference_storage_context( _: & storage_context::StorageContext ){
```
steelgeek091, line 17: Immutable references of the StorageContext type are allowed.

crates/rooch-integration-test-runner/tests/cases/entry_function_invalid_struct.move
```diff
@@ -0,0 +1,34 @@
...
+//# publish
+module creator::test {
+    use std::string;
+    use std::ascii;
+    use moveos_std::storage_context;
+    use moveos_std::object_id;
+    use moveos_std::tx_context;
+
+
+    entry public fun test_entry_function_invalid_struct_string( _str: string::String ){
```
steelgeek091, line 12: String type is allowed.

steelgeek091 (changes requested):

jolestar (commented):
Put the test cases into `rooch-framework-tests`, and we also need the negative test cases.

crates/rooch-framework-tests/tests/cases/entry_function/entry_function_with_return_value.move
```diff
@@ -0,0 +1,54 @@
+//# init --addresses creator=0x42
+
+//# publish
+module creator::test {
+
+    entry public fun test_entry_function_return_value_u8():u8{
```
jolestar, line 6: I think the entry function should not return values. @WGB5445  @steelgeek091

steelgeek091: Yes, the verifier also checks inside if the entry function has a return value, an error will be thrown. So here, the error message should be determined in the exp file.

https://github.com/rooch-network/rooch/blob/925ca13c486a6b81c8f866d2762ff5b55727ab1b/moveos/moveos-verifier/src/verifier.rs#L153

WGB5445: But there is no error reported in the actual exp file.
It should be a problem when publishing
I think this needs to be fixed in another pr

crates/rooch-integration-test-runner/tests/cases/entry_function_invalid_struct.move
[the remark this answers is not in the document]
WGB5445: > put all test case in rooch-framework-tests, do not repeat in rooch-integration-test-runner.

 Okay, I have moved all the tests to room framework tests

jolestar (approved):
LGTM, these test cases need to be refactored after the verifier resolves the script verification issue.

steelgeek091 (approved):

Issue #278, opened 2023-06-13
[ETH-Compatible] MoveCallData encoding


yubing744:

jolestar:
resolved on #280

PR #280, opened 2023-06-13
Feature owen eth call demo

yubing744:
I add a web project to the counter contract and complete the coding of the contract call

image

Ref https://github.com/rooch-network/rooch/issues/278


examples/counter/web/src/lib/rooch_types.yml
```diff
@@ -0,0 +1,94 @@
+---
```
jolestar, line 1: We can consider auto-generate rooch_types.yml in the ci.

jolestar (approved):
LGTM, @wow-sven take a look.

wow-sven (approved):

examples/counter/web/src/lib/rooch_types.yml
yubing744: I have tried to use serde-generate to generate TS, but an UnknownFormatInContainer error will appear. Through exclusion, it is determined that the indirect recursive definition of pub type_params: TypeTag in the StructTag structure causes serde_generate to report an error.

yubing744: If we use serde-generate to generate rooch_types.yaml should report the same error. Later, we need to fix the bug of serde-generate before we can generate rooch_types.yaml through CI.

yubing744:
Now there are a lot of MoveAction coding codes in this web Demo, and it feels like it needs to be refactored into rooch.js.

examples/counter/web/src/lib/rooch_types.yml
jolestar: help to record an issue.

yubing744: > help to record an issue.

OK

yubing744:
I added two subtasks:
https://github.com/rooch-network/rooch/issues/305
https://github.com/rooch-network/rooch/issues/306

Issue #45, opened 2023-04-03
AccountAbstraction (Compatible with ETH account)


yubing744:

baichuan3:
**Signature flow**
<img width="999" alt="image" src="https://user-images.githubusercontent.com/1904567/236368172-2bf59e0e-a286-43e1-b346-1d7931e7a9ac.png">

**Account Abstraction**
<img width="843" alt="image" src="https://user-images.githubusercontent.com/1904567/236368224-2ccb3ce8-f1b0-47c8-a397-a46e8e48c2e4.png">

PR #283, opened 2023-06-14
[vm] Introduce Move module binding and refactor transaction validator

### LINKED ISSUE #79: [examples] Refactor counter example
In the future, we need to disable the global storage instructions. So, we need to refactor the counter example and change the `move_to|borrow_global` to the account_storage function.

#73

### LINKED ISSUE #209: [VM & Framework] Implement the sequence number validate

jolestar:
1. Introduce Move module binding to invoke the Move function from Rust code conveniently. In the future, we can generate the binding via ABI.
2. Migrate the multi-chain address resolve and validate to `rooch-framework` via Move module binding.
3. The transaction flow:  multichain tx -> resolve(multichain address) to Move address -> construct MoveOSTransaction -> verify and validate the transaction -> execute the transaction.
4. Define the `finalize_function`, rooch define the finalize_function and pass the finalize_function to MoveOS via MoveOSConfig.
5. Migrate validate function from the account module to the transaction_validator module, validate the sequence_number in the validate function, and increment the sequence_number in the `transaction_validator::finalize` function.

resolve #209 fix #79

## Note
1. the Move integration test does not execute the validate and finalize function for writing the test case easier.
2. We must create the account on-chain before publishing the code or executing other functions. Should we auto-create the account by the first transaction?

## Ref

1. #45 
2. Discussion: https://discord.com/channels/1078938449974935592/1118434315173363823


crates/rooch-framework/src/bindings/address_mapping.rs
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) RoochNetwork
```
baichuan3, line 1: Consider removing the rust logic in the framework and put it in a separate crate, so that there is only move code in the framework?

jolestar: The `binding` is for binding a Rust code with the Move contract module. If we change the Move code, we also need to change the binding; I think putting them together is more suitable.

baichuan3:
with the ability of rust code to invoke move function, it means that there is no need to implement two codes for move and rust, which is very convenient

jolestar:
Force merge for #290.

Issue #273, opened 2023-06-13
[RPC] get TransactionExecutionInfo and TransactionSequenceInfo List API


baichuan3:
TransactionExecutionInfo  and TransactionSequenceInfo  also need a storage.

Consider introducing rooch-store crate, rooch-store and moveos-store can reuse kv storage components, and later including DA and Block are stored in rooch-store.

PR #287, opened 2023-06-14
Implements rooch store and in memory accumulator

baichuan3:
this is prepared for https://github.com/rooch-network/rooch/issues/273

1. Implements rooch storage to store data such as Transaction and Block etc;
2. Move the store implementation from the sequencer to the rooch storage;
3. Move TransactionExecutionInfo from rooch types to moveos types
4. Introduce InMemoryAccumulator and implement event accumulator root


jolestar (approved):
LGTM

PR #307, opened 2023-06-15
[CLI] Support for passing different types of parameters to entry fun

### LINKED ISSUE #217: [CLI] Support for passing different types of parameters to entry fun
The types of parameters I expect to be supported are here: #215

For primitive types, it seems that the current CLI can pass parameters like this:

```shell
--args @0xae43e34e51db9c833ab50dd9aa8b27106519e5bbfd533737306e7b69ef253647
--args 1u8
```

But we want entry fun to support some complex types, and we don't know how to pass in parameters via the CLI yet.

The method of sui is to directly support JSON in the cli, like:

```shell
sui client call --package _PACKAGE_ID_ --module day_summary_aggregate --function create \
--args u16_year_number \"string_calendar\" u8_month_is_leap u8_day_number \"string_day_time_zone\" \"string_description\" \"vector_u8_meta_data\" '["string_array_data_item1", "string_array_data_item2"]' '["vector_u8_optional_data"]' \"_DAY_SUMMARY_DAY_SUMMARY_ID_TABLE_OBJECT_ID_\" \
--gas-budget 100000
```

Also, I would like the entry fun's vector<u8> type argument to be passed in a string; but conversely, the String type argument is not allowed to be passed in a u8 vector, which I think is acceptable.

pause125:
resolve #217 

Supported types: u8, u16, u32, u64, u128, u256, bool, object_id, string, address, object_id, vector<inner_type>

**syntax**:

`address:0x1 bool:true u8:0 u256:1234 'vector<u32>:a,b,c,d'`

**address** and **uint** can be written in short form like `@0x1 1u8 4123u256`.

**Examples**:

see `examples/entry_function_arguments` and integration test `crates/testsuite/features/cmd.feature` [entry function example] scenario.


crates/rooch/src/types.rs
jolestar, line 1: rename to `cli_types.rs`

jolestar (commented):
LGTM

jolestar (approved):

baichuan3 (approved):
LGTM

Issue #303, opened 2023-06-15
[Framework] Cleanup unused modules


jolestar:
Cleanup unused modules for prepare generate framework documents

PR #309, opened 2023-06-16
cleanup unused modules of framework

baichuan3:
resolve https://github.com/rooch-network/rooch/issues/303


baichuan3:
Some modules need to be reimplemented in the next milestone, such as timestamp.move, genesis.move, chainid.move, etc.

jolestar (approved):

Issue #214, opened 2023-06-01
[Document] Framework documents


jolestar:
Generate framework document with move docs tool

baichuan3:
closed as completed via https://github.com/rooch-network/rooch/pull/317

PR #317, opened 2023-06-17
[Document] Generate Framework documents

WGB5445:
#214

### DIFF
--- a/moveos/moveos-stdlib-builder/src/lib.rs
+++ b/moveos/moveos-stdlib-builder/src/lib.rs
@@ -4,11 +4,11 @@
 use anyhow::{ensure, Result};
 use dependency_order::sort_by_dependency_order;
 use move_binary_format::CompiledModule;
+use move_cli::base::docgen::Docgen;
 use move_core_types::account_address::AccountAddress;
 use move_package::{compilation::compiled_package::CompiledPackage, BuildConfig};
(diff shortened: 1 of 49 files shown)


moveos/moveos-stdlib-builder/src/lib.rs
```diff
@@ -102,6 +103,22 @@ impl Stdlib {
                 "all modules must have same address"
             );
         }
+        let docgen = Docgen {
+            section_level_start: Option::None,
+            exclude_private_fun: false,
```
jolestar, line 108: make 

exclude_private_fun -> true
exclude_specs -> true

moveos/moveos-stdlib-builder/src/lib.rs
```diff
@@ -102,6 +103,22 @@ impl Stdlib {
...
+            exclude_specs: false,
+            independent_specs: false,
+            exclude_impl: false,
+            toc_depth: Option::None,
+            no_collapsed_sections: false,
+            output_directory: Option::None,
+            template: vec![],
```
jolestar, line 115: Add a template file

jolestar (commented):
Need to provide a method to ensure the document should be updated when the Move code changes.

jolestar (approved):

Issue #191, opened 2023-05-29
[Testing] moveos_stdlib::table


jolestar:

pause125:
I will work on this.

Issue #192, opened 2023-05-29
[Testing] moveos_stdlib::type_table


jolestar:

pause125:
I will work on this.

Issue #128, opened 2023-05-14
[MoveosStdLib] Maybe we need a "table_with_length" module


wubuku:
As discussed previously, we plan to use tables to represent the  (basic "one-to-many") relationships between entities.

When embedding a small number of instances of an entity ("non-aggregate root entities within an aggregate") as a table within another entity, it may sometimes be necessary to remove them to save resources.

This module can be like Aptos's: https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/aptos-stdlib/sources/table_with_length.move

jolestar:
How about directly adding the `length` to the table module? @baichuan3

baichuan3:
> How about directly adding the `length` to the table module? @baichuan3

table_with_length and table meet different scenarios,  my opinion is to provide them separately, and can be used in the implementation.

wubuku:
The main problem with the current table module is that it does not have a public "destroy" method, whereas table_with_length provides a more developer-friendly "destroy_empty" method.

jolestar:
`destroy_empty` can provide by the native function. If the SMT `state root` is the placeholder value, the smt is empty.

I think the `length` of a Table should be a generic requirement. 

Maintaining the length of the table may take some costs, but I think it is worth it.

PR #319, opened 2023-06-17
add unit tests for moveos_stdlib::table and  moveos_stdlib::type_table

pause125:
There are two failed cases:
- `table::test_destroy_nonempty_table`: it should aborts when destroy a non-empty table however not.
- `table::test_nested_table_destroy`: destroy the parent table of a nested table, the child table should not be accessable how not.

Please help review the test cases. @jolestar

### DIFF
--- a/moveos/moveos-stdlib/moveos-stdlib/sources/type_table.move
+++ b/moveos/moveos-stdlib/moveos-stdlib/sources/type_table.move
@@ -105,4 +105,102 @@ module moveos_std::type_table {
         raw_table::destroy(&handle)
     }
 
+    #[test_only]
+    struct TestType has key {
+        val: u64,
+    }
+
+    #[test(account = @0x1)]
+    fun test_all(account: signer) {
+        let sender = std::signer::address_of(&account);
+        let tx_context = moveos_std::tx_context::new_test_context(sender);
+        let table = new(&mut tx_context);
+
+        let t = TestType {
+            val: 1,
+        };
+
+        assert!(!contains<TestType>(&table), 1);
+        add<TestType>(&mut table, t);
+        assert!(contains<TestType>(&table), 2);
+
+        assert!(borrow<TestType>(&table).val == 1, 3);
+        borrow_mut<TestType>(&mut table).val = 2;
+        assert!(borrow<TestType>(&table).val == 2, 4);
+
+        let TestType {val:_} = remove<TestType>(&mut table);
+        assert!(!contains<TestType>(&table), 5);
+
+        drop_unchecked(table);
+    }
+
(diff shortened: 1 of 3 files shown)


jolestar:
It seam like a bug. Please help to record a bug.

pause125:
Ready for review:

fix #320 
resolve #191 #192

moveos/moveos-stdlib/src/natives/moveos_stdlib/raw_table/mod.rs
```diff
@@ -600,6 +600,11 @@ fn native_destroy_empty_box(
     let mut table_data = table_context.table_data.borrow_mut();
 
     let handle = get_table_handle(pop_arg!(args, StructRef))?;
+    if table_data.tables.contains_key(&handle)
```
jolestar, line 603: If we do not read any data from the table, the table's content will be empty, so we may still have bug #320 . We need to consider #128

moveos/moveos-stdlib/moveos-stdlib/sources/type_table.move
```diff
@@ -105,4 +105,102 @@ module moveos_std::type_table {
...
+            val: 1,
+        };
+
+        add<TestType>(&mut table, t);
+        assert!(contains<TestType>(&table), 1);
+
+        let t = TestType {
+            val: 2,
+        };
+        add<TestType>(&mut table, t);
```
jolestar, line 154: expect failure on this line?

jolestar (approved):
LGTM

But we also need some integration-test to test in a multi-transaction scenario.

moveos/moveos-stdlib/moveos-stdlib/sources/type_table.move
pause125: I think so. Is that as desired?

moveos/moveos-stdlib/src/natives/moveos_stdlib/raw_table/mod.rs
pause125: So the `Table.content.len` is not the real table length?

baichuan3 (approved):

moveos/moveos-stdlib/src/natives/moveos_stdlib/raw_table/mod.rs
jolestar: The `Talbe.content` is the runtime content that only contains the loaded table data, not all data.

jolestar:
I merge this first. The empty table issue needs to be addressed in the future.

PR #333, opened 2023-06-21
[Testing] UnitTest for rooch_framework::account  Fixes #195

### LINKED ISSUE #195: [Testing] rooch_framework::account

WGB5445:
#195

### DIFF
--- a/crates/rooch-framework/sources/account.move
+++ b/crates/rooch-framework/sources/account.move
@@ -336,4 +336,60 @@ module rooch_framework::account{
       create_account_entry(&mut ctx, sender);
       storage_context::drop_test_context(ctx);
    }
+
+   #[test(sender=@0x0)]
+   #[expected_failure(abort_code = 0x10005, location = Self)]
+   fun test_failur_entry_account_creation_reserved(sender: address){
+      let ctx = storage_context::new_test_context(sender);
+      create_account_entry(&mut ctx, sender);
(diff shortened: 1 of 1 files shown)


baichuan3 (approved):

crates/rooch-framework/sources/account.move
```diff
@@ -336,4 +336,60 @@ module rooch_framework::account{
...
+   #[test(sender=@0x0)]
+   #[expected_failure(abort_code = 0x10005, location = Self)]
+   fun test_failur_entry_account_creation_reserved(sender: address){
+      let ctx = storage_context::new_test_context(sender);
+      create_account_entry(&mut ctx, sender);
+      storage_context::drop_test_context(ctx);
+   }
+
+   #[test(sender=@0x42)]
+   #[expected_failure(abort_code = 0x1, location = moveos_std::account_storage)]
```
baichuan3, line 349: How to get the abort_code？by simulating the exception?

WGB5445: > How to get the abort_code？by simulating the exception?

In the test, the virtual machine executes this unit test. At this time, we can know in advance that this function uses the function in the error module to throw an exception, so that the abnormal value can be calculated.

This is also because the current exception is the exception handled by the move contract during execution and not the vm exception

jolestar (approved):

Issue #336, opened 2023-06-21
[`rooch move build`] command results in Success regardless failure in cmd output for private_generics syntax


feliciss:
When building private_generics.move in examples directory, we can see there's an error and also a success message in the end.

```bash
error: resource type "Data" in function "0x42::Test1::create_box" not defined in current module or not allowed
   ┌─ ./sources/private_generics.move:47:23
   │
47 │         let box_val = create_box<Data, Data, Box<u32>>(data);
   │                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

extended checks failed
Success
```

Need to fix it to ensure building alignment.

PR #337, opened 2023-06-21
Resolve the issue of returning an error message when rooch move build fails.

steelgeek091:
1. Fix: Return an error message when `rooch move build` fails. 
2. Delete the failed private_generics example in the `examples/private_generics` directory.

resolve https://github.com/rooch-network/rooch/issues/336.


examples/private_generics/Move.toml
```diff
@@ -7,9 +7,9 @@ version = "0.0.1"
 MoveosStdlib = { local = "../../moveos/moveos-stdlib/moveos-stdlib" }
 
 [addresses]
-rooch_examples = "_"
+rooch_examples = "0x42"
```
baichuan3, line 10: The example project uses placeholders, which is more convenient for testing

steelgeek091: Okay, fixed.

baichuan3 (approved):

Issue #342, opened 2023-06-22
[MoveosStd] Table need provide destroy_empty_table, is_empty and length function 


jolestar:
1. Whether to record length and where to record it. It doesn't feel right to keep the record in TableHandle. If it is recorded in TableInfo, how to get TableInfo is a problem (TableInfo is in ObjectStorage, there is a dependency cycle problem).
2. The StateResolver trait of Table only has the method of obtaining value through ObjectID(TableHandle) and key. How to express and obtain table length, or check `is_empty` is also a problem.

### Refs

1. Discussion: https://discord.com/channels/1078938449974935592/1121436866890170398

jolestar:
Two solutions:

1. Save length in `Table`, `TypeTable`, and `ObjectStorage`, like in Move's Table extension or Aptos's TableWithLength. The advantage is that it is easy to implement and can be implemented directly in Move without native or statedb support. The disadvantage is that the struct with a table as field, after modifications to the Table, the struct will also change.

2. Implement a native method to check if the Table is empty. However, this check cannot directly rely on statedb since the VM runtime does not update statedb in real-time. It requires the Table's Runtime cache and statedb to work together. In this case, we also need to save a length, which can be stored in the TableInfo. TableInfo is a global Object, currently only saving the Table SMT's state_root, which will be directly updated and created by StateDB. raw_table native does not know TableInfo. We need to refactor the raw_table runtime implemention.

pause125:
For option 1, as [this](https://github.com/rooch-network/rooch/blob/d8f75a09f80731fb006fc441fe9fbc1c40dd4499/moveos/moveos-stdlib/moveos-stdlib/sources/table.move#L28C93-L28C93) function can create a table from a given handle,  the init length may no longer be zero, there should have a way to get the length of the given handle.

PR #344, opened 2023-06-24
[MoveosStdlib] Public the `destory_empty` function in table,type_table,object_storage module

jolestar:
Part of #342 

1. Provide destory_empty in table,type_table,object_storage module
2. There is a bug in the native implementation of `destory_empty`. I give a test example.

The bug will be resolved in the future PR, based on Table's `length` solution.

### DIFF
--- a/moveos/moveos-stdlib/src/natives/moveos_stdlib/raw_table/mod.rs
+++ b/moveos/moveos-stdlib/src/natives/moveos_stdlib/raw_table/mod.rs
@@ -630,7 +630,7 @@ fn native_drop_unchecked_box(
     args: VecDeque<Value>,
 ) -> PartialVMResult<NativeResult> {
     assert_eq!(args.len(), 1);
-
+    //TODO remove the droped table from the table_data
     Ok(NativeResult::ok(gas_params.base, smallvec![]))
 }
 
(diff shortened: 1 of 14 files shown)


moveos/moveos-stdlib/src/natives/moveos_stdlib/raw_table/mod.rs
```diff
@@ -630,7 +630,7 @@ fn native_drop_unchecked_box(
     args: VecDeque<Value>,
 ) -> PartialVMResult<NativeResult> {
     assert_eq!(args.len(), 1);
-
+    //TODO remove the droped table from the table_data
```
baichuan3, line 633: The bug Is it not to check the table is empty?

baichuan3 (approved):

moveos/moveos-stdlib/src/natives/moveos_stdlib/raw_table/mod.rs
jolestar: The `drop_unchecked` is `test_only`, so it does not affect the transaction result. I tried to remove the useless `drop_unchecked` function and got that we need to remove the dropped table from the runtime TableData if we want to correct the Table runtime cache.  

Maybe using the dropped table again can produce a bug in the unit tests.

PR #348, opened 2023-06-25
Feature owen eth rooch address mapping

yubing744:
Successful debugging of calling the Move contract through MetaMask:

1. crates/rooch-framework/sources/address_mapping.move
When the Rooch address cannot be found according to ETH, use blake2b256(eth_address) to automatically generate the Rooch address
2. Three interfaces that Mock MetaMask depends on: eth_getTransactionReceipt, eth_getTransactionByHash and eth_getBlockByHash
3. Adjust the eth_getTransactionCount interface to return the real transaction number of the current account as the nonce

image

image

### DIFF
--- a/crates/rooch-framework/sources/address_mapping.move
+++ b/crates/rooch-framework/sources/address_mapping.move
@@ -5,6 +5,7 @@ module rooch_framework::address_mapping{
     use moveos_std::storage_context::{Self, StorageContext};
     use moveos_std::table::{Self, Table};
     use moveos_std::account_storage;
+    use rooch_framework::rooch_hash::{blake2b256};
 
     friend rooch_framework::transaction_validator;
 
@@ -46,10 +47,15 @@ module rooch_framework::address_mapping{
             let addr = table::borrow(&am.mapping, maddress);
             option::some(*addr)
         }else{
-            option::none<address>()
+            default_rooch_address(maddress)
         }
     }
     
+    fun default_rooch_address(maddress: MultiChainAddress): Option<address> {
+        let hash = blake2b256(&maddress.raw_address);
+        option::some(moveos_std::bcd::to_address(hash))
+    }
+
     /// Check if a multi-chain address is bound to a rooch address
     public fun exists_mapping(ctx: &StorageContext, maddress: MultiChainAddress): bool {
         if (is_rooch_address(&maddress)) {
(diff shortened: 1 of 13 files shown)


crates/rooch-framework/sources/address_mapping.move
```diff
@@ -46,10 +47,15 @@ module rooch_framework::address_mapping{
...
             option::some(*addr)
         }else{
-            option::none<address>()
+            default_rooch_address(maddress)
         }
     }
     
+    fun default_rooch_address(maddress: MultiChainAddress): Option<address> {
+        let hash = blake2b256(&maddress.raw_address);
+        option::some(moveos_std::bcd::to_address(hash))
```
jolestar, line 56: I think creating a new address in the resolve function is not a good idea because we can not ensure that the address mapping exists via the resolve function.

yubing744: An available Rooch address must be returned here, otherwise MetaMask wallet users will prompt an error when they initiate a transaction for the first time: the multiaddress sender({}) mapping record is not exists.

The idea here is to temporarily use the ETH address to obtain the Rooch address through Hash, and after the transaction is executed, complete the creation of the Rooch account and the mapping of the address in post_execute.

The reason why the ETH public key is not used as the Rooch address as previously discussed is because the ETH public key cannot be obtained in the resolve contract method. If the resolve signature is reconstructed later, it should be changed to use the ETH public key as the Rooch address. Rooch address. But I feel that this solution is not universal enough. When we need to support a new L1 public chain, we need to adapt here.

crates/rooch-server/src/server/eth_server.rs
```diff
@@ -258,12 +274,185 @@ impl EthAPIServer for EthServer {
...
+    async fn transaction_by_hash(&self, hash: H256) -> RpcResult<Option<Transaction>> {
+        let resp = self
+            .rpc_service
+            .get_transaction_by_hash(hash)
+            .await?
+            .map(Into::into);
+
+        let transaction = resp.map(|_transaction_view: TransactionView| -> Transaction {
+            Transaction {
+                hash: H256::from_str("0x7fd17d4a368fccdba4291ab121e48c96329b7dc3d027a373643fb23c20a19a3f").unwrap(),
```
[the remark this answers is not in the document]
yubing744: OK, but here's a question, is it possible to view transactions created through the Rooch API.

crates/rooch-framework/sources/address_mapping.move
jolestar: We can make this TBD. We can generate the address in the rust code. The rust code can get the ETH signature.

jolestar (approved):
Merge this PR first, make the eth address TBD

crates/rooch-framework/sources/address_mapping.move
jolestar: https://github.com/rooch-network/rooch/issues/449

Issue #350, opened 2023-06-25
Support compilation and detection of private_generics in the [rooch move integration-test] command.


steelgeek091:
Currently, the command `rooch move integration-test -p private_generics/ --named-addresses rooch_examples=0x1234` does not support the compilation and detection of `private_generics`.
It is necessary to detect the legality of using `private_generics` in the source code of the dependent projects during the compilation process of the `rooch move integration-test` command.

PR #351, opened 2023-06-25
Check if private_generics is used correctly in the integration-test.

steelgeek091:
The issue of not checking `private_generics` when executing the command `rooch move integration-test -p private_generics/ --named-addresses rooch_examples=0x1234` in the `examples/private_generics` directory has been resolved.

However, there are still some remaining issues:

1. The `datatest_stable::runner_with_opts` invocation is bypassed, and the test cases are executed directly. This means that some additional options, such as `rooch move test --list` to view the current test projects, will not be executed (but it does not affect the testing process).

2. Because the `integration-test` depends on a complete Move package, and it is itself a separate Move file, it is not possible to compile them together. Instead, the dependent Move package is compiled first, followed by the compilation of the individual Move file. Finally, the result of compiling the Move package is injected into the checking process of the individual Move file.

resolve https://github.com/rooch-network/rooch/issues/350.

### DIFF
--- a/crates/rooch/src/commands/move_cli/commands/integration_test.rs
+++ b/crates/rooch/src/commands/move_cli/commands/integration_test.rs
@@ -182,7 +182,8 @@ impl IntegrationTest {
             },
         }];
 
-        let program_res = construct_pre_compiled_lib(targets, None, move_compiler::Flags::empty())?;
+        let program_res =
+            construct_pre_compiled_lib(targets.clone(), None, move_compiler::Flags::empty())?;
         let pre_compiled_lib = match program_res {
             Ok(af) => af,
             Err((files, errors)) => {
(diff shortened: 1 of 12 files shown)


crates/rooch/src/commands/move_cli/commands/integration_test.rs
```diff
@@ -238,8 +246,8 @@ impl IntegrationTest {
             test_args.push(filter);
         }
 
-        let test_opts = datatest_stable::TestOpts::try_parse_from(test_args.as_slice())?;
```
jolestar, line 241: The `datatest_stable::TestOpt` can support filters, for example, running a specific test or running tests via prefix.

It also can run tests in multi-thread,  I think the datatest_stable is useful, we can refactor it.

steelgeek091: The issue has been fixed, and we have resumed using datatest for testing.

PR #352, opened 2023-06-25
[MoveosStdlib] draft a table length in TableInfo implementation

jolestar:
Try to keep the length in TableInfo. 

But if we want to use TableInfo, we need the global object_storage, so I put the `new_table` and `destroy_empty_table` in the StorageContext.

I will try a new native implementation.


jolestar:
Abort this approach.

jolestar:
#342

Issue #323, opened 2023-06-19
[verifier] Remove TxContext from the entry function supported argument type


jolestar:
If both `TxContext` and `StorageContext` are supported as arguments of the entry function, there are two possibilities for the mut reference returned by the entry function. 

And the `StorageContext` contains the `TxContext`, so for simplicity, the `TxContext` is discarded as the argument of the entry function, and the `StorageContext` is passed uniformly where needed.

PR #360, opened 2023-06-27
Remove TxContext from the check in the entry function.

steelgeek091:
resolve https://github.com/rooch-network/rooch/issues/323.


baichuan3:
the integration test exp file need to be update

steelgeek091:
Alright, it has been updated.

Issue #286, opened 2023-06-14
When publishing a module, verify the entry function.


steelgeek091:
When publishing a module, verify if the module has an entry function. If it exists, validate its parameters and return values to ensure compliance with the rules.

PR #379, opened 2023-06-27
check entry function at module publish

steelgeek091:
resolve https://github.com/rooch-network/rooch/issues/286.


moveos/moveos-verifier/src/verifier.rs
```diff
@@ -147,6 +148,44 @@ fn is_allowed_init_func_param(
...
+                PartialVMError::new(StatusCode::INVALID_MAIN_FUNCTION_SIGNATURE)
+                    .with_message("function should not return values".to_string())
+                    .finish(Location::Module(module.self_id())),
+            );
+        }
+
+        let func_parameters_types = module_bin_view
+            .signature_at(function_handle.parameters)
+            .0
+            .clone();
```
jolestar, line 175: Do not need `clone` here.

jolestar (approved):

moveos/moveos-verifier/src/verifier.rs
steelgeek091: Fixed.

PR #383, opened 2023-06-28
[compliance] add SPDX identifiers and scripts for adding SPDX.

### LINKED ISSUE #381: [compliance] add license identifiers to every .rs and .sh files
Like many other FOSS communities, e.g. Ethereum, we should provide SPDX IDs to every code files before releasing our code to the public.

There should be a bash script to do so, and CI to check and run the scripts.

SPDX Online Tool: https://tools.spdx.org/app/check_license/

feliciss:
resolve #381

### DIFF
--- a/crates/rooch-da/src/lib.rs
+++ b/crates/rooch-da/src/lib.rs
@@ -1,3 +1,6 @@
+// Copyright (c) RoochNetwork
+// SPDX-License-Identifier: Apache-2.0
+
 #[cfg(test)]
 mod test;
 
--- a/crates/rooch-da/src/test.rs
+++ b/crates/rooch-da/src/test.rs
@@ -1,3 +1,6 @@
+// Copyright (c) RoochNetwork
+// SPDX-License-Identifier: Apache-2.0
+
 use crate::*;
 
 struct TestDAService {}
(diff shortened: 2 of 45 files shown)


crates/rooch-open-rpc/src/generate_json_rpc_spec.rs
```diff
@@ -1,9 +1,6 @@
 // Copyright (c) RoochNetwork
 // SPDX-License-Identifier: Apache-2.0
 
-// Copyright (c) Mysten Labs, Inc.
```
jolestar, line 4: do not remote the original Copyright

jolestar (commented):
Skip the *.move file.

Because the origin move_stdlib's move file does not add the license. We follow the community.

jolestar (commented):
Can we auto-check the license in the CI?

feliciss:
> Can we auto-check the license in the CI?

I think we can. I can write a script to check with the CI.

However, I do not possess the accessibility to the CI. So may someone to pick the CI up after the script is ready.

crates/rooch-open-rpc/src/generate_json_rpc_spec.rs
feliciss: I leave it remained.

feliciss:
I may add another script with things related to the CI check. You can review this PR and merge if it looks okay.

jolestar (approved):

PR #388, opened 2023-06-29
Publish modules in Move

pause125:
Support publishing modules in Move by calling `moveos_stdlib::account_storage::publish_modules`.

TODO:

This feature has been tested by me in the local environment using the `examples/counter` module.

There is currently an issue: how to perform automated testing in CI. Since the `counter` module is compiled using my local account, the resulting module bytes can only be tested using that account. How can we compile the module based on a dynamic account in CI and then test it using the module bytes?


examples/publish_module/sources/publish_module.move
```diff
@@ -0,0 +1,17 @@
...
+        // Fixme: this module is compiled with my local account, so it may failed in other environment 
+        //  or with other signers.
+        let module_bytes: vector<u8> = x"a11ceb0b060000000b010006020608030e26043406053a32076c8c0108f8014006b802220ada02050cdf02560db5030200000101010200030c00020400000005000100000600010000070201000008030400010907080108010a09010108010b0a0b0108040605060606010708010002070801060c0106080101030107080001080002070801050107090003070801060c090002060801050106090007636f756e7465720f6163636f756e745f73746f726167650f73746f726167655f636f6e7465787407436f756e7465720e53746f72616765436f6e7465787408696e63726561736509696e6372656173655f04696e69740576616c756511676c6f62616c5f626f72726f775f6d75740e676c6f62616c5f6d6f76655f746f0d676c6f62616c5f626f72726f77ce0c46d8ef427442d6c3b05bcc4ff6d33290f8d3ef8abdf96e344466b7303a9d00000000000000000000000000000000000000000000000000000000000000020520ce0c46d8ef427442d6c3b05bcc4ff6d33290f8d3ef8abdf96e344466b7303a9d00020108030001040001030b0011010201010000050d0b00070038000c010a01100014060100000000000000160b010f0015020200000001060b000b0106000000000000000012003801020301000001060b000700380210001402000000";
```
jolestar, line 12: how about pass the module_bytes and entry function argument?

moveos/moveos-stdlib/moveos-stdlib/sources/account_storage.move
```diff
@@ -166,20 +166,26 @@ module moveos_std::account_storage {
...
+        
         while (i < len) {
-            let name = vector::pop_back(&mut module_names);
             let m = vector::pop_back(&mut modules);
-            table::add(&mut account_storage.modules, name, m);
-        }
+            vector::push_back(&mut module_bytes, move_module::module_bytes(m));
+            i = i + 1;
+        };
+        request_publish(account_address, module_bytes);
```
jolestar, line 181: Why not directly write the module bytes to the module table?

examples/publish_module/sources/publish_module.move
pause125: Entry function does not support  nested vector . But this may be a workround for a single file module.

pause125: > Entry function does not support nested vector . But this may be a workround for a single file module.

Entry function does support nested vector, but CLI does not support.

moveos/moveos-stdlib/moveos-stdlib/sources/account_storage.move
pause125: Fow now, verifying a compiled module requires a `ModuleResolver`. I havn't find a way to verify module in a native context.

jolestar: We can refactor the NativeContext. The `DataStore` in NativeContext can be a `ModuleResolver`.

pause125: OK， I will try this way.

pause125: > We can refactor the NativeContext. The `DataStore` in NativeContext can be a `ModuleResolver`.

The `ModuleResolver` need specific `Error` type. Refactoring `DataStore`  could potentially have a significant impact. This may not be a suitable way.

Issue #335, opened 2023-06-21
Support the [rooch move test] command to check private_generics.


feliciss:
rooch/examples/private_generics/sources/private_generics.move

Code:

```rust
#[test]
fun test() {
    let data = Data{ v: 123 };

    // Here, it will report that the U64 type is not allowed to appear within private_generics.
    let box_val = create_box<Data, u64, Box<u32>>(data);

    let _ = box_value(&box_val);
}
```

Command line output:

```bash
rooch move test
INCLUDING DEPENDENCY MoveStdlib
INCLUDING DEPENDENCY MoveosStdlib
BUILDING private_generics
Running Move unit tests
[ PASS    ] 0x42::Test1::test
Test result: OK. Total tests: 1; passed: 1; failed: 0
Success
```

Need to fix in the future for rooch cli `rooch move test`.

Discord:

https://discord.com/channels/1078938449974935592/1118824774886506567/1120964953088737401

PR #398, opened 2023-07-03
Compile and check private_generics during the execution of rooch move test

steelgeek091:
When executing the command `rooch move test`, check the functions decorated with the `#[test]` attribute to ensure they comply with the rules for calling `private_generics`.

resolve https://github.com/rooch-network/rooch/issues/335.

### DIFF
--- a/crates/rooch/src/commands/move_cli/commands/unit_test.rs
+++ b/crates/rooch/src/commands/move_cli/commands/unit_test.rs
@@ -2,16 +2,22 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use clap::Parser;
+use codespan_reporting::diagnostic::Severity;
 use move_cli::base::test;
+use move_command_line_common::address::NumericalAddress;
+use move_command_line_common::parser::NumberFormat;
 use move_core_types::account_address::AccountAddress;
 use move_package::BuildConfig;
 use move_unit_test::extensions::set_extension_hook;
 use move_vm_runtime::native_extensions::NativeContextExtensions;
 use moveos_stdlib::natives::moveos_stdlib::raw_table::NativeTableContext;
 use moveos_store::state_store::StateDB;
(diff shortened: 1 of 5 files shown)


crates/rooch/src/commands/move_cli/commands/unit_test.rs
```diff
@@ -36,6 +43,45 @@ impl Test {
...
+        let _ = run_extended_checks(&global_env);
+
+        if global_env.diag_count(Severity::Warning) > 0 {
+            let mut buffer = Buffer::ansi();
+            global_env.report_diag(&mut buffer, Severity::Warning);
+            let buffer_output = String::from_utf8_lossy(buffer.as_slice()).to_string();
+            let re = Regex::new("(/.*)(.move:[0-9]+:[0-9]+)").unwrap();
+            let ret = re
+                .replace(buffer_output.as_str(), "/tmp/tempfile$2".to_string())
+                .to_string();
```
jolestar, line 78: do not need to replace output file names in unit tests output.

moveos/moveos-verifier/src/build.rs
```diff
@@ -51,6 +269,34 @@ pub fn build_model(
     )
 }
 
+pub fn build_model_internal(
```
jolestar, line 272: Why `internal`? it is a public fn.

crates/rooch/src/commands/move_cli/commands/unit_test.rs
steelgeek091: That's correct, here it has been changed to not replace the content.

moveos/moveos-verifier/src/build.rs
steelgeek091: Makes sense, the function name has been changed to `build_model_with_test_attr`.

jolestar (approved):

baichuan3 (approved):

PR #403, opened 2023-07-04
update_blog_example

wubuku:
Adds an `Owner` property of type `address` to both the article and comment entities.

When creating them, this property needs to be filled in. Only their owners are allowed to update and delete them.

---

Added configuration and run instructions for off-chain service.

Since all code for the off-chain service can be regenerated by the tool, they are not submitted to this repository.


examples/blog_example/sources/article_add_comment_logic.move
```diff
@@ -15,6 +14,7 @@ module rooch_examples::article_add_comment_logic {
         comment_seq_id: u64,
         commenter: String,
         body: String,
+        owner: address,
```
jolestar, line 17: Could directly use the sender account as the owner?

wubuku: The CRUD methods are now automatically generated by the `MOVE_CRUD_IT` preprocessor set inside the dddml model document, which does not know what this Owner field means. The Create and Update methods, generated by default, require all properties to be entered by the user.

Your requirement can currently be implemented as follows:

Remove the use of the `MOVE_CRUD_IT` preprocessor from the comment entity definition in the dddml model.

Then, define a custom method in the model file like this:

```yaml
    methods.
      AddComment.
        event.
          name: CommentAdded
        parameters.
          CommentSeqId.
            type: u64
          Commenter.
            type: String
            length: 100
          Body.
            type: String
            length: 500
```

Note that the `Owner` parameter is no longer present in the method parameters above.

In the generated `article_add_comment_logic.move` file, the `verify` fun will have an empty body except for the signature part. The application developer will need to fill it in himself.

PR #429, opened 2023-07-07
Fix/cli doc

wow-sven:


crates/rooch/src/commands/event.rs
```diff
@@ -29,10 +30,14 @@ pub enum EventSubCommand {
 
 #[derive(Debug, clap::Parser)]
 pub struct GetEventsByEventHandle {
+    /// Struct name as `<ADDRESS>::<MODULE_ID>::<STRUCT_NAME><TypeParam1?, TypeParam2?>`
```
baichuan3, line 33: struct name doesn‘t contains `<TypeParam1?, TypeParam2?>`, just remove it as below
Struct name as `<ADDRESS>::<MODULE_ID>::<STRUCT_NAME>`

jolestar: StructTag can contain TypeParam. And `MODULE_ID = ADDRESS + MODULE_NAME`.

`ADDRESS::MODULE_NAME::STRUCT_NAME<TypeParam1?, TypeParam2?>`

 Remove the `<` to avoid the developer to confuse with TypeParam's '<'.

crates/rooch/src/commands/event.rs
```diff
@@ -29,10 +30,14 @@ pub enum EventSubCommand {
...
+    /// Example: `0x123::event_test::WithdrawEvent --cursor 0 --limit 1`
     #[clap(long = "event_handle_type")]
     event_handle_type: StructTag,
+    /// start position
     #[clap(long)]
     cursor: Option<u64>,
+    /// end position
```
baichuan3, line 40: In general, `end position` is equal to `cursor + limit`.
change `end position` to ` Max number of items returned per page`

crates/rooch/src/commands/transaction/commands/get_tx_by_hash.rs
```diff
@@ -6,13 +6,13 @@ use async_trait::async_trait;
 use rooch_rpc_api::jsonrpc_types::TransactionView;
 use rooch_types::{error::RoochResult, H256};
 
+/// Get transaction by hash
 #[derive(Debug, clap::Parser)]
 pub struct GetByHashCommand {
+    /// H56
```
baichuan3, line 12: H56 -> H256 ？

jolestar (approved):

PR #439, opened 2023-07-09
test before release

yourmoonlight:


jolestar:
This test should be in the "Build Docker" or Check-Build-Test workflow. We should ensure every PR does not break the release.

yourmoonlight:
> This test should be in the "Build Docker" or Check-Build-Test workflow. We should ensure every PR does not break the release.

Ok, this draft pr will be closed.

Issue #452, opened 2023-07-11
[integration-test] Support importing the stdlib and complete third-party Move libraries.


steelgeek091:
For example, executing the command `rooch move integration-test` in the `examples/counter` directory will locate and run the Move files in the `integration-test` directory. 

We will support importing the `stdlib` and `third-party Move libraries` in the `examples/counter` Move Package.

PR #464, opened 2023-07-12
Refactor the rooch move integration-test

steelgeek091:
Refactor the integration-test to allow it to import moveos_stdlib and other third-party Move Packages.

resolve https://github.com/rooch-network/rooch/issues/452.

### DIFF
--- a/examples/counter/sources/counter.move
+++ b/examples/counter/sources/counter.move
@@ -6,8 +6,12 @@ module rooch_examples::counter {
       value:u64,
    }
 
-   fun init(ctx: &mut StorageContext, account: &signer){
-      account_storage::global_move_to(ctx, account, Counter{value:0});
+   public fun init_for_test(ctx: &mut StorageContext, account: &signer) {
+      account_storage::global_move_to(ctx, account, Counter { value: 0 });
+   }
+
+   fun init(ctx: &mut StorageContext, account: &signer) {
+      account_storage::global_move_to(ctx, account, Counter { value: 0 });
    }
 
    public fun increase_(ctx: &mut StorageContext) {
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -185,39 +185,39 @@ rust_secp256k1 = { version = "0.27.0", package = "secp256k1", features = ["recov
 
 # Note: the BEGIN and END comments below are required for external tooling. Do not remove.
 # BEGIN MOVE DEPENDENCIES
-move-abigen = { git = "https://github.com/rooch-network/move", rev = "5597bcfd78d952f428360ba2eae81f2f79d2627c" }
-move-binary-format = { git = "https://github.com/rooch-network/move", rev = "5597bcfd78d952f428360ba2eae81f2f79d2627c" }
-move-bytecode-verifier = { git = "https://github.com/rooch-network/move", rev = "5597bcfd78d952f428360ba2eae81f2f79d2627c" }
-move-bytecode-utils = { git = "https://github.com/rooch-network/move", rev = "5597bcfd78d952f428360ba2eae81f2f79d2627c" }
-move-cli = { git = "https://github.com/rooch-network/move", rev = "5597bcfd78d952f428360ba2eae81f2f79d2627c" }
-move-command-line-common = { git = "https://github.com/rooch-network/move", rev = "5597bcfd78d952f428360ba2eae81f2f79d2627c" }
-move-compiler ={ git = "https://github.com/rooch-network/move", rev = "5597bcfd78d952f428360ba2eae81f2f79d2627c" }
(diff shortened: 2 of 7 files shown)


examples/counter/sources/counter.move
```diff
@@ -6,8 +6,7 @@ module rooch_examples::counter {
       value:u64,
    }
 
-   fun init(ctx: &mut StorageContext, account: &signer){
-      account_storage::global_move_to(ctx, account, Counter{value:0});
```
jolestar, line 10: Why remove the `init` function?

jolestar (commented):
Run the `integration-test` command for the examples project in CI to ensure the integration-test command is working.

examples/counter/sources/counter.move
steelgeek091: Okay, I will put back the previously deleted `init` function.

Another question is... I'm not sure if it is possible to write like this in the move file under the integration-test directory:

```move
script {
    use moveos_std::storage_context::StorageContext;
    use rooch_examples::counter;

    fun main(ctx: &mut StorageContext, sender: &signer) {
        counter::init(ctx, sender);
    }
}
```

Is it possible to automatically fill in `ctx: &mut StorageContext` inside the script section?

steelgeek091: I compiled the Move code according to the example above, but I received an error during compilation:

```move
Error: error: undeclared `0x2::storage_context::StorageContext`
  ┌─ /tmp/tempfile:8:26
  │
8 │     fun test1(_ctx: &mut StorageContext, sender: &signer) {
  │                          ^^^^^^^^^^^^^^
```

I found the reason: in the source code of the Move Compiler, when compiling the `Script` into `ModuleEnv` using `ModelBuilder`, it cannot find `StorageContext` in `struct_table`.
This seems to be a problem, I will investigate further.

steelgeek091:
> Run the `integration-test` command for the examples project in CI to ensure the integration-test command is working.

How can it be enabled in CI checks?

jolestar:
@yourmoonlight can help

examples/counter/sources/counter.move
jolestar: > Okay, I will put back the previously deleted `init` function.
> 
> Another question is... I'm not sure if it is possible to write like this in the move file under the integration-test directory:
> 
> ```move
> script {
>     use moveos_std::storage_context::StorageContext;
>     use rooch_examples::counter;
> 
>     fun main(ctx: &mut StorageContext, sender: &signer) {
>         counter::init(ctx, sender);
>     }
> }
> ```
> 
> Is it possible to automatically fill in `ctx: &mut StorageContext` inside the script section?

The code should be worked.

jolestar: Maybe, we should publish the modules of the project first, then run the integration-test?

steelgeek091: The reason for this issue is that during the compilation of `private_generics`, `build_model` is required, but the compiled `FullyCompiledProgram` is not imported as a dependency, resulting in an error.

It is not related to compiling the `Move Package` as a dependency or the subsequent `integration-test` errors.

Furthermore, since `build_model` does not support directly importing `FullyCompiledProgram` as a dependency, the results of the `Expansion` type in `FullyCompiledProgram` need to be manually imported into `ModelBuilder.struct_table` to resolve the issue.

I suggest merging this PR first because currently, as long as types from `stdlib` are not imported in the Script (importing functions is supported), there won't be any issues.

jolestar: Because the `init` is private, we can not invoke it in the script. Or add another public `init` function for the integration tests to call.

jolestar: Or `init_for_test`

steelgeek091: In the file `examples/counter/integration-tests/counter.move`, there is a call to the function `counter::init_`.

https://github.com/rooch-network/rooch/blob/ca3ef47d749663e030e55bb5b4f0c38d7cba6a0e/examples/counter/integration-tests/counter.move#L1-L11

However, due to the reason mentioned earlier, we currently cannot directly import the `&mut StorageContext` type in the Script.

yourmoonlight:
> @yourmoonlight can help

Ok, when this PR is ready, I'll add move integration-test to CI workflow

examples/counter/sources/counter.move
```diff
@@ -6,8 +6,12 @@ module rooch_examples::counter {
...
+   public fun init_for_test(ctx: &mut StorageContext, account: &signer) {
```
jolestar, line 9: Is it possible to make the `init_for_test` `#[test-only]` in the future?

jolestar (approved):

examples/counter/sources/counter.move
steelgeek091: Yes, we can add `#[test_only]` to prevent a function from being compiled.

By passing a special flag during the construction of the compiler in the Parse phase, this can be resolved. 

I will add this feature in the next PR.

PR #490, opened 2023-07-17
[scripts] add ubuntu installation script for rooch dev environment

geometryolife:


jolestar:
@yourmoonlight  

(Check-Build-Test)[https://github.com/rooch-network/rooch/blob/main/.github/workflows/check_build_test.yml] should not ignore the `scripts` dir.

The `paths-ignore` files still require the `Check-Build-Test` to pass.

Issue #487, opened 2023-07-17
[MoveosStd] Define and decode MoveAction in Move


jolestar:
This is part of #46,#460.

If we want to define some strategy to handle MoveAction, we need to decode MoveAction in Move.

Use case:

1. Define an allowlist auth validator. Only the module in the allowlist can be executed.
2. A function return MoveAction like a dynamic invocation to the function. #486

jolestar:
It is too hard to define MoveAction in Move.

One challenge is we cannot define TypeTag in Move because Move does not support the Enum type.

If we use another way to define TypeTag in Move, like https://github.com/move-language/move/pull/971, we can not keep the BSC binary of the type consistent between the Move and Rust.

So, close this issue, wait to Move support the Enum type.

PR #495, opened 2023-07-17
[MoveosStd] Define MoveAction in Move

jolestar:
A draft implementation for #487 

We can not define a `MoveAction` with the same bcs layout as the MoveAction in Rust.

So we need to provide a native function to decode the Rust's `MoveFunction` to Move's `MoveFunction`.

The PR https://github.com/move-language/move/pull/971 try to implement `StructTag` have the same challenges.


wow-sven (approved):

jolestar:
Use a simple TxMeta solution.

PR #515, opened 2023-07-19
Rooch dashboard scaffold

wow-sven:
We have explored other frameworks, such as Vue and React, and ultimately decided to use the same architecture as the Rooch Network website.

Advantage 1: This makes it easier to share some of the available components.(Lower maintenance threshold.)
Advantage 2: Ensures style consistency.

Preview
image


jolestar:
Is it possible to embed the dashboard in rooch binary?

jolestar (approved):

Issue #432, opened 2023-07-07
[Storage] implement Event storage


baichuan3:
Implement Event storage based on raw storage

baichuan3:
close as completed via https://github.com/rooch-network/rooch/pull/454 and https://github.com/rooch-network/rooch/pull/523

Issue #157, opened 2023-05-18
[Storage] implement storage


baichuan3:
support a node storage for statedb, eventdb, accumulatordb and other db.

- [x] Implements raw store component #431
- [x] State storage #9 #1
- [x] Event storage https://github.com/rooch-network/rooch/issues/432
- [ ] Accumulator storage #412

baichuan3:
I will close this issue first, the `accumulator storage` implementation is tracked as a separate issue

Issue #9, opened 2023-03-16
[statedb] Implement a node storage for StateDB


jolestar:
Currently, the node storage only implements an InMemoryStore. We need to provide a simple disk store first.

1. Introduce an embedded database, such as Rocksdb.
Or 2. Just using the file system to store the state tree's nodes. 

TBD

1. Should we split the SMT's leaf value from the node?

feliciss:
> 1. Introduce an embedded database, such as Rocksdb.

Better to look at https://github.com/holepunchto/hypercore?

Its Rust implementation: https://github.com/datrs/hypercore.

jolestar:
> > 1. Introduce an embedded database, such as Rocksdb.
> 
> Better to look for at https://github.com/holepunchto/hypercore?
> 
> Its Rust implementation: https://github.com/datrs/hypercore.

Has Hyperbee sufficient maturity?

feliciss:
> > > 1. Introduce an embedded database, such as Rocksdb.
> > 
> > 
> > Better to look for at https://github.com/holepunchto/hypercore?
> > Its Rust implementation: https://github.com/datrs/hypercore.
> 
> Has Hyperbee sufficient maturity?

No. I think it's a quite new p2p technology that emerged in 2020, mentioning its original research: https://github.com/hypercore-protocol/p2p-indexing-and-search.

There needs to be more precautious attentions to this tech. I recommend these readings:

1. https://gist.github.com/RangerMauve/ae271204054b62d9a649d70b7d218191
2. https://blog.hyper.io/p2p-databases-hyperbee/

Hyperbee is based on JavaScript, and from my research, there isn't a library for Rust yet.

baichuan3:
close as completed via https://github.com/rooch-network/rooch/pull/454 and https://github.com/rooch-network/rooch/pull/523

Issue #1, opened 2023-01-29
statedb design and implement


jolestar:
- [x] smt library: implement a standalone smt library.
- [x] two-level smt state tree design and implement.
- [x] provide a mock storage
- [x] embed a local database(rocksdb?)

wubuku:
RocksDB seems good, and there was a project named rocksplicator that implemented RocksDB's replication feature at https://github.com/pinterest/rocksplicator. 

But unfortunately, the project has been archived and there doesn't seem to be a successor.:(

feliciss:
Hyperbee is like a local database solution: https://docs.holepunch.to/building-blocks/hyperbee.

baichuan3:
close as completed via https://github.com/rooch-network/rooch/pull/454 and https://github.com/rooch-network/rooch/pull/523

PR #523, opened 2023-07-20
implements storage base on Rocksdb

baichuan3:
part of https://github.com/rooch-network/rooch/issues/157, resolve storage for both `moveos store` and `rooch store`:
- State storage https://github.com/rooch-network/rooch/issues/9 https://github.com/rooch-network/rooch/issues/1
- Event storage https://github.com/rooch-network/rooch/issues/432
- Transaction storage

And also provide a mock storage for test scenario, include integration test, testsuite, unittest.


baichuan3:
After restarting the service, some problems were encountered in reloading modules and resources from statedb, which are still being resolved.

jolestar (approved):

jolestar:
> After restarting the service, some problems were encountered in reloading modules and resources from statedb, which are still being resolved.
> 
> Eventdb can be loaded normally after restarting, only statedb has some problems,  help have a look @jolestar 

Need record the latest state root with a specific key, and init the state db with the state root

PR #541, opened 2023-07-27
[rtfsc] add storage_context module manual

geometryolife:


docs/website/pages/docs/developer/_meta.en-US.json
```diff
@@ -1,4 +1,5 @@
 {
   "installation": "Installation",
-  "rooch-cli": "Rooch CLI"
+  "rooch-cli": "Rooch CLI",
+  "rtfsc": "RTFSC"
```
jolestar, line 4: RTFSC means？ We need a document group to introduce stdlib and framework. This document should be part of it.

Issue #493, opened 2023-07-17
[Document] Move on Rooch


jolestar:
A document to introduce Move on Rooch.

* Introduce StorageAbstraction(part from #397)
* Introduce MoveosStd and RoochFramework
* Introduce AccountAbstraction(part from #468)

PR #552, opened 2023-08-01
[gh-548] add rooch account update command.

### LINKED ISSUE #548: [cli] associate auth key to ed25519 generated addresses
As discussed in the last meeting, we select simpler implementation of generating ed25519 addresses and bind/update custom auth keys to that addresses.

Here are two scenarios:

1. Only update auth keys to the addresses from built-in crypto schemes.
2. Update any auth keys to the addresses, regardless of whether it's built-in or not.

I prefer the first scenario.

And there're two possible implementations. One is to let the BTreeMap to have multiple key pairs recognized by schemes. For example, we can rename keys to:

```rs
keys: BTreeMap<RoochAddress, BTreeMap<BuiltinScheme, RoochKeyPair>>,
```

And use the first key pair as init auth key for ed25519 scheme as default.

The output is like:

```zsh
                     Rooch Address (Ed25519)                       |              Public Key (Base64)              | Scheme  | Active
--------------------------------------------------------------------------------------------------------------------------------------
0xcb48f40a8bd650a11f4a796ea81ac8f016267388a08da9875d67502650d73302 | AJEDPdj8miWTVzp7gfhO95BooxXmzTKVLK7I35KCVMcD  | ed25519 |  True 
0xcb48f40a8bd650a11f4a796ea81ac8f016267388a08da9875d67502650d73302 | A5QfCiKIICuty3IIkFqKX3KtcBYWdUNnUMfmDuVN2VK4  | schnorr |  True
```

The second option is to let the keys update every time we want to update them and replace the existing key pair to a new key pair, while remaining `address` and `secret phrase` unchanged.

For instance, we have such a command `rooch account update`:

```zsh
rooch-account-update 
Update an address with a new keypair from scheme to rooch.keystore

USAGE:
    rooch account update [OPTIONS] --address <ADDRESS> <MNEMONIC_PHRASE>
```

We take address from clear text, select a scheme, either ed25519, ecdsa or schnorr, etc. and generate new key pairs from an input mnemonic phrase.

This will result like a one-time key pair from the above example if we update it:

```zsh
                     Rooch Address (Ed25519)                       |              Public Key (Base64)              | Scheme  | Active
--------------------------------------------------------------------------------------------------------------------------------------
0xcb48f40a8bd650a11f4a796ea81ac8f016267388a08da9875d67502650d73302 | A5QfCiKIICuty3IIkFqKX3KtcBYWdUNnUMfmDuVN2VK4  | schnorr |  True
```

Which implementation do you prefer?

Update: I selected the first implementation.

Referential issue:

https://github.com/rooch-network/rooch/issues/540

Part of #558

feliciss:
Resolves #548.


crates/rooch-framework/sources/account.move
```diff
@@ -91,6 +96,30 @@ module rooch_framework::account{
...
+         account_storage::global_exists<Account>(ctx, existing_address),
+         error::not_found(EAccountNotExist)
+      ); 
+
+      update_account_unchecked(ctx, existing_address)
+   }
+
+   // TODO: update account should handle logics like installing auth validator with auth register
+   // for non-built-in custom validator scheme
+   fun update_account_unchecked(ctx: &mut StorageContext, existing_address: address): signer {
```
jolestar, line 114: We do not need a `update_account` function to install auth validator.`account_authentication::install_auth_validator` can install the validator, and the validator should call `account_authentication::rotate_authentication_key` to update the auth key.

feliciss: So I could call install_auth_validator to register an auth validator if it doesn't exist in BuiltInScheme?

What if I call with register with an existing BuiltInScheme? Will it result EAuthValidatorAlreadyInstalled?

From my understanding, all built in validators are registered in genesis init, so there's no need to handle updates for  BuiltInSchemes, and I could remove `.sign_and_execute` for account updates?

feliciss: If I replace `.sign_and_execute` with Ok(), it will not validate the key pairs added from this command.

jolestar: The BuiltInScheme does not need to register or install_auth_validator. Just call `rotate_authentication_key` to update the user's auth key for the validator.

feliciss: Okay. I'll learn how to do it.

For example, how to pass the private_generics of ValidatorType? Pass through function_name like `Identifier::new("rotate_authentication_key<ValidatorType>").unwrap()` and define a ValidatorType in the same Move module?

jolestar: Yes, like `rooch_framework::ed25519_validator::Ed25519Validator`

feliciss: Got it.

feliciss: I've added rotate auth key function for update.rs.

crates/rooch-framework/sources/account_authentication.move
```diff
@@ -37,6 +37,12 @@ module rooch_framework::account_authentication{
       }
    }
 
+   #[private_generics(ValidatorType)]
```
jolestar, line 40: Entry function can not limit the `ValidatorType` via private_generics

jolestar: The entry function of the rotate_authentication_key should be in the validator module; the validator maybe needs to do some checks before updating the authentication key.

crates/rooch/src/commands/account/commands/update.rs
```diff
@@ -0,0 +1,140 @@
...
+                    BuiltinScheme::Schnorr => Ok(Box::new(StructTag {
+                        address: SchnorrValidator::MODULE_ADDRESS,
+                        module: Identifier::new(SchnorrValidator::MODULE_NAME.to_string()).unwrap(),
+                        name: Identifier::new(String::from(stringify!(SchnorrValidator))).unwrap(),
+                        type_params: vec![],
+                    })),
+                    _ => Err(RoochError::CommandArgumentError(
+                        "Validator for this scheme is not implemented".to_owned(),
+                    )),
+                }?;
```
jolestar, line 105: This code can be a function of BuiltinScheme

crates/rooch-framework/sources/account_authentication.move
jolestar: If we want to provide a rotate_authentication_key_entry function to end-users, we can provide a private function `rotate_authentication_key_internal`, and no `private_generics` limit. the `rotate_authentication_key` and `rotate_authentication_key_entry` call the rotate_authentication_key_internal function.

crates/rooch-framework/sources/account_authentication.move
```diff
@@ -37,6 +37,12 @@ module rooch_framework::account_authentication{
...
+   /// Entry function of rotate_authentication_key. Do we need to limit the scope using private_generics?
+   public entry fun rotate_authentication_key_entry<ValidatorType>(ctx: &mut StorageContext, account: &signer, new_auth_key: vector<u8>) {
+      rotate_authentication_key<ValidatorType>(ctx, account, new_auth_key);
+   }
```
jolestar, line 44: @steelgeek091 This call should be failed when compiled. The private generics Type can not pass through.

crates/rooch-framework/sources/account_authentication.move
feliciss: Understand.

crates/rooch/src/commands/account/commands/update.rs
feliciss: Will refactor it.

crates/rooch-framework/sources/account_authentication.move
feliciss: > If we want to provide a rotate_authentication_key_entry function to end-users, we can provide a private function `rotate_authentication_key_internal`, and no `private_generics` limit. the `rotate_authentication_key` and `rotate_authentication_key_entry` call the rotate_authentication_key_internal function.

Because it will be a warning if calling it a normal function `rotate_authentication_key`:
```zsh
2023-08-03T22:50:32.253423Z  WARN connection{remote_addr=127.0.0.1:57018 conn_id=1}: moveos::moveos: execute tx(0x2ad8…406c) error: Err(VMError { major_status: EXECUTE_ENTRY_FUNCTION_CALLED_ON_NON_ENTRY_FUNCTION, sub_status: None, message: None, exec_state: None, location: Module(ModuleId { address: 0000000000000000000000000000000000000000000000000000000000000003, name: Identifier("account_authentication") }), indices: [(FunctionDefinition, 4)], offsets: [] })    
```

feliciss (commented):
See comment below.

crates/rooch/src/commands/account/commands/update.rs
```diff
@@ -0,0 +1,136 @@
...
+                println!(
+                    "{}",
+                    AccountAddress::from(existing_address).to_hex_literal()
+                );
+                println!(
+                    "Generated a new keypair for an existing address on scheme {:?} [{existing_address}]",
+                    scheme.to_owned()
+                );
+
+                let address = match scheme {
```
feliciss, line 84: I've updated it. However, due to cyclic import dependencies between `rooch-types` and `rooch-framework`, I couldn't import Validators' address and module into BuiltinScheme.

crates/rooch/src/commands/account/commands/update.rs
```diff
@@ -0,0 +1,136 @@
...
+                let module_name = match scheme {
+                    BuiltinScheme::Ed25519 => Ed25519Validator::MODULE_NAME,
+                    BuiltinScheme::MultiEd25519 => todo!(),
+                    BuiltinScheme::Ecdsa => EcdsaK1Validator::MODULE_NAME,
+                    BuiltinScheme::EcdsaRecoverable => EcdsaK1RecoverableValidator::MODULE_NAME,
+                    BuiltinScheme::Schnorr => SchnorrValidator::MODULE_NAME,
+                };
+
+                let validator_struct_arg: Box<StructTag> =
+                    scheme.create_validator_struct_tag(address, module_name.to_string())?;
```
feliciss, line 101: I've updated it. However, due to cyclic import dependencies between `rooch-types` and `rooch-framework`, I couldn't import Validators' address and module into BuiltinScheme.

crates/rooch/src/commands/account/commands/update.rs
```diff
@@ -0,0 +1,140 @@
...
+                        EcdsaK1RecoverableValidator::MODULE_NAME,
+                    ),
+                    BuiltinScheme::Schnorr => (
+                        SchnorrValidator::MODULE_ADDRESS,
+                        SchnorrValidator::MODULE_NAME,
+                    ),
+                };
+
+                let validator_struct_arg: Box<StructTag> =
+                    scheme.create_validator_struct_tag(address, module_name.to_string())?;
```
feliciss, line 105: I've updated it. Now it looks more elegant. However, due to cyclic import dependencies between rooch-types and rooch-framework, I couldn't refactor Validators' address and module into BuiltinScheme.

crates/rooch-framework/sources/account_authentication.move
feliciss: I've done these:

1. Removed ValidatorType for entry function.
2. Added checks and moved rotate_authentication_key_entry to validator modules.
3. `rotate_authentication_key_internal` is a public(friend) function in account_authentication.move, and I could not call it from other crates since it has restrictions. So I made `rotate_authentication_key_entry` to call `rotate_authentication_key` public function.

feliciss:
This commit is ready for merge. However, I've noticed there's a warning NUMBER_OF_ARGUMENTS_MISMATCH. I couldn't debug with it since the message of the warning is useless.

```zsh
2023-08-05T03:10:42.602539Z  WARN connection{remote_addr=127.0.0.1:61755 conn_id=3}: moveos::moveos: execute tx(0xa123…81b4) error: Err(VMError { major_status: NUMBER_OF_ARGUMENTS_MISMATCH, sub_status: None, message: Some("argument length mismatch: expected 3 got 4"), exec_state: None, location: Undefined, indices: [], offsets: [] }) 
```

crates/rooch/src/commands/account/commands/update.rs
```diff
@@ -0,0 +1,144 @@
...
+                        address,
+                        module_name.as_str(),
+                        "rotate_authentication_key_entry",
+                    )
+                })
+                .clone();
+                let action = MoveAction::new_function_call(
+                    rotate_authentication_key_entry_function,
+                    vec![TypeTag::Struct(validator_struct_arg)],
+                    vec![signer, public_key_bytes_vec],
```
jolestar, line 118: The `signer` argument will be auto-filled by the VM and does not need to pass the signer argument. This is why the transaction executed failed by the argument length error.

crates/rooch/src/commands/account/commands/update.rs
```diff
@@ -0,0 +1,144 @@
...
+                    ),
+                };
+
+                let validator_struct_arg: Box<StructTag> =
+                    scheme.create_validator_struct_tag(address, module_name.to_string())?;
+
+                let signer = bcs::to_bytes(&existing_address).unwrap();
+                let public_key_bytes_vec = public_key.as_ref().to_vec();
+
+                let rotate_authentication_key_entry_function = Lazy::new(|| {
```
jolestar, line 107: The Lazy is useless here, directly new the function.

jolestar (changes requested):

crates/rooch/src/commands/account/commands/update.rs
jolestar: We need add this to the document of move on rooch #493

crates/rooch/src/commands/account/commands/update.rs
```diff
@@ -0,0 +1,144 @@
...
+                    )
+                })
+                .clone();
+                let action = MoveAction::new_function_call(
+                    rotate_authentication_key_entry_function,
+                    vec![TypeTag::Struct(validator_struct_arg)],
+                    vec![signer, public_key_bytes_vec],
+                );
+
+                context
```
jolestar, line 121: Need to handle the result, if this transaction failed, should return the error.

feliciss: I'll do.

crates/rooch/src/commands/account/commands/update.rs
feliciss: Will do.

crates/rooch/src/commands/account/commands/update.rs
feliciss: > The `signer` argument will be auto-filled by the VM and does not need to pass the signer argument. This is why the transaction executed failed by the argument length error.

I see.

feliciss: > We need add this to the document of move on rooch #493

I will study.

crates/rooch-framework/sources/auth_validator/ed25519_validator.move
```diff
@@ -79,10 +103,13 @@ module rooch_framework::ed25519_validator {
             auth_validator::error_invalid_account_auth_key()
         );
```
feliciss, line 104: Please help refactor the auth key checking mechanism. Right now it's only working for Ed25519 scheme since the address is generated from Ed25519. Need to consider other schemes that only have public and private keys.

feliciss: We have to update other auth key schemes `schnorr`, `ecdsa`, `ecdsa-recoverable` once to the account storage slots:

```move
let authentication_key = AuthenticationKey<ValidatorType> {
  authentication_key: new_auth_key,
};
account_storage::global_move_to(ctx, account, authentication_key);
```

Then we can check the auth key in the slot under the scenario like:

```move
let auth_key_option = account_authentication::get_authentication_key<Ed25519Validator>(ctx, addr);
if(option::is_some(&auth_key_option)){
   option::extract(&mut auth_key_option)
}
```

jolestar (approved):

wow-sven (approved):

crates/rooch-framework/sources/account_authentication.move
jolestar: Record an issue #578

Issue #550, opened 2023-07-31
[move-explain] improve custom abort code.


feliciss:
This PR https://github.com/rooch-network/rooch/pull/538 resolves abort codes well for the in-bound errors at Rooch, for example:

```zsh
rooch move explain -l 0000000000000000000000000000000000000000000000000000000000000003::ecdsa_k1 -a 1
```

Output:

```zsh
Category:
  Name: EInvalidSignature
  Description:  Error if the signature is invalid.
Success
```

This abort code is defined in `ecdsa_k1.move`: 

```zsh
const EInvalidSignature: u64 = 1;
```

This works well, but if we have another custom error 66538:

```zsh
thread 'tests::ecdsa_k1_validator_tests::test_validate' panicked at 'called `Result::unwrap()` on an `Err` value: VMError with status ABORTED with sub status 66538 at location Module ModuleId { address: 0000000000000000000000000000000000000000000000000000000000000003, name: Identifier("ecdsa_k1_validator") } and message 0x0000000000000000000000000000000000000000000000000000000000000003::ecdsa_k1_validator::validate at offset 17 at code offset 17 in function definition 10'
```

And if we call the location with the abort code:

```zsh
rooch move explain -l 0000000000000000000000000000000000000000000000000000000000000003::ecdsa_k1_validator -a 66538
```

Will result an unexpected result:

```zsh
Unable to find a description for 0000000000000000000000000000000000000000000000000000000000000003::ecdsa_k1_validator::66538
Success
```

I think custom errors should also be handled? @steelgeek091 

The cause might be extending signature length that the payload doesn't support.

The payload defaults 64-byte signature, but in ECDSA we have 65-byte recoverable signature: https://github.com/rooch-network/rooch/pull/551/files#diff-a813ae75dd946c80da62e40d5cf6797d34629e913dbc1350bee6258788935067R113-R114. @jolestar

jolestar:
`rooch move explain -a 524289 -l 0x3::account` raise an error.

jolestar:
After #602 

```
rooch move explain -a 524289 -l 0x3::account
Category Code: 8
Category Name: ALREADY_EXISTS
Reason Code: 1
Reason Name: EAccountAlreadyExists
Code Description:  Account already exists
```

```
rooch move explain -a 1 -l 0x3::account
Category Code: 0
Category Name: Unknown
Reason Code: 1
Reason Name: EAccountAlreadyExists
Code Description:  Account already exists
```

PR #553, opened 2023-08-01
Extract the category from the abort code.

steelgeek091:
#resolve https://github.com/rooch-network/rooch/issues/550.

### DIFF
--- a/crates/rooch/src/commands/move_cli/commands/explain.rs
+++ b/crates/rooch/src/commands/move_cli/commands/explain.rs
@@ -7,11 +7,13 @@ use move_core_types::account_address::AccountAddress;
 use move_core_types::errmap::{ErrorDescription, ErrorMapping};
 use move_core_types::identifier::Identifier;
 use move_core_types::language_storage::ModuleId;
+use move_core_types::vm_status::AbortLocation;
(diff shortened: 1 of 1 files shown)


baichuan3 (approved):

crates/rooch/src/commands/move_cli/commands/explain.rs
```diff
@@ -55,18 +57,25 @@ impl Explain {
...
+                    None => {
+                        return Err(anyhow::Error::msg(format!(
+                            "Unable to find a description for {}::{}",
+                            self.location, self.abort_code
+                        )));
+                    }
+                    Some(error_desc) => println!(
+                        "Category:\n  Name: {}\n  Description: {}",
+                        error_desc.reason_name.unwrap(),
+                        error_desc.code_description.unwrap(),
```
jolestar, line 73: the description can be None

jolestar (approved):

crates/rooch/src/commands/move_cli/commands/explain.rs
steelgeek091: "description" is the comment for the abort code, but it seems like something is missing.

PR #554, opened 2023-08-01
Use TableData as MoveosDataCache backend

### LINKED ISSUE #399: [MoveVM] Refactor TransactionCache and NativeTableContext's Cache
The global storage instructions use `TransactionCache` in session, and AccountStorage uses the cache in the raw table NativeTableContext.

We will deprecate Move global storage instructions in #248. But the module publishes still using the session's TransactionCache.

We need to refactor the session's TransactionCache and let the session and NativeTableContext share the same runtime cache.

pause125:
Trying to use `TableData` as MoveosDataCache. resolve #399

### DIFF
--- a/crates/rooch-integration-test-runner/src/lib.rs
+++ b/crates/rooch-integration-test-runner/src/lib.rs
@@ -4,12 +4,14 @@
 use clap::Parser;
 use codespan_reporting::diagnostic::Severity;
 use codespan_reporting::term::termcolor::Buffer;
-use move_command_line_common::address::NumericalAddress;
-use move_command_line_common::files::verify_and_create_named_address_mapping;
-use move_command_line_common::{address::ParsedAddress, values::ParsableValue};
+use move_command_line_common::{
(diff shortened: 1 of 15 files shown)


crates/rooch-integration-test-runner/src/lib.rs
```diff
@@ -142,8 +145,28 @@ impl<'a> MoveOSTestAdapter<'a> for MoveOSTestRunner<'a> {
                     let (_, module_id) = m.module_id();
                     let mut bytes = vec![];
                     m.named_module.module.serialize(&mut bytes).unwrap();
-                    let op = Op::New(bytes);
-                    change_set.add_module_op(module_id, op).unwrap();
+                    let mut entries = BTreeMap::new();
+                    entries.insert(
```
jolestar: The StateChangeSet also needs to add some functions like `add_module_op` to operate the changeset easily.

jolestar (approved):

jolestar (approved):

baichuan3 (approved):

PR #565, opened 2023-08-02
[gh-556] clear schemes for init create and import.

### LINKED ISSUE #556: [cli] limit the abilities to select schemes for `rooch init` and `rooch create` commands.
To implement a default Ed25519 key pair and auth key to the Rooch address, it needs to remove the scheme selection for the `rooch init` and `rooch create` commands.

Part of #558

feliciss:
Resolves #556


jolestar (approved):
LGTM. @wow-sven take a review.

jolestar:
The `fastcrypto` has a breaking change, and maybe we need to lock the `fastcrypto` dependency to a commit hash.

jolestar (approved):

PR #561, opened 2023-08-02
Support the iterate query for statedb

baichuan3:
resolve https://github.com/rooch-network/rooch/issues/543

Two interfaces `listStates` and `listAnnotatedStates` are provided, which support state traversal in the following format:

```
1. /resource/$account_address
2. /module/$account_address
3. /table/$table_handle
```

Example 1
```
curl --location 'http://localhost:50051' \
--header 'Content-Type: application/json' \
--data '{
 "id":101, 
 "jsonrpc":"2.0", 
 "method":"rooch_listStates", 
 "params":["/resource/0xb4321fa441b5d9fdefb71f82856a56447451f7b1ba9478747b07e9f26b34c87",null, 5]
}'
```

```
{
    "jsonrpc": "2.0",
    "result": {
        "data": [
            {
                "value": "0x0600000000000000",
                "value_type": "0x3::account::Account"
            }
        ],
        "next_cursor": "0x52303030303030303030303030303030303030303030303030303030303030303030303030303030303030303030303030303030303030303030303030303030333a3a6163636f756e743a3a4163636f756e74",
        "has_next_page": false
    },
    "id": 101
}
```

Example 2

```
curl --location 'http://localhost:50051' \
--header 'Content-Type: application/json' \
--data '{
 "id":101, 
 "jsonrpc":"2.0", 
 "method":"rooch_listAnnotatedStates", 
 "params":["/table/0x0", "0xb4598eb64be589838077c1986bf594bbe85439f6519080352481cfeea8e1ae59", 5]
}'
```

```
{
    "jsonrpc": "2.0",
    "result": {
        "data": [
            {
                "state": {
                    "value": "0x8d60d50ae44f5fccde1501eaec8e5f6e4f30f5bd5065d8de64f1eabb16168c000000000000000000000000000000000000000000000000000000000000000000a6c0ea59f9069fb83ae45db3d4ecbcd6685688d9aef4d3246def59c836082199",
                    "value_type": "0x2::object::Object<0x2::raw_table::TableInfo>"
                },
                "move_value": {
                    "abilities": 0,
                    "type": "0x2::object::Object<0x2::raw_table::TableInfo>",
                    "value": {
                        "id": "0x8d60d50ae44f5fccde1501eaec8e5f6e4f30f5bd5065d8de64f1eabb16168c00",
                        "owner": "0x0000000000000000000000000000000000000000000000000000000000000000",
                        "value": {
                            "abilities": 8,
                            "type": "0x2::raw_table::TableInfo",
[...]

PR #570, opened 2023-08-03
refactor AccessPath to support Path with keys and without keys at the same time

baichuan3:
Refactor https://github.com/rooch-network/rooch/pull/561 for using AccessPath to support state iterate queries


jolestar (approved):
LGTM. Maybe we can and the `cursor` and `limit` to the AccessPath, and merge the `state` and `listState` API in the future?

baichuan3:
> LGTM. Maybe we can and the `cursor` and `limit` to the AccessPath, and merge the `state` and `listState` API in the future?

It is an way worth trying, which can simplify the state trait

Issue #248, opened 2023-06-08
[move] Deprecate Move global storage instructions


jolestar:
## Motivation

Rooch currently supports the global storage instructions in Move(`move_to|move_from|borrow_global|borrow_globao_mut|exists`). 

At the same time, we have also provided methods to operate global storage through AccountStorage. The two overlap in functionality, and if used together in the same function, bugs may occur due to inconsistent caching.

The global storage instructions use TransactionCache in session, and AccountStorage uses the cache in the raw table extension.

Advantages of the AccountStorage mode:

1. Borrowed resources can be returned from the function, and their lifetimes are bound to StorageContext.
2. It is easy to support application-specific state space in the future.

## Solution

1. Use the verifier to report errors for global storage instructions during compile time.
2. Use the gas meter implementation to report errors for global storage instructions during runtime.
3. Finally, the ChangeSet must be empty in the session output.

## Challenges

1. This approach may bring some new learning costs to developers and how to be compatible with existing Move libraries that use global storage instructions in the future. One idea is to have the compiler replace the global storage instructions with the corresponding function calls in AccountStorage during compilation.
2. Security evaluation: #82 

## Ref

1. Discussion: https://discord.com/channels/1078938449974935592/1116288172276584578
2. StorageAbstraction: https://github.com/rooch-network/rooch/blob/main/docs/design/storage_abstraction.md

wubuku:
I feel like it's okay.

PR #576, opened 2023-08-06
disable all built-in instruction related to global storage

stevenlaw123:
This is part of  https://github.com/rooch-network/rooch/issues/248

### DIFF
--- a/moveos/moveos-verifier/src/metadata.rs
+++ b/moveos/moveos-verifier/src/metadata.rs
@@ -104,6 +104,7 @@ impl<'a> ExtendedChecker<'a> {
                 self.check_private_generics_functions(module);
                 self.check_entry_functions(module);
                 self.check_init_module(module);
+                self.check_global_storage_access(module);
             }
         }
     }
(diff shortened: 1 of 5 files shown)


jolestar (approved):

moveos/moveos-verifier/src/metadata.rs
```diff
@@ -469,6 +470,116 @@ pub fn is_allowed_input_struct(name: String) -> bool {
...
+
+            if !invalid_bytecode.is_empty() {
+                let func_name = self
+                    .env
+                    .symbol_pool()
+                    .string(fun.get_name())
+                    .as_str()
+                    .to_string();
+
+                let error_msg = format!(
```
baichuan3, line 572: If use built-in instruction related to global storage, Is it directly replaced by the new instruction of Storage Abstract, or an error is thrown directly?

stevenlaw123: The following code inserts a compilation error in the environment, and causing the subsequent checking program to halt the compilation process.

```rust
self.env.error(&fun.get_loc(), error_msg.as_str());
```

baichuan3 (approved):

Issue #468, opened 2023-07-13
[Document] AccountAbstraction


jolestar:
- [ ] Authenticator Abstraction
- [ ] Session Key
- [ ] Transaction Fee Abstraction

Issue #419, opened 2023-07-06
[AccountAbstraction]Session Key


jolestar:
## Motivation

The session key is a specific type of authentication key with the following features:

1. Its lifespan is limited, and it becomes invalid after expiration.
2. Its usage is restricted to specific purposes, such as invoking functions within a designated module.
3. Automatic renewal can be implemented based on the last usage time.

So, the App can store the session key in the App storage(such as the browser storage), and enabling automatic transaction signing without constant interaction with the wallet enhances the user experience. 

## Solution

The session key solution is dependent on the #460 

### How to indicate that a transaction has used a session key?

1. Implement SessionKey as a special Authenticator base on Authenticator Abstraction.
   * Pros: It appears simple and isolates the logic for session keys.
   * Cons: It can become complicated if SessionKey needs to support different authenticator schemes.

2. Implement SessionKey as part of the Account. Attempt to retrieve the authentication key from the account and, if it is a session key, follow the validation process specific to session keys. Otherwise, proceed with alternative validation procedures.
   * Pros: SessionKey can reuse the authenticator scheme.
   * Cons: It strongly impacts the account and transaction validation processes.

jolestar:
Finished, the document needs to be improved. #468

PR #582, opened 2023-08-07
[AccountAbstraction] Implement SessionKey

jolestar:
1. The `transaction_validator::validate` return TxValidateResult.
2. The executor puts the `TxValidateResult` into TxContext.
3. The `transaction_validator::post_execute` update the SessionKey's active time.
4. Refactor the RustBindingTest, support execute tx before call function. cc @feliciss . Some auth validators need to update AuthenticationKey before calling the validate function and can use this approach to test. 

part of #419 

TODO
1. Add more tests.
2. CLI support SessionKey.
3. AccountAbstraction Documents #468. 
4. JS SDK support SessionKey.


crates/rooch-framework/sources/session_key.move
```diff
@@ -0,0 +1,144 @@
...
+            last_active_time: 0,
+            max_inactive_interval: max_inactive_interval,
+        };
+        if (!account_storage::global_exists<SessionKeys>(ctx, sender_addr)){
+            let keys = table::new<vector<u8>, SessionKey>(storage_context::tx_context_mut(ctx));
+            account_storage::global_move_to<SessionKeys>(ctx, sender, SessionKeys{keys});
+        };
+
+        let session_keys = account_storage::global_borrow_mut<SessionKeys>(ctx, sender_addr);
+        table::add(&mut session_keys.keys, authentication_key, session_key);
```
baichuan3, line 91: If the auth key rotated is implemented later, the corresponding sessionkey must be invalidated after being rotated, which requires a contract method

baichuan3 (approved):

PR #593, opened 2023-08-08
[gh-592] add other schems to session keys.

### LINKED ISSUE #592: [session-key] support other schemes for session keys
Right now there's only ed25519 scheme supported for session keys. Assuming ed25519 authentication payload as session keys's key map and ed25519 and other schemes' auth keys can be authentication keys in the session key's value map.

I'll add other schemes as feasible to #582.

feliciss:
Resolves #592 

This PR does:

1. Add hash variable for ECDSA, ECDSARecoverable and Schnorr schemes that need verify signature and message using predefined hash function.
2. Refactor authentication key for other schemes according to #582.
3. Support other schemes for session keys.


feliciss:
> Refactor the RustBindingTest, support execute tx before call function. cc @feliciss . Some auth validators need to update AuthenticationKey before calling the validate function and can use this approach to test.

@jolestar Can you explain in more detail? I added other schemes support for the session key and when running validator tests for other schemes ecdsa, schnorr, etc. through the `validate` function. It works well after calling `validate_signature` and get authentication keys from the authenticator payload through `get_authentication_key_from_payload` and from the sender account via `get_authentication_key`. This way, the first-time authentication key would be each scheme's public key to address so comparing auth key from payload and auth key in account, i.e. non-ed25519 address as auth key, would not cause any problem then.

However, with calling `update` command to rotate and update authentication keys other than `ed25519`, it comes with a problem:

```zsh
   ✘  Then cmd: "account update --address 0xebf29d2aed4da3d2e13a32d71266a302fbfd5ceb3ff1f465c006fa207f1789ce --scheme ecdsa --mnemonic-phrase "spike air embody solid upper grow mule slender shrimp suggest pride young""
      Step failed:
      Defined: features/cmd.feature:13:7
      Matched: crates/testsuite/tests/integration.rs:45:1
      Step panicked. Captured output: CLI should run successfully.: TransactionError("Update authentication key failed for scheme ecdsa on address 0xebf29d2aed4da3d2e13a32d71266a302fbfd5ceb3ff1f465c006fa207f1789ce. Reason: Transaction error: RPC call failed: ErrorObject { code: ServerError(-32000), message: \"validate failed: VMError with status ABORTED with sub status 66537 at location Module ModuleId { address: 0000000000000000000000000000000000000000000000000000000000000003, name: Identifier(\\\"ecdsa_k1_validator\\\") } and message 0x0000000000000000000000000000000000000000000000000000000000000003::ecdsa_k1_validator::validate at offset 19 at code offset 19 in function definition 9\", data: None }.")
```

It seems that `rotate_authentication_key_entry` can not be succeeded since something went wrong with a breaking change introduced #582?

When I execute `rotate_authentication_key_entry` from command line interface, there's an interesting error with `NUMBER_OF_ARGUMENTS_MISMATCH`. Probably it's because a new variable `hash: u8` was added in authenticator and session keys:

```zsh
Transaction error: Update authentication key failed for scheme ecdsa on address 0xdf30c42447c7271adafab210ec82f1ed6ac64e969a272927402ad62ec2ba81bb. Reason: Transaction error: RPC call failed: ErrorObject { code: ServerError(-32000), message: "validate failed: VMError with status NUMBER_OF_ARGUMENTS_MISMATCH at location UNDEFINED and message argument length mismatch: expected 4 got 5", data: None }.
```

Please indicate which way I should refactor auth keys validation referring #582. If you are available please help me to debug the above implications, thanks.

jolestar:
We need to add a test case for `rotate_authentication_key_entry` to avoid breaking like this.

feliciss:
> We need to add a test case for `rotate_authentication_key_entry` to avoid breaking like this.

I recorded an issue https://github.com/rooch-network/rooch/issues/595.

Since it's now TBD, I will go ahead to solve other issues in the Rooch roadmap.

jolestar:
We do not support other crypto schemes to session keys now.

PR #602, opened 2023-08-10
[Framework] Refactor AuthencationKey and Ed25519 AuthValidator

jolestar:
1. Refactor AuthencationKey, use a TypeTable to store all AuthencationKey of a user, and avoid calling `global_move_to` when rotating the authentication key. The Validator can rotate the authentication key via account `address`, and does not need a `signer`.  @feliciss 
2. Fix the `account move explain` bug. We should split the code into `category` and `reason` codes before getting the description from the error description file. @stevenlaw123 
3. Refactor Ed25519 AuthValidator, rename some function names, and update the AuthencationKey in the `transaction_validator::post_execute` function. @feliciss


baichuan3 (approved):

jolestar:
Find a VM bug, we keep the resolved_arg in the VerifiedMoveAction before, so the TxContext argument is using the value when verifying tx, and the ids_created value does not update if we create a table in `pre_execute`.

I fixed this bug, but there are still some problems with the TxContext if the transaction MoveAction is executed failed.

@stevenlaw123 @baichuan3

jolestar:
I record an issue #607, and merge this PR first.

Issue #544, opened 2023-07-28
[Storage] Add storage test cases to verify data reloading consistency after service restart


baichuan3:
The implementation idea provides storage dump command, simulates reloading the db scene after the service is restarted, and verifies the consistency of the data loaded from the db.

PR #603, opened 2023-08-10
Storage reload consistance testcase

baichuan3:
resolve https://github.com/rooch-network/rooch/issues/544

After dumping the full data from the state db, and then re-apply, the consistency verification of the statedb passes.


moveos/moveos-store/src/tests/test_state_store.rs
```diff
@@ -1,15 +1,103 @@
...
+use rand::{thread_rng, Rng};
 use smt::NodeStore;
 use std::str::FromStr;
 
-use crate::MoveOSStore;
+fn random_bytes() -> Vec<u8> {
+    H256::random().0.to_vec()
+}
+
+fn random_account_change_set() -> AccountChangeSet {
```
jolestar, line 22: The AccountChangeSet will deprecated after the #399

moveos/moveos-store/src/state_store/statedb.rs
```diff
@@ -297,6 +306,34 @@ impl StateDBStore {
...
+
+    // pub fn dump_iter(
+    //     &self,
+    //     handle: &ObjectID,
+    // ) -> Result<Option<SMTIterator<Vec<u8>, State, NodeDBStore>>> {
+    //     if handle == &state_resolver::GLOBAL_OBJECT_STORAGE_HANDLE {
+    //         self.global_table.iter().map(|v| Some(v))
+    //     } else {
+    //         self.get_as_table(*handle)
+    //             .and_then(|res| res.map_or(Ok(None), |(_, table)| table.iter().map(|v| Some(v))))
```
baichuan3, line 332: There is an problem with lifetime constraints, due to `SMTIerator<'a, K, V, R>` causing compilation failure:

```
   --> moveos/moveos-store/src/state_store/statedb.rs:332:67
    |
332 |                 .and_then(|res| res.map_or(Ok(None), |(_, table)| table.iter().map(|v| Some(v))))
    |                                                                   ------------^^^^^^^^^^^^^^^^^
    |                                                                   |
    |                                                                   returns a value referencing data owned by the current function
    |                                                                   `table` is borrowed here
```

or rewritten as something like
```
    pub fn dump_iter(
        &self,
        handle: &ObjectID,
    ) -> Result<Option<SMTIterator<Vec<u8>, State, dyn NodeStore + 'static>>> {
...
}
```

but it will also compile failure, error stack:
```
error[E0277]: the size for values of type `(dyn NodeStore + 'static)` cannot be known at compilation time
   --> moveos/moveos-store/src/state_store/statedb.rs:327:10
    |
327 |     ) -> Result<Option<SMTIterator<Vec<u8>, State, dyn NodeStore + 'static>>> {
    |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ doesn't have a size known at compile-time
    |
    = help: the trait `Sized` is not implemented for `(dyn NodeStore + 'static)`
```

Any suggestions for better writing? @jolestar

jolestar: I will give it a try.

PR #610, opened 2023-08-11
[gh-609] refactor validators for other schemes.

### LINKED ISSUE #609: [rooch-framework] naming and refactor auth validators for other schemes.
Renaming constant variables of error codes and functions.

Refactor `schnorr`, `ecdsa`, `ecdsa-recoverable` auth validators following #602.

Support authentication key length validation for `post_execute` and `validate` functions for non-ed25519 validators without default account addresses as authentication keys, ensuring authentication key length from authenticator payload, public key length when calling `rotate_authentication_key_entry`, and authentication key length in account are valid and not malformed in the transaction process for the `account update` and `account nullify` call.

feliciss:
Resolve #609 

Relative issue #602


crates/rooch-framework/sources/auth_validator/ecdsa_k1_validator.move
```diff
@@ -83,61 +82,95 @@ module rooch_framework::ecdsa_k1_validator {
...
+    /// Get the authentication key of the given authenticator from authenticator_payload.
+    public fun get_authentication_key_from_authenticator_payload(authenticator_payload: &vector<u8>): vector<u8> {
+        let public_key = get_public_key_from_authenticator_payload(authenticator_payload);
+        let addr = public_key_to_address(public_key);
         moveos_std::bcs::to_bytes(&addr)
     }
 
-    public fun ecdsa_k1_public_key_to_address(public_key: vector<u8>): address {
+    public fun public_key_to_address(public_key: vector<u8>): address {
+        moveos_std::bcs::to_address(public_key_to_authentication_key(public_key))
```
jolestar, line 92: ecdsa_k1 do not support public key to address. Or do we support generating the ETH address via the ecdsa_k1 public key?

crates/rooch-framework/sources/auth_validator/ecdsa_k1_validator.move
```diff
@@ -83,61 +82,95 @@ module rooch_framework::ecdsa_k1_validator {
...
+        validate_signature(&authenticator_payload, &tx_hash);
+
+        let auth_key_from_authenticator_payload = get_authentication_key_from_authenticator_payload(&authenticator_payload);
+        std::debug::print(&auth_key_from_authenticator_payload);
+        // Although we have checked public key length in rotate_authentication_key_entry function,
+        // it needs to validate the authentication key isn't empty or malformed.
+        assert!(
+           vector::length(&auth_key_from_authenticator_payload) == V_AUTHENTICATION_KEY_LENGTH,
+           error::invalid_argument(EInvalidAuthenticatorPayloadAuthKeyLength)
+        );
```
jolestar, line 146: We need to compare the auth_key from the payload with the auth_key from the account.
if the auth_key from the account is none, and the MultiChainAddress in context is not empty, we can use the ETH address as the auth_key.

crates/rooch-framework/sources/auth_validator/ecdsa_k1_validator.move
```diff
@@ -83,61 +82,95 @@ module rooch_framework::ecdsa_k1_validator {
...
         _ctx: &mut StorageContext,
     ) {}
 
     fun post_execute(
-        _ctx: &mut StorageContext,
-    ) {}
+        ctx: &mut StorageContext,
+    ) {
+        let account_addr = storage_context::sender(ctx);
+        let auth_key_in_account = get_authentication_key_from_account(ctx, account_addr);
```
jolestar: Same as above, we cant use MultiChainAddress to rotate_authentication_key_entry

crates/rooch-framework/sources/auth_validator/ecdsa_k1_recoverable_validator.move
```diff
@@ -87,60 +81,95 @@ module rooch_framework::ecdsa_k1_recoverable_validator {
...
+    /// Get the authentication key option of the given account.
+    public fun get_authentication_key_option_from_account(ctx: &StorageContext, addr: address): Option<vector<u8>> {
         let auth_key_option = account_authentication::get_authentication_key<EcdsaK1RecoverableValidator>(ctx, addr);
         if (option::is_some(&auth_key_option)) {
-            option::extract(&mut auth_key_option)
+            auth_key_option
         }else {
-            //if AuthenticationKey does not exist, return addr as authentication key
-            moveos_std::bcs::to_bytes(&addr)
+            option::none<vector<u8>>()
```
jolestar, line 108: Just return `auth_key_option` here.

crates/rooch-framework/sources/auth_validator/ecdsa_k1_validator.move
```diff
@@ -83,61 +81,96 @@ module rooch_framework::ecdsa_k1_validator {
...
+        let account_addr = storage_context::sender(ctx);
+        if (is_authentication_key_in_account(ctx, account_addr)) {
+            let auth_key_in_account = get_authentication_key_from_account(ctx, account_addr);
+            std::debug::print(&auth_key_in_account);
+            // Although we have checked public key length in rotate_authentication_key_entry function,
+            // it needs to validate the authentication key isn't empty or malformed.
+            assert!(
+                vector::length(&auth_key_in_account) == V_AUTHENTICATION_KEY_LENGTH,
+                error::invalid_argument(EInvalidAccountAuthKeyLength)
+            );
```
jolestar, line 165: Do not need to check auth length in the post_execute. We must ensure the post_execute function is always executed success.

crates/rooch-framework/sources/auth_validator/ecdsa_k1_validator.move
feliciss: Yes. Both ecdsa_k1 and ecdsa_k1_recoverable's public key length are 33 bytes. The public key can then be used to generate ETH addresses:

```move
let uncompressed = decompress_pubkey(&pubkey);

// Take the last 64 bytes of the uncompressed pubkey.
let uncompressed_64 = vector::empty<u8>();
let i = 1;
while (i < 65) {
 let value = vector::borrow(&uncompressed, i);
 vector::push_back(&mut uncompressed_64, *value);
 i = i + 1;
};

// Take the last 20 bytes of the hash of the 64-bytes uncompressed pubkey.
let hashed = hash::keccak256(&uncompressed_64);
let addr = vector::empty<u8>();
let i = 12;
while (i < 32) {
 let value = vector::borrow(&hashed, i);
 vector::push_back(&mut addr, *value);
 i = i + 1;
};

addr
```

I think the same way applies for Bitcoin address using ripemd-160 hash.

feliciss: May split `public_key_to_address` to `public_key_to_ethereum_address` and `public_key_to_bitcoin_address`. Just need to determine which address to use as the authentication key.

crates/rooch-framework/sources/auth_validator/ecdsa_k1_validator.move
feliciss: Okay. Will remove them.

crates/rooch-framework/sources/auth_validator/ecdsa_k1_validator.move
jolestar: So the address should not be `address`. We should define an EVMAddress type.

crates/rooch-framework/sources/auth_validator/ecdsa_k1_validator.move
feliciss: Did you mean context the TxContext?

```zsh
[debug] 0x2::storage_context::StorageContext {
  tx_context: 0x2::tx_context::TxContext {
    sender: @0xebf29d2aed4da3d2e13a32d71266a302fbfd5ceb3ff1f465c006fa207f1789ce,
    tx_hash: 0x6ab9102e46eccda49488d033219b1ba9992f446608115f72061a70e0313e5c81,
    ids_created: 0,
    map: 0x2::simple_map::SimpleMap<0x1::string::String, 0x2::copyable_any::Any> {
      data: []
    }
  },
  object_storage: 0x2::object_storage::ObjectStorage {
    handle: 0x2::object_id::ObjectID {
      id: @0x0
    }
  }
}
```

Unfortunately, I couldn't find for MultiChainAddress, and the sender in the context, the auth_key from the account, is default to none for non-ed25519 schemes since only ed25519 address exists at the beginning and I need to call rotate auth keys to update a one in the validator store.

crates/rooch-framework/sources/auth_validator/ecdsa_k1_validator.move
feliciss: You mean MultiChainAddress a Rooch address? I'll refactor `public_key_to_authentication_key` function to convert to a ETH or BTC address, TBD, and pass it in `rotate_authentication_key_entry ` to update an ETH or BTC address as an auth key.

crates/rooch-framework/sources/auth_validator/ecdsa_k1_validator.move
feliciss:
```rs
impl AccountAddress {
    pub const fn new(address: [u8; Self::LENGTH]) -> Self {
        Self(address)
    }

    /// The number of bytes in an address.
    /// Default to 16 bytes, can be set to 20 bytes with address20 feature.
    pub const LENGTH: usize = if cfg!(feature = "address20") {
        20
    } else if cfg!(feature = "address32") {
        32
    } else {
        16
    };
```

Both ETH and BTC are in 20-byte address. While although Rooch address is in 32-byte address, according to the above code from `MoveValue::Address`, it can be set to 20 bytes if needed? Or in anyway EVMAddress type is what we need and we should define it in `move_core_types::value::MoveValue` or somewhere else?

feliciss: A handling way to implement `address20` feature is:

```toml
move-core-types = { git = "https://github.com/rooch-network/move", rev = "55b8bf6967e76ff11933fffaccb76d5a4e15a23a", features = ["address32", "address20"] }
```

But it needs to be well tested in all imported libraries and could be a breaking change.

crates/rooch-framework/sources/auth_validator/ecdsa_k1_validator.move
jolestar: MultiChainAddress for ETH address when the chain is ETH.

crates/rooch-framework/sources/auth_validator/ecdsa_k1_validator.move
jolestar: If the user uses MetaMask and submits the Ethereum transaction, the MultiChainAddress will in the TxContext

https://github.com/rooch-network/rooch/blob/420a92379b226777689fce8245c0e6cdb7405c79/crates/rooch-framework/sources/transaction_validator.move#L96-L104

crates/rooch-framework/sources/auth_validator/ecdsa_k1_validator.move
jolestar: We need to define the EVMAddress in Move. The `address` type is only for Rooch or Move addresses.

jolestar: Like https://github.com/starcoinorg/starcoin-framework/blob/1f8ebec924fe8768888b02975614650c4a1c63c9/sources/Signature.move#L46

crates/rooch-framework/sources/auth_validator/ecdsa_k1_recoverable_validator.move
```diff
@@ -38,32 +38,24 @@ module rooch_framework::ecdsa_k1_recoverable_validator {
...
-        account_authentication::rotate_authentication_key<EcdsaK1RecoverableValidator>(
-            ctx,
-            account_addr,
-            ecdsa_k1_recoverable_authentication_key
-        );
+    fun rotate_authentication_key(ctx: &mut StorageContext, account_addr: address, authentication_key: vector<u8>) {
+        account_authentication::rotate_authentication_key<EcdsaK1RecoverableValidator>(ctx, account_addr, authentication_key);
     }
 
     public entry fun remove_authentication_key_entry<EcdsaK1RecoverableValidator>(ctx: &mut StorageContext, account: &signer) {
```
jolestar, line 54: `EcdsaK1RecoverableValidator` is in the current module, not a generic type.

jolestar: Same as the rotate_authentication_key_entry function

feliciss: Fixed by https://github.com/rooch-network/rooch/pull/610/commits/b7e01c8095f3046369a319f621719a86aeecfe69 and #614.

jolestar (approved):
We can merge this first. Then We need to discuss the built-in scheme and auth validators mentioned in #613

feliciss:
```shell
status EXECUTED
Error: error[E03002]: unbound module
```

Need to rerun the CI. @jolestar

crates/rooch-framework/sources/auth_validator/ecdsa_k1_validator.move
feliciss: > If the user uses MetaMask and submits the Ethereum transaction, the MultiChainAddress will in the TxContext
> 
> https://github.com/rooch-network/rooch/blob/420a92379b226777689fce8245c0e6cdb7405c79/crates/rooch-framework/sources/transaction_validator.move#L96-L104

I've sent a transaction from MetaMask and it seems it has successfully included a map data:

```zsh
[debug] 0x2::storage_context::StorageContext {
  tx_context: 0x2::tx_context::TxContext {
    sender: @0x5920595edf45d7e863493a3f9182b4101dd2bc27146fce813081c087d3e1d3b9,
    tx_hash: 0x2928a5913dfe71aa9a809238e9d7235b608c02d29ce625abc9f970836c94f0fd,
    ids_created: 0,
    map: 0x2::simple_map::SimpleMap<0x1::string::String, 0x2::copyable_any::Any> {
      data: [
        0x2::simple_map::Element<0x1::string::String, 0x2::copyable_any::Any> {
          key: "0000000000000000000000000000000000000000000000000000000000000002::tx_meta::TxMeta",
          value: 0x2::copyable_any::Any {
            type_name: "0000000000000000000000000000000000000000000000000000000000000002::tx_meta::TxMeta",
            data: 0x010195b15595f939a8cf5ece25d44fc832acaf218a6028472b9c0785338fae21bef907636f756e74657208696e637265617365
          }
        }
      ]
    }
  },
  object_storage: 0x2::object_storage::ObjectStorage {
    handle: 0x2::object_id::ObjectID {
      id: @0x0
    }
  }
}
```

Anyway can I verify the `TxMeta`'s data `0x010195b15595f939a8cf5ece25d44fc832acaf218a6028472b9c0785338fae21bef907636f756e74657208696e637265617365` is indeed the MetaMask sender's address?

feliciss: @jolestar I used MetaMask to connect to a localhost Rooch RPC, but it seems that there's no multi-chain address mapping in the `tx_context`?

PR #611, opened 2023-08-11
add doc comment and fix typo to the server command

geometryolife:


crates/rooch/src/commands/server/mod.rs
```diff
@@ -26,7 +26,7 @@ impl CommandAction<String> for Server {
 }
 
 #[derive(clap::Subcommand)]
-#[clap(name = "servre")]
+#[clap(name = "server")]
```
jolestar, line 29: How long this typo exists? I use the `rooch server start` is worked.

jolestar (approved):

PR #634, opened 2023-08-15
add `rooch server clean` to rooch cli

### LINKED ISSUE #584: Rooch CLI server add clean subcommand
Clean storage

bytedeveloperr:
Resolves #584

### DIFF
--- a/crates/rooch/src/commands/server/commands/clean.rs
+++ b/crates/rooch/src/commands/server/commands/clean.rs
@@ -1,0 +1,39 @@
+// Copyright (c) RoochNetwork
+// SPDX-License-Identifier: Apache-2.0
+
+use clap::Parser;
+use rooch_config::store_config::StoreConfig;
+use rooch_types::error::{RoochError, RoochResult};
+use std::{fs, path::Path};
+
+/// Clean the Rooch server storage
+#[derive(Debug, Parser)]
+pub struct CleanCommand;
+
+impl CleanCommand {
+    pub async fn execute(self) -> RoochResult<()> {
+        let rooch_store = StoreConfig::get_rooch_store_dir();
+        let moveos_store = StoreConfig::get_moveos_store_dir();
+
+        self.remove_store_dir(&rooch_store, "Rooch")?;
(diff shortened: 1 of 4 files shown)


crates/rooch/src/commands/server/commands/clean.rs
```diff
@@ -0,0 +1,39 @@
...
+
+impl CleanCommand {
+    pub async fn execute(self) -> RoochResult<()> {
+        let rooch_store = StoreConfig::get_rooch_store_dir();
+        let moveos_store = StoreConfig::get_moveos_store_dir();
+
+        self.remove_store_dir(&rooch_store, "Rooch")?;
+        self.remove_store_dir(&moveos_store, "MoveOS")?;
+
+        println!("Rooch server storage successfully cleaned");
```
jolestar, line 21: Do we need to remove the Keystore file in the clean command? @wow-sven

bytedeveloperr: that doesn't remove the keystore @jolestar,  It's the dbpath

bytedeveloperr: the keystore is stored in `~/.rooch/rooch_config` while the code removes files under `~/.rooch/roochdb` @jolestar

jolestar: Yes. I want to ask about the goal of clean command. Sometimes, the Keystore serialize format changes also break the CLI, so should we support cleaning the Keystore?

bytedeveloperr: the goal of the clean command as far as I can tell is to remove the server db storage i.e rooch db store and moveos db store, in some cases that command might be needed.

regarding the keystore, cleaning the keystore doesn't make make sense and very risky because the users private keys are stored there. so, cleaning it means the user will loose their keys forever

jolestar: Ok, we need to care about the keystore file format.

jolestar (approved):

PR #640, opened 2023-08-15
[SessionKey] Check the transaction with session scope

### LINKED ISSUE #631: [Framework] Implement the SessionKey scope check
Check the SessionKey scope. 

If the transaction MoveAction's function is beyond the scope, need to abort the transaction.

### Dependency
1. #487

jolestar:
1. Define TxMeta, and put TxMeta in TxContext. It is too hard to define MoveAction in Move #495 , so I use a simple way to define a TxMeta.
2. Check the TxMeta with the session scope.

resolve #631


baichuan3 (approved):
LGTM

Issue #644, opened 2023-08-16
[VM] handle the transaction validate error 


jolestar:
Return a custom Error type or VMStatus, and let the caller can handle the transaction validate error via error code.

https://github.com/rooch-network/rooch/blob/d4dba6c8d8d1197610b91f3d367a088cac63dcdb/crates/rooch-executor/src/actor/executor.rs#L102-L109

PR #646, opened 2023-08-16
[VM] Refactor view_function result, return the VMStatus to caller

jolestar:
Prepare for #644 

1. Define `FunctionResult`, and return the `VMStatus` and `Vec<FunctionReturnValue>` to the caller.
2. The RPC `execute_view_function` response changed from `Vec<AnnotatedFunctionReturnValue>` to `AnnotatedFunctionResultView`. This change will break the RPC, and some documents need to be updated. @geometryolife 

```bash
rooch move view --function 0x2763f0616014847fc57b78fe267d0999f4277770ddf521507005059839ac0946::counter::value
```
```json
{
  "vm_status": "Executed",
  "return_values": [
    {
      "value": {
        "type_tag": "u64",
        "value": "0x0000000000000000"
      },
      "move_value": "0"
    }
  ]
}
```

Now, the caller can handle the function abort code.

### DIFF
--- a/crates/rooch-executor/src/actor/executor.rs
+++ b/crates/rooch-executor/src/actor/executor.rs
@@ -22,6 +22,7 @@ use moveos_store::transaction_store::TransactionStore;
 use moveos_store::MoveOSStore;
 use moveos_types::event::AnnotatedMoveOSEvent;
 use moveos_types::event::EventHandle;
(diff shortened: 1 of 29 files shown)


templexxx (approved):

jolestar:
This change will break the sdk @wow-sven @yubing744

PR #655, opened 2023-08-17
[MoveOS & Executor] Refactor validator to let caller got VMStatus

### LINKED ISSUE #644: [VM] handle the transaction validate error

jolestar:
follow #646 

resolve #644 

1. Return VMStatus from VM execute_view_function to validate function. We need to define some embed Result struct, eg: `Result<Result<T>, VMStatus>`.
2. Improve the session key test, to check the move abort code.

cc @feliciss

### DIFF
--- a/crates/rooch-executor/src/actor/executor.rs
+++ b/crates/rooch-executor/src/actor/executor.rs
@@ -11,13 +11,14 @@ use crate::actor::messages::{
     ListAnnotatedStatesMessage, ListStatesMessage,
 };
 use accumulator::inmemory::InMemoryAccumulator;
-use anyhow::bail;
 use anyhow::Result;
 use async_trait::async_trait;
 use coerce::actor::{context::ActorContext, message::Handler, Actor};
 use move_core_types::account_address::AccountAddress;
+use move_core_types::vm_status::VMStatus;
 use move_resource_viewer::MoveValueAnnotator;
 use moveos::moveos::MoveOS;
+use moveos::vm::vm_status_explainer::explain_vm_status;
 use moveos_store::transaction_store::TransactionStore;
 use moveos_store::MoveOSStore;
 use moveos_types::event::AnnotatedMoveOSEvent;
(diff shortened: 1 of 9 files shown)


crates/rooch-executor/src/actor/executor.rs
```diff
@@ -115,41 +121,58 @@ impl ExecutorActor {
...
+                            .validate(ctx, authenticator.authenticator.payload)?
+                            .into_result();
+                        match auth_validator_function_result {
+                            Ok(_) => {
+                                // pre_execute_function: TransactionValidator first, then AuthValidator
+                                let pre_execute_functions = vec![
+                                    TransactionValidator::pre_execute_function_call(),
+                                    auth_validator_caller.pre_execute_function_call(),
+                                ];
+                                // post_execute_function: AuthValidator first, then TransactionValidator
```
baichuan3, line 146: Why pre_execute_function: TransactionValidator first, then AuthValidator , and post_execute_function the order is reversed?

crates/rooch-framework-tests/src/tests/transaction_validator_tests.rs
```diff
@@ -37,6 +39,8 @@ fn test_validate_ed25519() {
 
     transaction_validator
         .validate(&move_tx.ctx, auth_info)
+        .unwrap()
```
baichuan3, line 42: Is it used `?` more suitable then `unwrap` ?

crates/rooch-executor/src/actor/executor.rs
jolestar: Yes, like AOP, TransactionValidator is on the outside.

baichuan3 (approved):

PR #659, opened 2023-08-17
[gh-652] implement ETHAddress Move module and return type, tests for the implementation.

### LINKED ISSUE #652: [ETHAddress] define ETH address type from public key to Ethereum addresses
Part of #615

feliciss:
Resolves #652


crates/rooch-types/src/crypto.rs
```diff
@@ -975,11 +976,14 @@ mod tests {
...
-            address.to_string(),
-            "0x8c891976da9498ec1d3ff778a5d6c40c217d63cc8c48539c959f8b683eedf5a4"
-        );
+        let public_key = keypair.public();
+        let uncompressed = public_key.pubkey.serialize_uncompressed();
+        let uncompressed_64 = uncompressed[1..65].to_vec();
+        let hashed = keccak256(uncompressed_64);
+        let address_bytes = hashed[12..32].to_vec();
+        let address_str = format!("0x{}", hex::encode(address_bytes)); // Include "0x" prefix
+        let expected_address = "0x1a642f0e3c3af545e7acbd38b07251b3990914f1";
```
feliciss, line 985: This address has collision with an address already exists on Mainnet Ethereum.

crates/rooch-framework/sources/address_type/ethereum_address.move
```diff
@@ -0,0 +1,74 @@
+module rooch_framework::ethereum_address {
+    use std::vector;
+    use std::error;
+    use rooch_framework::ecdsa_k1_recoverable;
+    use rooch_framework::hash;
+
+    // Ethereum addresses are always 20 bytes
+    const ETHEREUM_ADDR_LENGTH: u64 = 20;
```
jolestar, line 8: The const started with `E` will be treated as an error code. We need to add a prefix to it.

jolestar (approved):

crates/rooch-framework/sources/address_type/ethereum_address.move
feliciss: Okay.

feliciss: I've fixed it.

feliciss: @jolestar 

Also is this needed to be fixed, since it's starting with `E`?

```move
```

feliciss: We discuss it in #663.

feliciss: > The const started with `E` will be treated as an error code. We need to add a prefix to it.

Reference:

Constant names: should be upper camel case and begin with an E if they represent error codes (e.g., EIndexOutOfBounds) and upper snake case if they represent a non-error value (e.g., MIN_STAKE).

crates/rooch-types/src/crypto.rs
feliciss: This code only implements key to address conversion. It means it's accepting a public key only. If someone wants to recover a public key from signature and message hash, they can do so and pass the recovered public key to the function.

The address here can be tested at https://toolkit.abdk.consulting/ethereum#key-to-address. Just input the public key `0x03a4f75d35449bb6c72757f3e2e765eecb6b15407d9f55678616b7ecf7e7e48211 ` and you get the same address `0x6A84EAE48a19f1ec014C627d4A1ffc50ce42c516 `.

The code works.

PR #664, opened 2023-08-18
[gh-663] unify scheme, variables and constant errors, add script and GitHub Action for checking error prefixes. 

### LINKED ISSUE #663: [style] code style for dealing with `E` variable in Move.
Several implementations showed mishandling of constants. Such as:

```move
```

And to tackle with it, some implementations suggested using a prefix `V`, for example:

```move
/// constant codes
const V_ECDSA_K1_RECOVERABLE_TO_ETHEREUM_SCHEME_LENGTH: u64 = 1;
```

I suggest to unify the coding style to avoid the `E` starting word. It should be started with `VALID` instead of `V` which indicates a variable:

```move
const VALID_ETHEREUM_ADDR_LENGTH: u64 = 20;
```

Because the const in move with every capital words already indicates it's a variable.

Update:

I think the upper snake case is okay for beginning with `E`? @jolestar 

Reference:

- **Constant names**: should be upper camel case and begin with an `E` if they represent error codes (e.g., `EIndexOutOfBounds`) and upper snake case if they represent a non-error value (e.g., `MIN_STAKE`).

feliciss:
This PR closes #663 

1. Unify scheme to begin with `SCHEME_`.
2. Rename `V_` to `VALID_` for a better presentation.
3. Rename error code prefix to `E` snake case for move-stdlib, `Error` camel case for moveos-stdlib, rooch-framework, and examples.
4. Add script and GitHub Action for checking the error code prefix compliance in Move.


crates/rooch-framework/sources/crypto/ecdsa_k1.move
```diff
@@ -2,32 +2,32 @@ module rooch_framework::ecdsa_k1 {
     use std::vector;
 
     /// constant codes
-    const V_ECDSA_K1_TO_BITCOIN_SCHEME_LENGTH: u64 = 1;
-    const V_ECDSA_K1_COMPRESSED_PUBKEY_LENGTH: u64 = 33;
-    const V_ECDSA_K1_SIG_LENGTH: u64 = 64;
+    const VALID_ECDSA_K1_TO_SCHEME_BITCOIN_LENGTH: u64 = 1;
```
jolestar, line 5: Do we still need the `VALID_` prefix?

scripts/check_move_constant_errors.sh
jolestar, line 1: How about putting this script in the `pr.sh`? For the developer to quickly check the code before creating a PR.

jolestar:
@feliciss, please help resolve the conflicts, and we can merge this PR first. I am refactoring the genesis and stdlib builder and can do the error code description refactor together.

jolestar (approved):

crates/rooch-framework/sources/crypto/ecdsa_k1.move
feliciss: > Do we still need the `VALID_` prefix?

Yes until we define the error map you mentioned above and then we can remove `VALID_` prefix.

scripts/check_move_constant_errors.sh
feliciss: The purpose of running `pr.sh` is to execute Rust binaries and dependencies. As such, I think we don't need necessarily to put the script into `pr.sh`.

feliciss: If developers want to check, they can execute the script directly, same as to `update_license.sh`.

PR #665, opened 2023-08-18
[gh-650] add tests for BTCAddress Move module and support passing authentication key type for Bitcoin addresses under ecdsa scheme.

### LINKED ISSUE #650: [BTCAddress] define BTC address type from public key to Bitcoin addresses
Part of #615

feliciss:
Resolve #650 again.

1. Add tests for `bitcoin_address.move`, `encoding.move`, `decoding.move`, `bitcoin_validator.move` and `crypto.rs` files.
2. Support passing authentication key types from Base58 `p2pkh` and `p2sh` into command line when updating Rooch account address with a new Bitcoin authentication key for ECDSA scheme.

Example:

1. p2pkh authentication key hex: `0x313647396e6d7a42364e5532536e53324b774c4e7671714539475631735266766a52`. p2pkh address decoded from the hex: `16G9nmzB6NU2SnS2KwLNvqqE9GV1sRfvjR`. 
2. p2sh authentication key hex: `0x33344d537568586f614e57424a614676677278643963385466337337756e634d4742`. p2sh address decoded from the hex: `34MSuhXoaNWBJaFvgrxd9c8Tf3s7uncMGB`. 

It can be viewed on a blockchain explorer, e.g. https://www.blockchain.com/explorer/addresses/btc/34MSuhXoaNWBJaFvgrxd9c8Tf3s7uncMGB.

Todo Bech32:

1. Add authentication key types for `p2tr` for Schnorr serialized X only public key length 32-bytes.
4. Add authentication key types for `p2wsh` witness script hashed public key length 32-bytes.
5. Add authentication key types for `p2wpkh` h160 (ripemd160(sha256)) hashed public key length 20-bytes.


feliciss:
@jolestar This PR is ready for review.

crates/rooch-types/src/framework/bitcoin_validator.rs
```diff
@@ -54,11 +54,17 @@ impl<'a> BitcoinValidatorModule<'a> {
...
-    pub fn rotate_authentication_key_action<V: MoveStructType>(public_key: Vec<u8>) -> MoveAction {
+    pub fn rotate_authentication_key_action<V: MoveStructType>(
+        public_key: Vec<u8>,
+        decimal_prefix_or_version: u8,
+    ) -> MoveAction {
         Self::create_move_action(
             Self::ROTATE_AUTHENTICATION_KEY_ENTRY_FUNCTION_NAME,
             vec![V::type_tag()],
-            vec![MoveValue::vector_u8(public_key)],
+            vec![
```
jolestar, line 64: `rotate_authentication_key` do not need the TypeTag argument.

crates/rooch-framework/sources/auth_validator/bitcoin_validator.move
```diff
@@ -23,19 +23,24 @@ module rooch_framework::bitcoin_validator {
         BITCOIN_SCHEME
     }
 
+    /// `rotate_authentication_key_entry` only supports rotating authentication key to a Bitcoin legacy address
+    /// becuase ecdsa k1 scheme only supports key generation of 33-bytes compressed public key at this time.
```
jolestar, line 27: the generic type `<T>` should be removed.

crates/rooch-types/src/authentication_key.rs
```diff
@@ -14,6 +15,26 @@ use std::{fmt::Display, str::FromStr};
 #[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize)]
 pub struct AuthenticationKey(#[serde_as(as = "Readable<Hex,_>")] Vec<u8>);
 
+/// Define authentication key type for each blockchain
+#[derive(Debug, Clone, Copy, PartialEq, Eq, ArgEnum)]
+pub enum AuthenticationKeyType {
+    P2PKH,
+    P2SH,
+}
```
jolestar, line 23: AuthenticationKey does not need a `type`. It is just bytes. `P2PKH` and `P2SH` should be the Bitcoin address type?

feliciss: Yes. Since we have converted payload bytes to an address as an authentication key, I do think there should be an authentication key type representing the address type, e.g. Bitcoin address.

feliciss: I originally wrote AddressType but when it is being used in `Update.rs`, then we have address and address-type flags. But the address-type flag only means not the Rooch address type but the address contained in the authentication key, so I changed address type to authentication key type as the name.

feliciss: I think bitcoin-address-type is more appropriate. So I will refactor it for now.

crates/rooch-types/src/framework/bitcoin_validator.rs
feliciss: Yes. It doesn't need such type tag.

crates/rooch-types/src/authentication_key.rs
jolestar: Maybe we should not support AuthenticationKey and rotate_authentication_key for the Eth validator and Bitcoin validator. We just verified the signature with the sender's multi-chain address. 

```
assert!(sender_address == address in the signature).
```

Because Eth and Bitcoin do not support rotate_authentication_key, If we support it, it will confuse users.

feliciss: > Maybe we should not support AuthenticationKey and rotate_authentication_key for the Eth validator and Bitcoin validator. We just verified the signature with the sender's multi-chain address.
> 
> ```
> assert!(sender_address == address in the signature).
> ```
> 
> Because Eth and Bitcoin do not support rotate_authentication_key, If we support it, it will confuse users.

I understand that Rooch address is using blake2b256 to hash bytes as authentication keys, but in the Ethereum and Bitcoin contexts, it is directly used address as authentication keys. Maybe we can use the pre-context of the address, such as base58 bytes as the authentication keys and users can opt in to encode the base58 bytes to an bitcoin address. I think that way to go is okay.

Alternatively, we can also remove the authentication key and rotate authentication key for ETH and BTC as you described it's not needed to have an address as the authentication key.

So, in two ways, we can refactor it:

1. Support pre-context bytes as the authentication key, like Rooch is using blake2b256 hash as the authentication key. Similarly, Bitcoin can use pre-context bytes as authentication key that can be encoded by base58 to form a Bitcoin address.
2. Remove authentication keys for ETH and Bitcoin.

feliciss: > Maybe we should not support AuthenticationKey and rotate_authentication_key for the Eth validator and Bitcoin validator. We just verified the signature with the sender's multi-chain address.
> 
> ```
> assert!(sender_address == address in the signature).
> ```
> 
> Because Eth and Bitcoin do not support rotate_authentication_key, If we support it, it will confuse users.

The prerequisite is that we can recover the signature to an address. This is applied for Ethereum but not Bitcoin. Because Ethereum is using 65-bytes signature and the last bytes can be used to recover to an address. Different blockchains have different algorithms and implementations. We should unify them as one.

jolestar: It is hard to unify. We follow the origin chain's algorithms and implementations.

feliciss: > It is hard to unify. We follow the origin chain's algorithms and implementations.

So what's your conclusion? Remove Ethereum's rotate authentication keys function and only recover the signature to an ETH address and compare it with the sender, while keeping Bitcoin's rotate authentication keys unchanged?

crates/rooch-framework/sources/auth_validator/bitcoin_validator.move
feliciss: Done.

jolestar (approved):

crates/rooch-types/src/authentication_key.rs
jolestar: I merge this PR first, we can discuss this later. The Eth validator's scenarios is for support Eth transaction and Eth wallet(eg. MetaMask), but the Bitcoin validator's scenarios not clear, maybe we can make it TBD.

PR #668, opened 2023-08-18
ci: use script to check lint

emptinesssubodhi:
## Summary

1. use script to check lint ( I found it wasn't being used anywhere before）
2. fix files' license which copied from aptos
3. use target-cpu instead of target-feature:
> target-feature is annoying
> skylake was introduced in 2015 and its safe enough considering the fact that x86 is backward compatible


.github/workflows/check_build_test.yml
```diff
@@ -32,7 +32,9 @@ jobs:
       - name: Check code format
         run: cargo fmt -- --check
       - name: Lint rust sources
-        run: cargo clippy --all-targets --all-features --tests --benches -- -D warnings
+        run: |
+          cargo install cargo-sort
+          ./scripts/rust_lint.sh --check
```
jolestar, line 37: Or use `./scrpts/pr.sh -c`

jolestar (approved):

PR #674, opened 2023-08-21
Define ModuleUpgradeFlag

pause125:
## Summary

1. Define `account_storage::ModuleUpgradeFlag` struct to indicate module upgrading when upgrading modules in Move.
2. Add `contains` operation for integration `assert` condition


jolestar (approved):
LGTM

baichuan3 (approved):

PR #677, opened 2023-08-21
[gh-676] resolve bug MetaMask sending Ethereum transactions to Rooch RPC endpoint.

### LINKED ISSUE #676: [bug] breaking changes calling from MetaMask Ethereum RPC introduced by PR #532
Calling from MetaMask to local Rooch RPC endpoint `http://0.0.0.0:50051` results:

```zsh
Recovery id: 0
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: TransactionError("Invalid recovery ID.")', crates/rooch-types/src/transaction/ethereum.rs:166:22
```

It should be included `0` as the recovery id. `0` indicates that the MetaMask network is a custom network.

MetaMask network:

> Network name:
> Rooch Network
> 
> New RPC URL:
> http://0.0.0.0:50051
> 
> Chain ID:
> 10001
> 
> Currency symbol:
> BCHT

feliciss:
## Summary

1. Add `normalize_recovery_id` to normalize recovery id when calling from a custom network on MetaMask.
2. Append scheme to RSV signature and public key to meet the Rooch signature's requirement.

This approach has below benefits:

- There's no need to recover a signature to a public key and to an Ethereum address, `convert_eth_transaction_signature_to_rooch_signature` function helps recovering a public key from an Ethereum signature, then one can use the public key from `ecdsa_k1_recoverable.move` and retrieve an Ethereum address in `ethereum_validator.move`. `ethereum_address.move` also helps encoding a public key from a sender, e.g. from MetaMask endpoint, to the sender's address and compare and verify the identity integrity.

- Closes #676


crates/rooch-types/src/transaction/ethereum.rs
```diff
@@ -161,8 +125,7 @@ impl AbstractTransaction for EthereumTransaction {
             //TODO should change the seqence_number to u256?
             seqence_number: self.0.nonce.as_u64(),
             authenticator: Authenticator::ecdsa_recoverable(
-                // TODO need to support handling ethereum signature to Rooch signature
-                self.convert_eth_signature_to_recoverable_secp256k1_signature()
+                self.convert_eth_transaction_signature_to_rooch_signature()
```
jolestar, line 128: We should not convert the signature in Rust. We need to put the original ethereum signature and decode it in the Move contract.

feliciss: > We should not convert the signature in Rust. We need to put the original ethereum signature and decode it in the Move contract.

The Ethereum transaction contains original signature R, S, V, and I did put them into constructing Rooch signature so they're the same signature bytes in Rust and can be decoded in Move.

feliciss: Or I should rename it to `self.into_signature` for better understanding?

feliciss: Originally,

`Ethereum Transaction` -> `Ethereum Signature`.

Now,

`Ethereum Transaction` -> `Rooch Ethereum Signature`.

jolestar: As the dev meeting discussion, the rooch native validator & signature is ed25519. `Rooch Ethereum Signature` is meanless.

jolestar (approved):
Merge this first, and refactor later base on #678

Issue #633, opened 2023-08-15
[VM] Implementing GasMeter and CostTable.


stevenlaw123:
During VM runtime, support for gas statistics and recording of built-in instructions and functionalities into CostTable.

jolestar:
#604

PR #681, opened 2023-08-22
[types] Add chain_id and max_gas_amount to Transaction

jolestar:
## Summary

1. Add  chain_id and max_gas_amount to Transaction.
2. Add sequence_number and max_gas_amount to TxContext.
3. Add chain_id as transaction_validator::validate argument.
4. Add TxResult to TxContext after tx is executed for the transaction validator to get gas_used in the post_execute function.

resolve #680

PR #684, opened 2023-08-23
The CostTable that has implemented tiered pricing.

stevenlaw123:
this is part of https://github.com/rooch-network/rooch/issues/633.

TODO:

1. Initialize the GasMeter and specify the CostTable instance when constructing the VM instance.
2. Deduct the GasFee from the user's account upon completion of VM execution.


jolestar:
Part of 2 implement in #681

baichuan3 (approved):

moveos/moveos/src/moveos.rs
```diff
@@ -173,7 +173,7 @@ impl MoveOS {
                 action
             );
         }
-        let gas_meter = MoveOSGasMeter::new(ctx.max_gas_amount);
+        let gas_meter = MoveOSGasMeter::new_unmetered();
```
jolestar, line 176: We need to use the max_gas_amount as the init budget.

PR #718, opened 2023-08-29
[Refactor] Remove signer argument from init func

### LINKED ISSUE #699: [Refactor] Remove signer argument from `fun init(ctx, signer)`
Some module init functions need `signer` to call `account_storage::global_move_to`.

The `signer` is the module address account. This will ensured by the VM. The developer may be suspicious about the correctness of the account, so will make an extra check on the address.

We can replace the `signer` with `moveos_std::signer::module_signer<T>`, and remove the signer argument.

https://github.com/rooch-network/rooch/blob/23ebddb7155979da5cc05b4b74a96253609d01fa/crates/rooch-framework/sources/address_mapping.move#L28-L36

WGB5445:
## Summary

Summary about this PR

- Closes #699


crates/rooch-framework/sources/address_mapping.move
```diff
@@ -25,7 +26,8 @@ module rooch_framework::address_mapping{
         mapping: Table<MultiChainAddress, address>,
     }
 
-    fun init(ctx: &mut StorageContext, sender: &signer) {
+    fun init(ctx: &mut StorageContext) {
+        let sender = &moveos_signer::module_signer<AddressMapping>();
```
jolestar, line 30: Remove the `rooch_framework::core_addresses::assert_rooch_framework(sender)`

jolestar (approved):

PR #719, opened 2023-08-29
Refactor genesis and stdlib builder

jolestar:
## Summary

1. Build genesis with chain id. fix #716 
2. Include the stdlib in binary, and remove the genesis file from binary.
3. Optimize document generate, fix #363

PR #723, opened 2023-08-29
[gh-663] remove `VALID_` prefix for crypto libraries.

### LINKED ISSUE #663: [style] code style for dealing with `E` variable in Move.

feliciss:
## Summary

Following PR for #664 to remove `VALID_` prefix after commit #719 

- Closes #663 again.


jolestar:
This commit added to #727, and close this.

PR #727, opened 2023-08-30
Refactor genesis builder

### LINKED ISSUE #714: [Dev] Add build.rs file to generate the genesis and RPC spec file
Add build.rs file to generate the genesis and RPC spec file.

jolestar:
## Summary

1. use build.rs to generate stdlib and error code map files.
2. use build.rs to generate RPC spec file.
3. remove stdlib and error code map file from the source code repo.
4. pickup the commit from #723 

- Closes #714


crates/rooch-framework/sources/crypto/ecdsa_k1.move
```diff
@@ -2,9 +2,9 @@ module rooch_framework::ecdsa_k1 {
     use std::vector;
 
     /// constant codes
-    const VALID_ECDSA_K1_TO_SCHEME_BITCOIN_LENGTH: u64 = 1;
-    const VALID_ECDSA_K1_COMPRESSED_PUBKEY_LENGTH: u64 = 33;
-    const VALID_ECDSA_K1_SIG_LENGTH: u64 = 64;
+    const ECDSA_K1_TO_SCHEME_BITCOIN_LENGTH: u64 = 1;
```
baichuan3, line 5: So, the convention for final constants is to choose without any prefix?

baichuan3:
Does genesis generated file also removed from the source code repo? I don’t see the genesis generated file in the commit.

crates/rooch-framework/sources/crypto/ecdsa_k1.move
jolestar: error code with `Error` prefix

jolestar:
> Does genesis generated file also removed from the source code repo? I don’t see the genesis generated file in the commit.

The genesis file has already been removed in the #719

crates/rooch-framework/sources/crypto/ecdsa_k1.move
baichuan3: I mean CONST prefix

jolestar: Do not need the `CONST` prefix for other `const`.

baichuan3 (approved):

PR #733, opened 2023-08-30
Feature/dashboard sdk

wow-sven:
## Summary

Make sure it works first.

Use tsup instead of vite packaging to solve the problem of not running in node environment.

However, it is worth noting that there may still be some remaining issues. For instance, warnings may be encountered when using certain frameworks, particularly those that do employ Server-Side Rendering (SSR). 

Summary about this PR

- Closes #issue


wow-sven:
I tried vite, esbuild, tsup

Always encounter node environment or browser environment conflict

@yubing744 Do you have any suggestions?

jolestar (approved):

PR #745, opened 2023-09-02
fix repeated init function request

### LINKED ISSUE #742: [MoveOS] Repeat call module init function when publish module
https://github.com/rooch-network/rooch/blob/3c187d689a44d23b647f6c5fd7a2eee6a1fa88a2/moveos/moveos-stdlib/moveos-stdlib/sources/account_storage.move#L186-L201

Call the `request_init_functions` in a loop, if the modules len > 1, will cause the repeat init functions.

pause125:
## Summary

Fix repeated calling module init function when publish module 

- Closes #742


jolestar (approved):
update the coins examples to check this bug is being fixed

baichuan3 (approved):

PR #760, opened 2023-09-06
Implements Account balance list cli

baichuan3:
resolve https://github.com/rooch-network/rooch/issues/744

**Problems**
1, It is hard to support converting from AnnotatedMoveValueView to AnnotatedMoveValue, mainly because of  it hard to support convert SpecificStruct to AnnotatedMoveStruct. As a result, parsing AnnotatedCoinStoreView from the RPC API is very tedious. relative of https://github.com/rooch-network/rooch/issues/758
2, The current implementation of AnnotatedMoveValueView may be flawed. Coin balance defines the U256 type, but the type obtained from AnnotatedMoveValueView is U64(StrView(10000), and the expected type is U256(StrView(10000). May return U128(StrView(value) type.

```
value: {Identifier("value"): U64(StrView(10000))} })
```

**TODO**
1, Coin balance needs to depend on CoinInfo to display the decimal point after processing decimals
2，Add test case for the balance command

**Example**

- Publish coin examples
```
rooch move publish -p examples/coins --sender-account 0xe1176537c0175d336353dad12f7eb60c658ce526eeb3cd08409e6fd8c2dfa1d7  --named-addresses coins=0xe1176537c0175d336353dad12f7eb60c658ce526eeb3cd08409e6fd8c2dfa1d7
```

- Call faucet to get test coin
```
rooch move run --function 0xe1176537c0175d336353dad12f7eb60c658ce526eeb3cd08409e6fd8c2dfa1d7::fixed_supply_coin::faucet  --sender-account 0xe1176537c0175d336353dad12f7eb60c658ce526eeb3cd08409e6fd8c2dfa1d7
```

- Call Account balance cli
```
rooch account balance -a 0xe1176537c0175d336353dad12f7eb60c658ce526eeb3cd08409e6fd8c2dfa1d7
```
Result
```
                                              Coin Type                                                |             Balance
------------------------------------------------
0x3::coin::Coin<0xe1176537c0175d336353dad12f7eb60c658ce526eeb3cd08409e6fd8c2dfa1d7::fixed_supply_coin::FSC> |              10000
0x3::coin::Coin<0x3::gas_coin::GasCoin> |            9996919761
```

- Call Account balance cli with coin_type
```
[...]

PR #774, opened 2023-09-10
Refactor account balance for support balance API and add some test case

baichuan3:
resolve TODOs in https://github.com/rooch-network/rooch/pull/760

And refactor account balance for support Rooch balance API.


baichuan3:
Encounter a deadlock in RPC Server.

Try to solve RPC server hang caused by futures::executor::block_on when impl MoveFunctionCall for the RPC Server.

crates/rooch-rpc-api/src/jsonrpc_types/rooch_types.rs
```diff
@@ -132,19 +135,19 @@ impl CoinView {
     }
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
 pub struct AnnotatedCoinView {
-    type_: StructTagView,
+    struct_type: StructTagView,
```
jolestar, line 140: We need to unify the `StructTag` field name.

Sometimes, we use `struct_tag`

https://github.com/rooch-network/rooch/blob/a63525896448ae9da62600c0f1fe86926f122883/moveos/moveos-types/src/object.rs#L377-L381 

Sometimes, we use `type_`

https://github.com/rooch-network/rooch/blob/a63525896448ae9da62600c0f1fe86926f122883/crates/rooch-rpc-api/src/jsonrpc_types/move_types.rs#L49-L55

But in the move-language/move repo, AnnotatedMoveStruct use `type_`.

```
pub struct AnnotatedMoveStruct {
    pub abilities: AbilitySet,
    pub type_: StructTag,
    pub value: Vec<(Identifier, AnnotatedMoveValue)>,
}
```

What do you think about this? @wow-sven @stevenlaw123

crates/rooch-rpc-server/src/service/aggregate_service.rs
```diff
@@ -0,0 +1,302 @@
...
+        &self,
+        _ctx: &TxContext,
+        function_call: FunctionCall,
+    ) -> Result<FunctionResult> {
+        // let function_result = futures::executor::block_on(async move {
+        // let function_result = futures::executor::block_on(async move {
+        //     self.rpc_service.execute_view_function(function_call).await
+        // })?;
+
+        let function_result =
```
jolestar, line 223: We can not use `futures::executor::block_on` here, we need to get the tokio Runtime and use the Runtime block_on. https://tokio.rs/tokio/topics/bridging

crates/rooch-rpc-server/src/service/aggregate_service.rs
```diff
@@ -0,0 +1,167 @@
...
+            balance_info.symbol = coin_info.symbol.clone();
+            balance_info.decimals = coin_info.decimals;
+        }
+
+        Ok(result)
+    }
+}
+
+lazy_static! {
+    static ref RUNTIME: Runtime = tokio::runtime::Builder::new_multi_thread()
```
jolestar, line 145: The tokio runtime needs to pass from the main function to the service. This runtime is another runtime, not in the same runtime which start the service.

baichuan3: It will encounter hung up if only one runtime, when use `futures::executor::block_on`.

I tried use tokio::block_up, but also failed when `call_function` execute.

jolestar: The main function runtime should be built with multi-thread. If this approach works, we can refactor it later.

crates/rooch-framework/sources/coin.move
```diff
@@ -239,6 +239,17 @@ module rooch_framework::coin {
         }
     }
 
+    /// Return coin info handle
+    public fun coin_info_handle(ctx: &StorageContext): Option<ObjectID> {
+        if (account_storage::global_exists<CoinInfos>(ctx, @rooch_framework))
```
jolestar, line 244: I think this should be ensured via the Genesis transaction, so it should always exist.

crates/rooch-rpc-api/src/jsonrpc_types/rooch_types.rs
jolestar: Unify to `type_` and rename via serde?

```rust
     #[serde(rename = "type")] 
     pub type_: StructTagView,
```

baichuan3: already unify to `type_`

crates/rooch-framework/sources/coin.move
baichuan3: done

jolestar (approved):

PR #823, opened 2023-09-18
Auto create account for coin transfer to new address via entry function

### LINKED ISSUE #793: [Framework] Auto create account when coin::transfer to a new address
Because the `account.move` module depends on the `coin.move` module, it can not directly call `account::create_account` in the coin module.

One solution is to migrate the `coin::transfer_entry` function to the `coin_entry.move` module, check if the `to` account does not exist, and auto-create the account in the entry function.

If a module directly calls the `coin::transfer` function, it must ensure the `to` account exists by itself.

bytedeveloperr:
## Summary
This adds a new module `0x3::coin_entry.move`. It auto creates account when `coin::transfer` to a new address

Summary about this PR

- Closes #793


crates/rooch-framework/sources/coin_entry.move
```diff
@@ -0,0 +1,20 @@
+module rooch_framework::coin_entry {
+    use rooch_framework::account;
+    use moveos_std::storage_context::StorageContext;
+    use rooch_framework::coin;
+
+    /// Transfer `amount` of coins `CoinType` from `from` to `to`.
+    /// This public entry function requires the `CoinType` to have `key` and `store` abilities.
+    public entry fun transfer_entry<CoinType: key + store>(
```
jolestar, line 8: Migrate all coin module entry functions to `coin_entry`?  or rename the `coin_entry` module to `tranfer` module? `transfer_entry` => `transfer`.

@baichuan3

baichuan3: I also prefer the alone transfer module.

And there is another problem. There is another `transfer_extend` method in the coin module. It does not implicitly create an account and needs to be commented.

jolestar: 👍， we can also provide transfer NFT in the transfer module.

jolestar: > And there is another problem. There is another `transfer_extend` method in the coin module. It does not implicitly create an account and needs to be commented.

The `transfer` and `transfer_extend` in the coin module can not auto-create the `to` account, as #823 mentions. We can add documents to describe it. If a developer invokes `coin::transfer` in the contract, they should check the 'to' account before invoking it.

bytedeveloperr: What's the conclusion regarding this?

jolestar: As your solution. @bytedeveloperr 

1. `coin_entry.move` to `transfer.move`.
2. `transfer_entry` function to `transfer_coin` function. `transfer_coin` is better than `coin_transfer`?

In the future, we can add `transfer_nft` to the `transfer` module.

bytedeveloperr: Alright, That makes sense. Thank you

jolestar (approved):

Issue #819, opened 2023-09-15
[RPC] RPC handle error after introduce another tokio Runtime


baichuan3:
In https://github.com/rooch-network/rooch/pull/774 for implementing MoveFunctionCall for the RPC Server, the solution is to introduce another tokio Runtime.

Then it is discovered that the RPC server may not be able to directly capture and process an error. The phenomenon is that the call RPC returns a blank, but there is an error log in the console output.

The tokio Runtime implemented in AggrateService, both AggrateService and RPCService will occur the above situation.

**Exception case**
```
let coin_store_handle = coin_module.coin_store_handle(account_addr)?.except("Coin store handle does not exist");
```

error stack
```
thread 'tokio-runtime-worker' panicked at 'Coin store handle does not exist', crates/rooch-rpc-server/src/service/aggregate_service.rs:73:31
```

**Normal case after rewriting**
```
let coin_store_handle = coin_module.coin_store_handle(account_addr)?;
// Handle error due to `coin_store_handle` executed in another tokio Runtime
let coin_store_handle = match coin_store_handle {
    Some(v) => v,
    None => anyhow::bail!("Coin store handle does not exist"),
};
```

baichuan3:
It is determined that it is not caused by the introduction of dual tikio runtime. The same problem existed in the earlier RPC Server. The reason is that methods such as `expect` or `unwrap` will cause panic exceptions, causing the program to exit directly.

jolestar:
I take this.

jolestar:
In the multi-thread executor, if a thread panics, the RPC will return nothing.
We need to fix the panic bug or convert it to an error.

This is not a bug. I close this issue. and record two:
1. #826
2. #827

Issue #827, opened 2023-09-18
Check all `expect` and `unwrap`


jolestar:
Check all `expect` and `unwrap`

PR #830, opened 2023-09-19
handle expect and unwrap for RPC to ensure no panic

baichuan3:
Resolve https://github.com/rooch-network/rooch/issues/819 and part of https://github.com/rooch-network/rooch/issues/827

Handle `expect` and `unwrap` for RPC to ensure no panic occurs, and avoid the RPC returning nothing.


jolestar (approved):

jolestar:
Need to add a test case for the balance RPC.

PR #841, opened 2023-09-20
Feature owen dashboard support session key

yubing744:
## Summary

Summary about this PR

- Closes #issue


wow-sven:
> @jolestar @wow-sven
> 
> Panic occurs when create_session_key_with_multi_scope_entry is called via MetaMask
> 
> c240771b612025ec34b0fc7d7f0ad72

The suggested solution to this issue is to modify the gas of the transaction.

About metamask will cache the sequence number after we restart the service. The transaction triggers the 1002 error, the solution is to start the service by adding --chain-id test

jolestar (approved):

PR #852, opened 2023-09-22
[gh-851] fix testnet url link.

### LINKED ISSUE #851: [bug] couldn't fetch chain id from a remote RPC URL address with MetaMask even switched to the env
### Using `0.0.0.0` port:

Screenshot 2023-09-21 at 20 09 28 cleaned

### Using a remote RPC port:

Screenshot 2023-09-21 at 20 09 51 cleaned

Even switched to `active_env: test` didn't work:

```zsh
envs:
- chain_id: 20230103
  alias: default
  rpc: http://0.0.0.0:50051
  ws: null
- chain_id: 20230102
  alias: test
  rpc: https://seed-testnet.rooch.network/
  ws: null
active_env: test
```

feliciss:
## Summary

Fix Rooch Testnet URL.

- Closes #851

### DIFF
--- a/crates/rooch-rpc-client/src/client_config.rs
+++ b/crates/rooch-rpc-client/src/client_config.rs
@@ -15,7 +15,7 @@ use serde_with::serde_as;
 use std::fmt::{Display, Formatter, Write};
 
 pub const DEFAULT_EXPIRATION_SECS: u64 = 30;
-pub const ROOCH_TEST_NET_URL: &str = "https://seed-testnet.rooch.network/";
+pub const ROOCH_TEST_NET_URL: &str = "https://dev-seed.rooch.network/";
 
 #[serde_as]
 #[derive(Serialize, Deserialize)]


crates/rooch-rpc-client/src/client_config.rs
```diff
@@ -15,7 +15,7 @@ use serde_with::serde_as;
 use std::fmt::{Display, Formatter, Write};
 
 pub const DEFAULT_EXPIRATION_SECS: u64 = 30;
-pub const ROOCH_TEST_NET_URL: &str = "https://seed-testnet.rooch.network/";
+pub const ROOCH_TEST_NET_URL: &str = "https://dev-seed.rooch.network/";
```
baichuan3, line 18: Const name should be consistent with value. Maybe you can add a new Const `ROOCH_DEV_NET_URL`

feliciss: Yes. It needs to be changed but at this time the dev net isn't working.

jolestar (approved):

Issue #755, opened 2023-09-05
[CLI] Make cli stable


jolestar:
To make cli stable, we must ensure compatibility as much as possible.

- [x] To make cli config stable
- [x] To make the keystore stable
- [x] Share the keystore file between different network
- [x] #854
- [x] #855
- [x] #932
- [x] #933 

Encrypting keystore files can affect the developer experience and existing test cases. There are several options to consider:

1. Do not encrypt the keystore file if the user does not provide a password at the `init` command, and then record whether or not it is encrypted in the keystore file.
2. Default the password to an empty string and decrypt it with the empty string if the user does not enter it.

@feliciss @wow-sven @baichuan3

feliciss:
2 seems to be a bad choice.

feliciss:
>  Encrypt the private key in the keystore file

What's the case for encrypting the private key? As I know the keys in key store file are actually key pair bytes.

> Provide a password option for the cli commands

I think the password can be used to generate a seed below:

```rs
let seed = Seed::new(&mnemonic, password_str);
```

As you mentioned in 2, the default seed phrase is empty string "", and we can use the input password as the seed phrase password and use the seed phrase to generate a nonce and a encryption key to encrypt the key pairs.

Ref: #856

jolestar:
we just encrypted the private key in keystore file, not the file.

baichuan3:
discussion:
1. Mark whether the keystore is encrypted and supports single password encryption. If it is not encrypted, you will not be prompted to enter a password.
2,
3. Only supports a single mnemonic word, removes import, and supports passing in the mnemonic word recovery address during rooch init.

PR #788, opened 2023-09-12
[gh-786] Add Ethereum RPC client API to client library.

feliciss:
## Summary

Continuing after #787.

Add Ethereum RPC client API to client library.

- Closes #786

PR #856, opened 2023-09-23
[gh-755] encrypt w/ ChaCha20Poly1305 and password algorithm Argon2id

### LINKED ISSUE #854: Encrypt the private key in the keystore file

### LINKED ISSUE #855: Provide a password option for the cli commands

feliciss:
## Summary

1. Add encryption algorithm `ChaCha20Poly1305` to encrypt the key pair.
2. Add password encryption algorithm `Argon2id` to encrypt password in plaintext and use input password compared with cipher password.
3. Decryption takes `nonce`, `ciphertext`, `tag`, `password` from key store to compare with input password and if successful, retrieve the private key and form a key pair to sign transactions.

ChaCha20Poly1305 is a standardized encryption algorithm (AEAD) widely used by network softwares:

1. https://en.wikipedia.org/wiki/ChaCha20-Poly1305
2. https://datatracker.ietf.org/doc/html/rfc7539

Argon2id is a password encryption algorithm recommended by OWASP in their recent document:

https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html

- Closes #854 and closes #855

### DIFF
--- a/crates/rooch-key/src/key_derive.rs
+++ b/crates/rooch-key/src/key_derive.rs
@@ -40,6 +40,24 @@ pub const DERVIATION_PATH_PURPOSE_SCHNORR: u32 = 44;
 pub const DERVIATION_PATH_PURPOSE_ECDSA: u32 = 54;
 pub const DERVIATION_PATH_PURPOSE_SECP256R1: u32 = 74;
 
+type EncryptionKeyResult = Result<(Vec<u8>, Vec<u8>, Vec<u8>), RoochError>;
+pub struct EncryptionResult {
+    pub hashed_password: String,
+    pub nonce: Vec<u8>,
+    pub ciphertext: Vec<u8>,
+    pub tag: Vec<u8>,
+}
+pub struct GenerateNewKeyPairResult {
+    pub key_pair_type: KeyPairType,
+    pub encryption: EncryptionResult,
+    pub mnemonic: String,
+}
+pub struct GeneratedKeyPair<Addr, KeyPair> {
+    pub address: Addr,
+    pub key_pair: KeyPair,
+    pub result: GenerateNewKeyPairResult,
+}
+
 pub trait CoinOperations<Addr, KeyPair, PrivKey> {
     fn derive_private_key_from_path(
         &self,
@@ -62,7 +80,7 @@ pub trait CoinOperations<Addr, KeyPair, PrivKey> {
         &self,
         private_key: PrivKey,
         password: Option<String>,
-    ) -> Result<(Vec<u8>, Vec<u8>, Vec<u8>), RoochError>;
+    ) -> EncryptionKeyResult;
     fn encrypt_password(
         &self,
         private_key: PrivKey,
@@ -116,9 +134,9 @@ impl CoinOperations<RoochAddress, RoochKeyPair, Ed25519PrivateKey> for KeyPairTy
             let address: RoochAddress = kp.public().into();
             Ok((address, kp.into())) // Cast to KeyPair
         } else {
-            return Err(RoochError::InvalidPasswordError(
+            Err(RoochError::InvalidPasswordError(
                 "Password is invalid".to_owned(),
-            ));
+            ))
         }
     }
 
@@ -135,7 +153,7 @@ impl CoinOperations<RoochAddress, RoochKeyPair, Ed25519PrivateKey> for KeyPairTy
         let mut output_key_material = [0u8; 32]; // Can be any desired size
         Argon2::default()
             .hash_password_into(
-                &password.unwrap_or("".to_owned()).as_bytes(),
+                password.unwrap_or("".to_owned()).as_bytes(),
                 &nonce,
                 &mut output_key_material,
             )
@@ -202,7 +220,7 @@ impl CoinOperations<EthereumAddress, Secp256k1RecoverableKeyPair, Secp256k1Recov
         let child_xprv = XPrv::derive_from_path(seed, &path)
             .map_err(|e| RoochError::SignatureKeyGenError(e.to_string()))?;
         let sk = Secp256k1RecoverablePrivateKey::from_bytes(
-            &child_xprv.private_key().to_bytes().as_slice(),
+            child_xprv.private_key().to_bytes().as_slice(),
         )
         .map_err(|e| RoochError::SignatureKeyGenError(e.to_string()))?;
         Ok(sk)
@@ -238,11 +256,11 @@ impl CoinOperations<EthereumAddress, Secp256k1RecoverableKeyPair, Secp256k1Recov
                     .map_err(|e| RoochError::SignatureKeyGenError(e.to_string()))?,
             );
             let address: EthereumAddress = EthereumAddress::from(kp.public.clone());
-            Ok((address, kp.into())) // Cast to KeyPair
+            Ok((address, kp)) // Cast to KeyPair
         } else {
-            return Err(RoochError::InvalidPasswordError(
+            Err(RoochError::InvalidPasswordError(
                 "Password is invalid".to_owned(),
-            ));
+            ))
         }
     }
 
@@ -256,9 +274,8 @@ impl CoinOperations<EthereumAddress, Secp256k1RecoverableKeyPair, Secp256k1Recov
         let nonce = ChaCha20Poly1305::generate_nonce(&mut OsRng);
 
         // Create a ChaCha20Poly1305 cipher with the key material from password
-        let cipher =
-            ChaCha20Poly1305::new_from_slice(&password.unwrap_or("".to_owned()).as_bytes())
-                .map_err(|e| RoochError::KeyConversionError(e.to_string()))?;
+        let cipher = ChaCha20Poly1305::new_from_slice(password.unwrap_or("".to_owned()).as_bytes())
+            .map_err(|e| RoochError::KeyConversionError(e.to_string()))?;
 
         // Encrypt the private key data to a ciphertext with a tag
         let ciphertext_with_tag = match cipher.encrypt(&nonce, private_key.as_bytes()) {
@@ -378,19 +395,7 @@ pub fn generate_new_key_pair<Addr, KeyPair, PrivKey>(
     derivation_path: Option<DerivationPath>,
     word_length: Option<String>,
     password: Option<String>,
-) -> Result<
-    (
-        Addr,
-        KeyPair,
-        KeyPairType,
-        String,
-        Vec<u8>,
-        Vec<u8>,
-        Vec<u8>,
-        String,
-    ),
-    anyhow::Error,
->
+) -> Result<GeneratedKeyPair<Addr, KeyPair>, anyhow::Error>
 where
     KeyPairType: CoinOperations<Addr, KeyPair, PrivKey>,
 {
@@ -412,16 +417,24 @@ where
 
     let (address, key_pair) = key_pair_type.derive_key_pair_from_ciphertext(ciphertext.clone())?;
 
-    Ok((
-        address,
-        key_pair,
-        key_pair_type,
+    let encryption = EncryptionResult {
         hashed_password,
         nonce,
         ciphertext,
         tag,
-        mnemonic.phrase().to_string(),
-    ))
+    };
+
+    let result = GenerateNewKeyPairResult {
+        key_pair_type,
+        encryption,
+        mnemonic: mnemonic.phrase().to_string(),
+    };
+
+    Ok(GeneratedKeyPair {
+        address,
+        key_pair,
+        result,
+    })
 }
 
 // Verify the password against Argon2
@@ -446,7 +459,7 @@ pub fn decrypt_private_key(
     let mut output_key_material = [0u8; 32]; // Can be any desired size
     Argon2::default()
         .hash_password_into(
-            &password.unwrap_or("".to_owned()).as_bytes(),
+            password.unwrap_or("".to_owned()).as_bytes(),
             &nonce,
             &mut output_key_material,
         )
@@ -457,7 +470,7 @@ pub fn decrypt_private_key(
 
     // Concatenate the tag and the ciphertext to reconstruct ciphertext_with_tag
     let mut ciphertext_with_tag = Vec::with_capacity(tag.len() + ciphertext.len());
-    ciphertext_with_tag.extend_from_slice(&ciphertext);
+    ciphertext_with_tag.extend_from_slice(ciphertext);
     ciphertext_with_tag.extend_from_slice(&tag);
 
     // Decrypt the ciphertext_with_tag to a private key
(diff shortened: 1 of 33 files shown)


crates/rooch-key/src/key_derive.rs
```diff
@@ -148,12 +170,14 @@ pub fn generate_new_key_pair<Addr, KeyPair>(
...
     word_length: Option<String>,
+    password: Option<String>,
 ) -> Result<(Addr, KeyPair, KeyPairType, String), anyhow::Error>
 where
     KeyPairType: CoinOperations<Addr, KeyPair>,
 {
     let mnemonic = Mnemonic::new(parse_word_length(word_length)?, Language::English);
-    let seed = Seed::new(&mnemonic, "");
+    let password_str = password.as_deref().unwrap_or("");
+    let seed = Seed::new(&mnemonic, password_str);
```
jolestar, line 180: I think using the `password_str` as part of the seed is not a good idea for CLI.

1. The user needs to save the mnemonic and password and use the mnemonic and password to recover the account. If use a wrong password, will recover a new account.
2. The user can not change the password.

Some wallet provides a `salt` to mnemonic, like this, and the user needs to know it is salt, not the password.

jolestar (commented):
When to unlock the private key?

crates/rooch-key/src/key_derive.rs
feliciss: > The user needs to save the mnemonic and password and use the mnemonic and password to recover the account. If use a wrong password, will recover a new account.

This is a common practice for hardware and software authentication schemes e.g. MetaMask with mnemonic and password, or BitBox02 with password option. Here, password is merely an option for the user and the user can opt in to choose whether to remember it.

> The user can not change the password.

If changing password would generate a new account, it is the common practice for the bitcoin software bc we are using bip39 mnemonic word Rust lib. In other words, changing password would mean collapsing the existing seed infrastructure at the first time generated. There's an article introducing the passphrase trade-offs: https://help.blockstream.com/hc/en-us/articles/8712301763737-What-is-a-BIP39-passphrase-#h_01G8K3SK5VBHBYZSX5XS4309YG.

I think it's only a choice made by a user. 

> Some wallet provides a salt to mnemonic, like this, and the user needs to know it is salt, not the password.

The accurate answer is passphrase: https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#user-content-From_mnemonic_to_seed. Salt is from a password with mnemonic string according to the original code:

```rs
let salt = format!("mnemonic{}", password);
```

We can add documents about how the seed phrase is constructed.

feliciss:
> When to unlock the private key?

When decrypting private key, the user would need to provide:

1. Mnemonic phrase
2. Optional password

To decrypt the private key and sign the transaction, by design.

feliciss:
Discussion:

1. Remove salt, default to "".
2. Provide an optional password when init, and record it using `argon2` hash to key store for later verification.
3. Provide a clap to enter the password in asterisk.

crates/rooch/src/commands/init.rs
```diff
@@ -23,6 +24,9 @@ pub struct Init {
     /// Command line input of custom server URL
     #[clap(short = 's', long = "server-url")]
     pub server_url: Option<String>,
+    /// Whether a password should be provided.
+    #[clap(short = 'p', long = "password")]
+    password_required: bool,
```
jolestar, line 29: Make `password_required` default value to `true`? And we can set it to `false` in the test case.

So, if we want to let `password_required` accept value, it should be `Option<bool>`.


```
#[clap(short = 'p', long = "password")]
password_required: Option<bool>,
```

feliciss: I've added it.

feliciss:
TODO: do some refactor to key store and retrieve private keys from key store params in signing functions.

It may take some days and if it isn't included in Q3 release, I will handle this issue at a stable pace.

feliciss:
This design may need further refactor:

1. `nonce` shouldn't be used twice for Argon2 and ChaCha20Poly1305 cipher in `encrypt_private_key`.
2. In `encrypt_password`, considering generate a random `salt` or encrypt current `private_key` as salt.

crates/rooch-key/src/key_derive.rs
```diff
@@ -31,49 +41,296 @@ pub const DERVIATION_PATH_PURPOSE_SCHNORR: u32 = 44;
...
+        let ciphertext = ciphertext_with_tag[..ciphertext_with_tag.len() - 16].to_vec();
+
+        // Extract the tag (last 16 bytes)
+        let tag = {
+            let start = ciphertext_with_tag.len() - 16;
+            let end = ciphertext_with_tag.len();
+            let mut tag = Vec::with_capacity(16);
+            tag.extend_from_slice(&ciphertext_with_tag[start..end]);
+            tag
+        };
```
jolestar, line 186: Why split the tag? or include the tag in the ciphertext? I can not find the use case of the tag.

feliciss: > Why split the tag? or include the tag in the ciphertext? I can not find the use case of the tag.

The tag is useless here, and why did I split it is cause the private key should be 32 bytes. The ciphertext is composed of 32 bytes encrypted text and 16 bytes tag thus 48 bytes. In order to achieve key derivations, i.e. parsing the encrypted cipertext to private keys and vice versa, it needs to be split.

feliciss: > [2.5](https://www.rfc-editor.org/rfc/rfc7539#section-2.5).  The Poly1305 Algorithm
> 
>    Poly1305 is a one-time authenticator designed by D. J. Bernstein.
>    Poly1305 takes a 32-byte one-time key and a message and produces a
>    16-byte tag.  This tag is used to authenticate the message.

https://www.rfc-editor.org/rfc/rfc7539

jolestar: Add a comment to the tag.

jolestar:
Merge this PR first or wait to implement the password option to decrypt private key for commands, like `rooch move run`, `rooch move publish`, etc.

feliciss:
> Merge this PR first or wait to implement the password option to decrypt private key for commands, like `rooch move run`, `rooch move publish`, etc.

You can merge this PR first and then I will refactor key store, yaml, etc. and test the signature verification feature with encrypted key pair.

crates/rooch/src/commands/move_cli/commands/publish.rs
```diff
@@ -66,6 +70,15 @@ impl CommandAction<ExecuteTransactionResponseView> for Publish {
...
         let context = self.context_options.build().await?;
 
+        let password = if self.password_required == Some(false) {
+            // Use an empty password if not required
+            String::new()
+        } else {
+            // Prompt for a password if required
+            rpassword::prompt_password("Enter a password to encrypt the keys in the rooch keystore. Press return to have an empty value: ").unwrap()
+        };
+        println!("Your password is {}", password);
```
jolestar, line 80: Do not need to print the password when decrypting the keypair.

crates/rooch/src/commands/move_cli/commands/publish.rs
```diff
@@ -48,6 +48,10 @@ pub struct Publish {
     /// `moveos_std::account_storage::publish_modules_entry`
     #[clap(long, parse(from_flag))]
     pub by_move_action: bool,
+
+    /// Whether a password should be provided
+    #[clap(short = 'p', long = "password")]
+    password_required: Option<bool>,
```
jolestar, line 54: Is it possible to auto-detect that if the private key is encrypted, then prompt the password input?

jolestar:
Some tests need to be fixed.

crates/rooch/src/commands/move_cli/commands/publish.rs
feliciss: The private key is always encrypted. When the user does not enter the password, the password defaults to "".

crates/rooch/src/commands/move_cli/commands/publish.rs
feliciss: Resolved.

feliciss:
This command will fail:

```rs
rooch move run --function 0x3::empty::empty --sender-account 0xac00d67a15ae97af88aff69cd80befdfe430330a08a6092163de84753cb220c4 --session-key 975a3d052b5ee4bdcc7c8c8a16ae878ef427b4fe6ecb1aa9088b43645111ae20 --password false
```

The session key `975a3d052b5ee4bdcc7c8c8a16ae878ef427b4fe6ecb1aa9088b43645111ae20` seems not have been saved properly, and it results the error:

`Sign message error: signature error: Cannot find SessionKey for address: [0xac00d67a15ae97af88aff69cd80befdfe430330a08a6092163de84753cb220c4]`

I didn't know the session key's design, but is it required to save it as key pair or in `session_keys: {}`?

crates/testsuite/features/cmd.feature
```diff
@@ -10,30 +10,28 @@ Feature: Rooch CLI integration tests
...
+      Then cmd: "account list --password false"
+      Then cmd: "account import --mnemonic-phrase "fiber tube acid imitate frost coffee choose crowd grass topple donkey submit" --password false"
+      Then cmd: "account update --address 0xebf29d2aed4da3d2e13a32d71266a302fbfd5ceb3ff1f465c006fa207f1789ce --mnemonic-phrase "spike air embody solid upper grow mule slender shrimp suggest pride young" --password false"
+      Then cmd: "account nullify --address 0xebf29d2aed4da3d2e13a32d71266a302fbfd5ceb3ff1f465c006fa207f1789ce --password false"
 
       # session key
-      Then cmd: "session-key create --sender-account {default} --scope 0x3::empty::empty"
-      Then cmd: "move run --function 0x3::empty::empty --sender-account {default} --session-key {{$.session-key[-1].authentication_key}}"
+      Then cmd: "session-key create --sender-account {default} --scope 0x3::empty::empty --password false"
+      Then cmd: "move run --function 0x3::empty::empty --sender-account {default} --session-key {{$.session-key[-1].authentication_key}} --password false"
```
jolestar, line 21: It seems like the `session-key` command output is broken, so this command can not get the `--session-key` argument.

crates/rooch-key/src/key_derive.rs
```diff
@@ -31,49 +40,291 @@ pub const DERVIATION_PATH_PURPOSE_SCHNORR: u32 = 44;
 pub const DERVIATION_PATH_PURPOSE_ECDSA: u32 = 54;
 pub const DERVIATION_PATH_PURPOSE_SECP256R1: u32 = 74;
 
-pub trait CoinOperations<Addr, KeyPair> {
-    fn derive_key_pair_from_path(
+type EncryptionKeyResult = Result<(Vec<u8>, Vec<u8>, Vec<u8>), RoochError>;
+
+pub trait CoinOperations<Addr, KeyPair, PrivKey> {
```
jolestar, line 45: The `CoinOperations` name is not suitable yet.

crates/rooch-key/src/key_derive.rs
```diff
@@ -31,49 +40,291 @@ pub const DERVIATION_PATH_PURPOSE_SCHNORR: u32 = 44;
...
+        // https://www.rfc-editor.org/rfc/rfc7539
+        let tag = {
+            let start = ciphertext_with_tag.len() - 16;
+            let end = ciphertext_with_tag.len();
+            let mut tag = Vec::with_capacity(16);
+            tag.extend_from_slice(&ciphertext_with_tag[start..end]);
+            tag
+        };
+
+        Ok((nonce.to_vec(), ciphertext, tag))
```
jolestar, line 307: Most lines repeat with `CoinOperations<RoochKeyPair>`

crates/rooch-key/src/keystore.rs
```diff
@@ -47,30 +59,39 @@ pub enum Keystore<K: Ord, V> {
 }
 
 #[enum_dispatch]
-pub trait AccountKeystore<Addr: Copy, PubKey, KeyPair, Sig, TransactionData>: Send + Sync {
+pub trait AccountKeystore<Addr: Copy, PubKey, KeyPair, Sig, TransactionData, PrivKey>:
```
jolestar, line 62: This should be part of https://github.com/rooch-network/rooch/pull/788. As discussed in the PR, Maybe we do not need to distinguish the Keystore via the KeyPair type and TransactionData.

crates/rooch/src/commands/init.rs
```diff
@@ -109,19 +113,37 @@ impl CommandAction<String> for Init {
...
+                println!("Secret Recovery Phrase : [{}]", result.result.mnemonic);
                 let dev_env = Env::new_dev_env();
                 let active_env_alias = dev_env.alias.clone();
                 ClientConfig {
                     keystore,
                     envs: vec![env, dev_env],
-                    active_address: Some(new_address),
+                    password: Some(result.result.encryption.hashed_password),
+                    nonce: Some(result.result.encryption.nonce.encode_hex()),
+                    ciphertext: Some(result.result.encryption.ciphertext.encode_hex()),
```
jolestar, line 144: `keystore` file should be an independent file, not part of client_config.

crates/rooch-key/src/keystore.rs
feliciss: Will refactor it in #788 then.

crates/rooch/src/commands/init.rs
feliciss: Removed.

crates/rooch-key/src/key_derive.rs
feliciss: Simplified.

crates/rooch-key/src/key_derive.rs
feliciss: Renamed.

crates/rooch/src/commands/session_key/commands/create.rs
```diff
@@ -36,6 +41,15 @@ impl CreateCommand {
...
         let mut context = self.context_options.build().await?;
 
+        let password = if self.password_required == Some(false) {
+            // Use an empty password if not required
+            String::new()
+        } else {
+            // Prompt for a password if required
+            rpassword::prompt_password("Enter a password to encrypt the keys in the rooch keystore. Press return to have an empty value: ").unwrap()
+        };
+        println!("Your password is {}", password);
```
jolestar, line 51: Do not print the password when decrypting.

jolestar (approved):

jolestar:
We can merge this PR first and discuss some refactors later:

1. Should every private key have a new password? Or the whole keystore uses one password.
2. Is it possible to eliminate the `--password false` options?

feliciss:
> We can merge this PR first and discuss some refactors later:
> 
> 1. Should every private key have a new password? Or the whole keystore uses one password.
> 2. Is it possible to eliminate the `--password false` options?

1. Every key has a password, but the user can ignore it. There's an issue with `account list` when different keys have different passwords.
2. Yes. We can eliminate the option and make every key encrypted with `""` string.

crates/testsuite/features/cmd.feature
feliciss: This issue has been fixed.

crates/rooch/src/commands/session_key/commands/create.rs
feliciss: I leaved it with empty string for a new password.

jolestar:
> 2\. Yes. We can eliminate the option and make every key encrypted with `""` password string.

I didn't mean to remove the password prompt. But it is OK. I will merge this PR first, and we discuss how to refactor it.

Issue #866, opened 2023-09-26
[Cleanup] Clear WalletApi API from RPC


jolestar:
https://github.com/rooch-network/rooch/blob/c815906a23a65f8e211419f4c3b3b41a878a7272/crates/rooch-rpc-api/src/api/wallet_api.rs#L8-L20

Issue #857, opened 2023-09-23
[Cli] Process the corresponding relationship between env and chainid in Cli and RPC Server and optimize the experience


baichuan3:
Currently, Cli and RPC Server each have an environment. Cli connects RPC and signature (including chainid information) by configuring env. When RPC Server starts, it needs to specify different networks. The two environments are separated.

discussion https://discord.com/channels/1078938449974935592/1154633667105271888/1154634509954850948

other relative issues from @feliciss 

1. When adding a new environment, verify the `rpc`, `alias` and `chain_id` don't collapse with existing ones. 
2. Once a new env is added, the chain id is still shown as the last element and it's not updated. For example:

> Chain ID
> 20230103
> The RPC URL you have entered returned a different chain ID (20230102). Please update the Chain ID to match the RPC URL of the network you are trying to add.

Config:

```zsh
envs:
- chain_id: 20230104
  alias: default
  rpc: http://0.0.0.0:50051
  ws: null
- chain_id: 20230103
  alias: dev
  rpc: https://devnet-seed.rooch.network/
  ws: null
- chain_id: 20230102
  alias: test
  rpc: https://testnet-seed.rooch.network/
  ws: null
- chain_id: 20230103
  alias: devnet
  rpc: http://0.0.0.0:50051/
  ws: null
active_env: devnet
```

3. After switching to `devnet`, command `rooch server start -n test` and `rooch server start -n dev` result in a panic error:

> thread 'main' panicked at 'Invalid sequencer order', crates/rooch- sequencer/src/actor/sequencer.rs:45:24

4. After a new env is added, starting rooch server will not be added a custom chain id:

> rooch server start -n devnet
> error: Invalid value "devnet" for '--chain-id <CHAIN_ID>': Invalid Custom chain id devnet, custom chain id format is: chain_name:chain_id

baichuan3:
1. Automatically switch env when use start server, if env is local or dev seed
2. Verify env alias uniqueness
3. The default local network use fixed directory, modify the test case to support temporary directory
4. Remove wallet RPC

Issue #869, opened 2023-09-26
[RPC] Upgrade jsonrpsee to solved HttpClient Port force check


baichuan3:
The currently used jsonrpsee version is 0.16. When the rpc client creates the HttpClientBuilder, if the seed node `https:://dev-seed.rooch.network/` is passed in, an error will be thrown.

```
Error: Networking or low-level protocol error: Invalid Url: Port number is missing in the URL
```

The reason is that jsonrpsee 0.16 and previous versions have a bug when constructing the http client, which requires the port to be forced to be passed in. The 0.17 version fixes the bug and automatically recognizes https and http, and then adds port 443 or 80.

However, starting from version 0.17, the HttpClient and RpcResult structures of jsonrpsee have undergone major changes. 
1. HttpClient has added generics.

```
/// HTTP Transport Client.
#[derive(Debug, Clone)]
pub struct HttpTransportClient<S> {
/// Target to connect to.
target: String,
/// HTTP client
client: S,
/// Configurable max request body size
max_request_size: u32,
/// Configurable max response body size
max_response_size: u32,
/// Max length for logging for requests and responses
///
/// Logs bigger than this limit will be truncated.
max_log_length: u32,
/// Custom headers to pass with every request.
headers: HeaderMap,
}
```

2. RpcResult is adjusted from std::result::Result<T, Error> to std::result::Result<T, ErrorObjectOwned>

When upgrading jsonrpsee, need to refactor the RPC server and RPC client before. Also ensure compatibility of RPC returns.

Workaround solution, display specified port, such as `https://dev-seed.rooch.network:443/`

PR #870, opened 2023-09-26
Make env match network when server start And default local network use fixed directory

baichuan3:
resolve https://github.com/rooch-network/rooch/issues/857 and resolve https://github.com/rooch-network/rooch/issues/866

## Summary

1. Automatically switch env when use start server, if env is local or dev seed
2. Verify env alias uniqueness
3. The default local network use fixed directory, modify the test case to support temporary directory
4. Remove wallet RPC


crates/rooch-rpc-client/src/client_config.rs
```diff
@@ -15,8 +15,8 @@ use serde_with::serde_as;
 use std::fmt::{Display, Formatter, Write};
 
 pub const DEFAULT_EXPIRATION_SECS: u64 = 30;
-pub const ROOCH_DEV_NET_URL: &str = "https://dev-seed.rooch.network/";
-pub const ROOCH_TEST_NET_URL: &str = "https://test-seed.rooch.network/";
+pub const ROOCH_DEV_NET_URL: &str = "https://dev-seed.rooch.network:443/";
```
feliciss, line 18: The default https URL port is 443? There's no need to add `443`.

crates/rooch-rpc-client/src/client_config.rs
```diff
@@ -15,8 +15,8 @@ use serde_with::serde_as;
...
+pub const ROOCH_TEST_NET_URL: &str = "https://test-seed.rooch.network:443/";
```
feliciss, line 19: Also there.

feliciss (approved):
LGTM. Some backend ports may need to be modified to original format without `:port`.

crates/rooch-rpc-client/src/client_config.rs
baichuan3: It is related to this issue https://github.com/rooch-network/rooch/issues/869. It is caused by a bug in jsonrpsee and needs to be upgraded. I tried the upgrade and found that it is quite troublesome. The upgrade requires the refactor of RPC result and RPC client. This version does not have enough time and will be dealt with in the next version.

PR #908, opened 2023-10-04
Optimize moveos_stdlib::table 

### LINKED ISSUE #320: [Bug] moveos_stdlib::table::destroy can be called with non-empty tables.
Here is a test case to trigger the bug:
```
  fun test_destroy_nonempty_table(account: signer) {
      let sender = std::signer::address_of(&account);
      let tx_context = moveos_std::tx_context::new_test_context(sender);
      let t = new<u64, u8>(&mut tx_context);
      let key: u64 = 100;
      add(&mut t, key, 1);

      destroy(t);
  }
```

### LINKED ISSUE #342: [MoveosStd] Table need provide destroy_empty_table, is_empty and length function

### LINKED ISSUE #899: [MoveosStd] Support drop Table<K,V> if the V has drop ability
If the value of Table has the `drop` ability, we can support destroying the table.

There are some relative issues:

1. #320 
2. #342

pause125:
## Summary

1. Destroy the table if the value of table has the drop ability. Close #899 
2. Add interface of table::length and table::is_empty, support destroy empty table. Close #342
3. Fix bug when table::destroy non-empty tables. Close #320

### DIFF
--- a/moveos/moveos-stdlib/moveos-stdlib/sources/raw_table.move
+++ b/moveos/moveos-stdlib/moveos-stdlib/sources/raw_table.move
@@ -87,12 +87,12 @@ module moveos_std::raw_table {
     }
 
     /// Returns the size of the table, the number of key-value pairs
-    public fun length(table_handle: &ObjectID): u64 {
+    public(friend) fun length(table_handle: &ObjectID): u64 {
         box_length(table_handle)
     }
 
     /// Returns true iff the table is empty (if `length` returns `0`)
-    public fun is_empty(table_handle: &ObjectID): bool {
+    public(friend) fun is_empty(table_handle: &ObjectID): bool {
         length(table_handle) == 0
     }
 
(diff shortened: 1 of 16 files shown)


moveos/moveos-stdlib/moveos-stdlib/sources/raw_table.move
```diff
@@ -118,7 +130,9 @@ module moveos_std::raw_table {
 
     native fun remove_box<K: copy + drop, V, B>(table_handle: &ObjectID, key: K): Box<V>;
 
-    native fun destroy_empty_box(table_handle: &ObjectID);
+    native fun destroy_empty_box_unchecked(table_handle: &ObjectID);
 
     native fun drop_unchecked_box(table_handle: &ObjectID);
```
jolestar, line 135: Merge the `drop_unchecked_box` and `destroy_empty_box_unchecked`?

jolestar (approved):

Issue #906, opened 2023-10-03
[Cli] Newly account will be overwritten in keystore file and can't list


baichuan3:
Find a bug about `account account create` and `account account list`, the newly created account will be overwritten in the keystore and cannot be displayed in the account list. It will be located and resolved in a separate PR.

jolestar:
Make the `account account create` to a local operation; do not execute the transaction.

PR #910, opened 2023-10-04
refactor wallet context and fixup account cli

baichuan3:
resolve https://github.com/rooch-network/rooch/issues/906

Summary
1. Remove the keystore from the client config, leaving only the keystore path in the client config
2. Load keystore in Wallet Context, keystore is used as an attribute of WalletContext
3. Make the account account create to a local operation; do not execute the transaction

### DIFF
--- a/crates/rooch-rpc-client/src/wallet_context.rs
+++ b/crates/rooch-rpc-client/src/wallet_context.rs
@@ -11,7 +11,7 @@ use rooch_config::config::{Config, PersistedConfig};
 use rooch_config::server_config::ServerConfig;
 use rooch_config::{rooch_config_dir, ROOCH_CLIENT_CONFIG, ROOCH_SERVER_CONFIG};
 use rooch_key::keypair::KeyPairType;
-use rooch_key::keystore::AccountKeystore;
+use rooch_key::keystore::{AccountKeystore, FileBasedKeystore, Keystore};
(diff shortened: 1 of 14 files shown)


crates/rooch-rpc-client/src/wallet_context.rs
```diff
@@ -27,18 +27,19 @@ use std::sync::Arc;
 use std::time::Duration;
 use tokio::sync::RwLock;
 
-pub struct WalletContext {
+pub struct WalletContext<K: Ord, V> {
     client: Arc<RwLock<Option<Client>>>,
-    pub client_config: PersistedConfig<ClientConfig<RoochAddress, RoochKeyPair>>,
+    pub client_config: PersistedConfig<ClientConfig>,
     pub server_config: PersistedConfig<ServerConfig>,
+    pub keystore: Keystore<K, V>,
```
jolestar, line 34: We have not found a suitable approach to supporting the Ethereum transaction in CLI. We can make the `Keystore<K<V>` to `Keystore<RoochAddress, RoochKeyPair>`. After we support other KeyPairType to the WalletContext, we can refactor it later.

jolestar (approved):

crates/rooch-rpc-client/src/wallet_context.rs
jolestar: I merge this PR first, we can refactor it later.

PR #942, opened 2023-10-10
[gh-933] use one password for rooch keystore

### LINKED ISSUE #755: [CLI] Make cli stable

### LINKED ISSUE #933: [CLI] Use one password for keystore
Part of #755 

1. Use one password for the keystore. Not every private key has one password.
2. Record the hashed password in the keystore.
3. Record if the password is empty in the keystore. If the password is empty, do not prompt the password input.

feliciss:
## Summary

1. Use one password for the keystore. Not every private key has one password.
2. Record the hashed password in the keystore.
3. Record if the password is empty in the keystore. If the password is empty, do not prompt the password input.'

In addition,

4. Providing an option to enter password when initiating rooch keystore.
5. Removing KeyPairType.

Todo list / to discuss:

- [ ] Providing a function to convert Rooch address and transaction into Ethereum address & tx. 

- Closes #933 in part of #755


crates/rooch-rpc-client/src/client_config.rs
```diff
@@ -21,6 +21,8 @@ pub const ROOCH_TEST_NET_URL: &str = "https://test-seed.rooch.network:443/";
 #[derive(Serialize, Deserialize, Debug)]
 pub struct ClientConfig {
     pub keystore_path: PathBuf,
+    pub password_hash: Option<String>,
+    pub is_password_empty: bool,
```
jolestar, line 25: `password_hash` and `is_password_empty` should be part of `ClientConfig` or `KeyStore`?

feliciss: Okay. I overlooked it. It looks like it can be moved to `BaseKeyStore`.

crates/rooch-types/src/keypair_type.rs
```diff
@@ -1,61 +0,0 @@
...
-)]
-#[cfg_attr(any(test, feature = "fuzzing"), derive(Arbitrary))]
-#[strum(serialize_all = "lowercase")]
-pub enum KeyPairType {
-    RoochKeyPairType,
-    EthereumKeyPairType,
-}
-
-impl KeyPairType {
-    const ROOCH_KEY_PAIR_TYPE: &str = "rooch";
```
baichuan3, line 37: Keystore only supports rooch, and will no longer consider supporting other types such as ethereum?

crates/rooch-key/src/keystore.rs
```diff
@@ -38,237 +30,153 @@ use std::fs::File;
...
+    fn get_public_key(&self, password: Option<String>) -> Result<PublicKey, anyhow::Error>;
     fn get_key_pairs(
         &self,
-        address: &Addr,
+        address: &RoochAddress,
         password: Option<String>,
-    ) -> Result<Vec<KeyPair>, anyhow::Error>;
-    fn get_key_pair_by_type_password(
+    ) -> Result<Vec<RoochKeyPair>, anyhow::Error>;
+    fn get_key_pair_by_password(
```
baichuan3, line 61: Would `get_key_pair_by_password ` -> `get_key_pair_with_password ` be more appropriate?

feliciss: Yep. It's just a naming.

crates/rooch-types/src/keypair_type.rs
feliciss: Yep.

feliciss: It's not supported for now.

jolestar (approved):
I merge this first, @feliciss you can resolve the comment in new PR.

PR #935, opened 2023-10-09
[Framework & RPC] Implement ObjectRef, refactor CoinStore and balance RPC

jolestar:
## Summary

Implement ObjectRef, refactor CoinStore, and balance RPC.

part of #901 #915 

1. Add `borrow_from_global` `borrow_mut_from_global` to `raw_table`, allowing borrowing Objects directly from global storage without `Context`.
2. Introduce ObjectRef<T>. A struct that holds ObjectRef<T> can directly operate T, and is not restricted by private_generic. 
3. Refactor Coin and CoinStore with ObjectRef. Now, every CoinStore is an Object, And Coin<T> is a hot potato type that can only be handled via coin module, and when storing Coin<T> to CoinStore, the Coin<T> becomes Balance.
4. Remove generic type from CoinInfo<T> and CoinStore<T>, add coin_type field to them.
5. Refactor Balance query logic.

TODO:

1. Refactor `coin.move`, and plan to split it to `coin.move`, `coin_store.move`, `account_coin_store.move`
2. Refactor some examples with ObjectRef and use ObjectRef to replace ObjectID.
3. Now, ObjectRef can be `store`,`copy`,`drop`, do we need to track the reference count for Object? This can ensure developer always can get the Object via ObjectRef.

Issue #901, opened 2023-10-02
[MoveosStd] Introduce ObjectRef


jolestar:
## Motivation

Currently, The Object in Rooch is a Hot Potato, and it can't be `store`,`drop`,`copy`.

```move
struct Object<T>{
   id: ObjectID,
   owner: address,
   value: T,
}
```

We can not use the Object as a field type.

```move
struct Hero has key{
  equipment: Object<Equipment>,
}
```

So, we should introduce ObjectRef<T> to support this application scenario.

## Design

```move
struct ObjectRef<phantom T>{
   id: ObjectID,
}
```

The ObjectRef only contains the ObjectID and does not include the value.

```move
struct Object<T> {
  id: ObjectID,
  owner: address,
  ref_count: u64,
  value: T,
}
```

We add a `ref_count` field to the Object and track the ObjectRef count of the Object. If the `ref_count` > 0, the Object can not be unpacked(destroyed).

Add some native function to support borrowing Object from ObjectRef.

```move

module moveos_std::object_ref{
  
  native public fun borrow<T:key>(ref: &ObjectRef<T>) : &Object<T>;
  native public fun borrow_mut<T:key>(ref: &ObjectRef<T>) : &mut Object<T>;

}

```

Now, we can use ObjectRef as the field type.

```move
struct Hero has key{
  equipment: ObjectRef<Equipment>,
}
```

## Reference

1. Previous discussion #51 
2. #912

jolestar:
Implement `ObjectRef`, but do not implement the `ref_count`. The `ObjectRef` has the `drop` ability.

PR #948, opened 2023-10-11
[MoveosStd] Refactor ObjectRef, context:new_object and context::remove_object

jolestar:
## Summary

[MoveosStd] Refactor ObjectRef, context:new_object and context::remove_object

1. Make `ObjectRef` can not `copy`.
2. `context::new_object<T>` auto add Object to storage, and return ObjectRef<T>
3. `context::remove_object<T>` auto unpack the Object, and return `(id, owner, T)`
4. Refactor examples to use ObjectRef.

Continue #935 

Part of #901 

@geometryolife Some documents and example readme need to be updated.
@wubuku the code generator needs to be updated with new Object and ObjectRef API.


moveos/moveos-stdlib/moveos-stdlib/sources/context.move
```diff
@@ -131,15 +121,26 @@ module moveos_std::context {
...
     /// Create a new Object with owner
-    public fun new_object_with_owner<T: key>(self: &mut Context, owner: address, value: T): Object<T> {
-        object::new(&mut self.tx_context, owner, value)
+    /// Add the Object to the global object storage and return the ObjectRef
+    public fun new_object_with_owner<T: key>(self: &mut Context, owner: address, value: T): ObjectRef<T> {
+        let object_id = fresh_object_id(self);
+        new_object_with_id(self, object_id, owner, value)
+    }
+
+    public(friend) fun new_object_with_id<T: key>(self: &mut Context, id: ObjectID, owner: address, value: T) : ObjectRef<T> {
```
baichuan3, line 139: If new an object and return `ObjectRef `by default, need to additionally call the `object_ref::id` method to get the object_id. Looking at the example,  only need the object_id in many cases, which will increase the burden to the developer.

WGB5445:
ObjectRef is a way to divide Object permissions. ObjectRef allows anyone to operate on the Object body. Should we abstract the permissions of sending and extending Objects, so that Objects are more flexible?

moveos/moveos-stdlib/moveos-stdlib/sources/context.move
jolestar: Previously, developers needed to call `object::id(&obj)` to get ObjectID, now, replaced by `object_ref::id(&obj_ref)`.

jolestar:
> ObjectRef is a way to divide Object permissions. ObjectRef allows anyone to operate on the Object body. Should we abstract the permissions of sending and extending Objects, so that Objects are more flexible?

Do you have some ideas about this?

jolestar:
> Here, we limit the sending of Objects of type to the module of T. We can modify Objects so that they can be sent anywhere.
> 
> https://github.com/rooch-network/rooch/blob/aa85fd70355e8b9ad77a2004f22301185fe74489/moveos/moveos-stdlib/moveos-stdlib/sources/object.move#L56-#L60
> 
> We can limit the sending permissions of Objects by adding some fields. Only users who have TransferRef can send this Object.
> 
> ```
> struct Object {
>     allow_ungated_transfer: bool,
> }
> ```
> 
> As a result, all Objects can be sent arbitrarily, and the creator can close the sending of some Objects. Here is the implementation of Aptos, which I think is very good.
> 
> https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/aptos-framework/sources/object.move#L400-L410

We can provide a `public_transfer` to `Object`, if needed. The `public_transfer` requires the `T` has key + store ability but does not require `private_generics(T)` like Sui.

moveos/moveos-stdlib/moveos-stdlib/sources/context.move
```diff
@@ -105,21 +106,10 @@ module moveos_std::context {
...
-        let object_id = object_ref::into_id(object_ref);
-        storage_context::remove<T>(&mut self.storage_context, object_id)
-    }
-
-    #[private_generics(T)]
-    /// Add object to object store
-    public fun add_object<T: key>(self: &mut Context, obj: Object<T>) {
-        storage_context::add<T>(&mut self.storage_context, obj)
+    /// Remove object from object store, and unpack the Object
+    public fun remove_object<T: key>(self: &mut Context, object_id: ObjectID): (ObjectID, address, T) {
```
baichuan3, line 110: After remove object, the original ObjectRef reference still exists. If call `object_ref::borrrow` or `object_ref::borrrow_mut`, the underlying raw_table will throw a NotFound error code. The Rust life cycle management mechanism fails in this dimension. 

 Should we consider adding track ObjectRef?

baichuan3 (approved):

moveos/moveos-stdlib/moveos-stdlib/sources/context.move
jolestar: Record a TODO about this

https://github.com/rooch-network/rooch/blob/e79f988fbac899e7c60fdc745fac091fd9210a9f/moveos/moveos-stdlib/moveos-stdlib/sources/object_ref.move#L22-L25

Currently, before using ObjectRef, the dev needs to call `object_ref::exist_object` to ensure the object exists.

moveos/moveos-stdlib/moveos-stdlib/sources/context.move
```diff
@@ -105,21 +106,10 @@ module moveos_std::context {
...
+        let obj = storage_context::remove<T>(&mut self.storage_context, object_id);
+        object::unpack_internal(obj)
     }
 
     public fun contains_object(self: &Context, object_id: ObjectID): bool {
```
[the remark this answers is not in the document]
jolestar: Done

Issue #932, opened 2023-10-09
[CLI] Use one mnemonic phrase for keystore


jolestar:
Part of #755 

1. Support input mnemonic phrase when `rooch init`.
2. Record the mnemonic phrase in the keystore file.
3. Reuse the mnemonic phrase when `rooch account create`.
4. Remove `rooch account import` command. We do not support multi mnemonic phrase keystore now.
5. Encrypt the mnemonic phrase with keystore password.

PR #950, opened 2023-10-11
Use one mnemonic phrase for keystore

baichuan3:
resolve https://github.com/rooch-network/rooch/issues/932

Summary
1. Refactor the keystore module. The original module exceeded 1600 lines and the keystore was split into multiple modules.
2. Support input mnemonic phrase when rooch init.
3. Record the mnemonic phrase in the keystore file.
4. Reuse the mnemonic phrase when rooch account create.
5. Remove `rooch account import` command. We do not support multi mnemonic phrase keystore now.
6. Remove `rooch account update` command. We do not support multi mnemonic phrase keystore now.
7. Encrypt the mnemonic phrase with keystore password.

PS.
1. The kestore format is incompatible. Need to delete the keystore file first `rm -rf ~/.rooch`

2. Supports specifying mnemonic words during initialization and reusing mnemonic words to create the same account, similar to Metamask creation of accounts.
```rooch init -m "{your mnemonic phrase}"```

### CODE REVIEW

**Comment by baichuan3** [comment]
There is a major refactor of the keystore, which has many conflicts with several of your previous PRs. It has been merged manually. @feliciss can help review

File: crates/rooch/src/commands/init.rs [hunk 1]
```diff
@@ -26,11 +28,17 @@ pub struct Init {
     /// Command line input of custom server URL
     #[clap(short = 's', long = "server-url")]
     pub server_url: Option<String>,
-    #[clap(flatten)]
-    pub context_options: WalletContextOptions,
+    /// Whether a non-empty password should be provided to rooch.keystore when it comes to the init command
+    #[clap(long = "encrypt-keystore")]
+    pub encrypt_keystore: Option<bool>,
```
jolestar, line 33: this option is repeat with 'skip-password'

crates/rooch-key/src/keystore/account_keystore.rs
```diff
@@ -0,0 +1,221 @@
...
+        password: Option<String>,
+    ) -> Result<Signature, anyhow::Error>
+    where
+        T: Serialize;
+
+    fn addresses(&self) -> Vec<RoochAddress>;
+
+    fn generate_and_add_new_key(
+        &mut self,
+        mnemonic_phrase: Option<String>,
```
jolestar, line 89: Do we need this argument? The mnemonic shoud be a required field when the keystore create.

baichuan3: needed when `rooch init` with `mnemonic phrase` option

jolestar: Maybe the init command should create a new keystore with mnemonic, we can refactor it later.

jolestar (approved):

baichuan3:
I merge this pr first

feliciss:
@baichuan3 @jolestar

jolestar:
One mnemonics is mapping to multi address, so, we can not use the address as the mnemonics's key.

But the addresses in MnemonicData, seem useless, we can remove it.

baichuan3:
> One mnemonics is mapping to multi address, so, we can not use the address as the mnemonics's key.
> 
> But the addresses in MnemonicData, seem useless, we can remove it.

Addresses in MnemonicData are used to restore accounts according to BIP44 and generate the address index.
According to the metamask account logic, the wallet supports generating multiple accounts by mnemonic words and importing by private keys, so keys and addresses can be inconsistent. I think it is necessary to record the addresses generated by mnemonic words in MnemonicData.

PR #954, opened 2023-10-11
[gh-933] move password hash into keystore.

### LINKED ISSUE #933: [CLI] Use one password for keystore

feliciss:
## Summary

Continuing of #942 

- Closes #933

### DIFF
--- a/crates/rooch-key/src/keystore.rs
+++ b/crates/rooch-key/src/keystore.rs
@@ -58,7 +58,7 @@ pub trait AccountKeystore {
         address: &RoochAddress,
         password: Option<String>,
     ) -> Result<Vec<RoochKeyPair>, anyhow::Error>;
-    fn get_key_pair_by_password(
+    fn get_key_pair_with_password(
         &self,
         address: &RoochAddress,
         password: Option<String>,
@@ -254,7 +254,7 @@ impl AccountKeystore for Keystore {
         }
     }
 
-    fn get_key_pair_by_password(
+    fn get_key_pair_with_password(
         &self,
         address: &RoochAddress,
         password: Option<String>,
@@ -262,10 +262,10 @@ impl AccountKeystore for Keystore {
         // Implement this method to get the key pair by coin ID from the appropriate variant (File or InMem)
         match self {
             Keystore::File(file_keystore) => {
-                file_keystore.get_key_pair_by_password(address, password)
+                file_keystore.get_key_pair_with_password(address, password)
             }
             Keystore::InMem(inmem_keystore) => {
-                inmem_keystore.get_key_pair_by_password(address, password)
+                inmem_keystore.get_key_pair_with_password(address, password)
             }
         }
     }
@@ -428,7 +428,7 @@ impl BaseKeyStore {
 }
 
 impl AccountKeystore for BaseKeyStore {
-    fn get_key_pair_by_password(
+    fn get_key_pair_with_password(
         &self,
         address: &RoochAddress,
         password: Option<String>,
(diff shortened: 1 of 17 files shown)


crates/rooch/src/commands/init.rs
```diff
@@ -27,9 +27,9 @@ pub struct Init {
     pub server_url: Option<String>,
     #[clap(flatten)]
     pub context_options: WalletContextOptions,
-    /// Whether a non-empty password should be provided to rooch.keystore when it comes to the init command
-    #[clap(long = "encrypt-keystore")]
-    pub encrypt_keystore: Option<bool>,
+    /// This is a tweak command flag to skip the integration test when executing cmd.feature file
+    #[clap(long = "skip-integration-test")]
+    skip_integration_test: bool,
```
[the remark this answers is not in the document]
feliciss: The `skip-password` is the same as pressing enter key with the command `rooch init`. However, we can rename it.

crates/rooch-key/src/keystore.rs
```diff
@@ -712,6 +769,27 @@ impl AccountKeystore for FileBasedKeystore {
...
+    fn set_password_hash_with_indicator(
+        &mut self,
+        password_hash: String,
+        is_password_empty: bool,
+    ) -> Result<(), anyhow::Error> {
+        self.keystore.password_hash = Some(password_hash);
+        self.keystore.is_password_empty = is_password_empty;
+        if std::env::var_os("TEST_ENV").is_none() {
+            self.save()?;
+        }
```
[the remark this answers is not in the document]
feliciss: Okay. I remove it anyway in this PR.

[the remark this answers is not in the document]
baichuan3: I have already removed `TEST_ENV` in another pr https://github.com/rooch-network/rooch/pull/950, so just ignored it

crates/rooch/src/commands/init.rs
```diff
@@ -27,9 +28,9 @@ pub struct Init {
     pub server_url: Option<String>,
     #[clap(flatten)]
     pub context_options: WalletContextOptions,
-    /// Whether a non-empty password should be provided to rooch.keystore when it comes to the init command
-    #[clap(long = "encrypt-keystore")]
-    pub encrypt_keystore: Option<bool>,
+    /// This is a tweak command flag to skip the integration test when executing cmd.feature file or use with rooch init to ignore entering password
```
jolestar, line 31: The command help also needs to improve. The user does not know the `cmd.feature` things.

jolestar (approved):

crates/rooch/src/commands/init.rs
jolestar: @feliciss I will merge this PR first. You can improve it in the new PR.

jolestar: > > The command help also needs to improve. The user does not know the `cmd.feature` things.
> 
> This comment is not only for the users, but the developers. Thus, I don't think it's entirely for users. The users may also include developers.

Yes, the major users are the DApp developer on rooch, but not rooch's developer. DApp developers do not need to know what `cmd.feature` things.  If the comment is for rooch's developer, it can be the code comment, but not the command help.

feliciss: > > > The command help also needs to improve. The user does not know the `cmd.feature` things.
> > 
> > 
> > This comment is not only for the users, but the developers. Thus, I don't think it's entirely for users. The users may also include developers.
> 
> Yes, the major users are the DApp developer on rooch, but not rooch's developer. DApp developers do not need to know what `cmd.feature` things. If the comment is for rooch's developer, it can be the code comment, but not the command help.

Got it.

PR #965, opened 2023-10-15
[RPC] Refactor rpc list view

jolestar:
## Summary

1. Refactor ListView from Vec<Option<(Vec<u8>, State)>> to Vec<(Vec<u8>, State)>.
2. Remove `List` from `PageView`.

### DIFF
--- a/crates/rooch-rpc-server/src/server/rooch_server.rs
+++ b/crates/rooch-rpc-server/src/server/rooch_server.rs
@@ -12,9 +12,9 @@ use rooch_rpc_api::api::{MAX_RESULT_LIMIT, MAX_RESULT_LIMIT_USIZE};
 use rooch_rpc_api::jsonrpc_types::account_view::BalanceInfoView;
 use rooch_rpc_api::jsonrpc_types::transaction_view::TransactionWithInfoView;
 use rooch_rpc_api::jsonrpc_types::{
-    AccessPathView, AccountAddressView, AnnotatedEventView, AnnotatedStateView, EventPageView,
-    ExecuteTransactionResponseView, FunctionCallView, H256View, ListAnnotatedStatesPageView,
-    ListBalanceInfoPageView, ListStatesPageView, StateView, StrView, StructTagView,
+    AccessPathView, AccountAddressView, AnnotatedEventView, AnnotatedStateView,
+    AnnotatedStatesPageView, BalanceInfoPageView, EventPageView, ExecuteTransactionResponseView,
+    FunctionCallView, H256View, StateView, StatesPageView, StrView, StructTagView,
     TransactionWithInfoPageView,
 };
 use rooch_rpc_api::{api::rooch_api::RoochAPIServer, api::DEFAULT_RESULT_LIMIT};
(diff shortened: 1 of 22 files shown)


crates/rooch-rpc-server/src/server/rooch_server.rs
```diff
@@ -111,31 +111,28 @@ impl RoochAPIServer for RoochServer {
         access_path: AccessPathView,
```
baichuan3, line 111: Since the list needs to be removed, consider renaming the fucntion name `list_states` into a unified prefix or suffix, such as `get_states_page` or `query_states`, or is there a better name?
The `list` in other modules are the same as above.

crates/rooch-rpc-server/src/service/rpc_service.rs
```diff
@@ -113,7 +113,7 @@ impl RpcService {
         access_path: AccessPath,
         cursor: Option<Vec<u8>>,
         limit: usize,
-    ) -> Result<Vec<Option<(Vec<u8>, State)>>> {
+    ) -> Result<Vec<(Vec<u8>, State)>> {
```
baichuan3, line 116: Is it that the format vec returned by all list methods no longer contains option, or that the vct returned by all rpc methods no longer contains option?
That is, if the rpc result is a vec structure, the format is vec<T> or vec<option <T>>, I feel that it needs to be unified. The underlying database returns vec<option<T>> by default.

crates/rooch-rpc-server/src/server/rooch_server.rs
jolestar: I think the function starting with `list` is ok. the result with `List` is repeat with `Page`

crates/rooch-rpc-server/src/service/rpc_service.rs
jolestar: I have change the `smt::list` to return `Vec<(K,V)>`

baichuan3: Not only the states API, but also other APIs, such as the transaction API, the return value is Vec<Option<TransactionWithInfoView>>

if the rpc result is a vec structure, the format is vec\<T\> or vec<option\<T\>>, need unified?

jolestar: It needs to be handled case by case.

Such as `GetTxExecutionInfosByHashMessage`, if the ExecutionInfo of `tx_hash` does not exist, it should be Option. We need to ensure the TxHash length equals the result Vec length.

baichuan3: `tx_hash` is valid but ExecutionInfo does not exist. This situation should not exist, right? The current processing will directly throw an error `Invalid tx hash or tx order`

baichuan3 (approved):

crates/rooch-rpc-server/src/service/rpc_service.rs
jolestar: Raising an error should also be a solution. if we raise error, the result should be Vec<T>, not Vec<Option<T>>

jolestar: I record an issue #966

PR #971, opened 2023-10-15
Auto sort modules by dependency order and review native function errors

### LINKED ISSUE #928: [Framework] Check module dependency order when publish
When publishing modules through `moveos_stdlib::account_storage::publish_modules_entry(ctx: &mut Context, account: &signer, modules: vector<vector<u8>>)`, the input parameter `modules` may not be sorted in dependency order which will cause `MiscellaneousError` for developers.  A previous issue #775.

We can optimize developer experience through the following options:

- Option 1: Check the order when verify modules, abort with meaning code if modules not sorted by dependency order.
- Option 2: Auto sort modules by dependency order when verify modules.

### LINKED ISSUE #960: [NativeFunction] Native Function error code
Some native functions directly raise the error code.

https://github.com/rooch-network/rooch/blob/4a649e19a2040d8e3c9378f0701bf8217bf84fee/crates/rooch-framework/src/natives/rooch_framework/crypto/ecdsa_k1_recoverable.rs#L50-L52

This needs to be converted by the `moveos_types::move_std::error` function, like

https://github.com/rooch-network/rooch/blob/4a649e19a2040d8e3c9378f0701bf8217bf84fee/moveos/moveos-stdlib/src/natives/moveos_stdlib/raw_table/mod.rs#L482-L485

pause125:
## Summary

1. Auto sort modules by dependency order. Resolve #928 
2. Review native functions and unify error code. Resolve #960


crates/rooch-framework/src/natives/rooch_framework/crypto/encoding.rs
```diff
@@ -18,10 +18,10 @@ use move_vm_types::{
...
 use std::collections::VecDeque;
 
-pub const INVALID_PUBKEY: u64 = 1;
-pub const EXCESSIVE_SCRIPT_SIZE: u64 = 2;
-pub const INVALID_DATA: u64 = 3;
-pub const INVALID_SCRIPT_VERSION: u64 = 4;
+pub const E_INVALID_PUBKEY: u64 = 1;
+pub const E_EXCESSIVE_SCRIPT_SIZE: u64 = 2;
+pub const E_INVALID_DATA: u64 = 3;
+pub const E_INVALID_SCRIPT_VERSION: u64 = 4;
```
jolestar, line 24: The error code should also be defined in the Move module, and then the `move explain` command can explain the error code.

moveos/moveos-stdlib/moveos-stdlib/sources/account_storage.move
```diff
@@ -170,18 +170,31 @@ module moveos_std::account_storage {
...
+        let i = 0;
+        let len = vector::length(modules);
+        while (i < len) {
+            let m = vector::borrow(modules, i);
+            if (move_module::module_name(m) == name) {
+                return vector::remove(modules, i)
+            };
+            i = i + 1;
+        };
+        abort(0x0) // unreachable.
```
jolestar, line 183: if the module of the `name` does not exist, will abort 0x0?

moveos/moveos-stdlib/moveos-stdlib/sources/raw_table.move
```diff
@@ -19,8 +19,10 @@ module moveos_std::raw_table {
     const ErrorAlreadyExists: u64 = 1;
     /// Can not found the key in the table
     const ErrorNotFound: u64 = 2;
+    /// Duplicate operation on the table
+    const ErrorDuplicateOperation: u64 = 3;
```
jolestar, line 23: These errors are different from the raw_table native implemention.

jolestar: It is better to define it and make sure the error code in Rust is consistent with error code in Move.

jolestar (approved):

PR #973, opened 2023-10-15
 Refactor rpc result view 

wow-sven:
## Summary

Enhanced maintainability: With clear differentiation between parameters and return values, future modifications and debugging will be more straightforward.
Consistent coding style: Applying this convention across the codebase ensures a uniform and standardized approach to RPC function design.

### DIFF
--- a/crates/rooch-rpc-api/src/jsonrpc_types/execute_tx_result.rs
+++ b/crates/rooch-rpc-api/src/jsonrpc_types/execute_tx_result.rs
@@ -159,21 +159,21 @@ impl From<TransactionOutput> for TransactionOutputView {
 }
 
 #[derive(Debug, Clone)]
-pub struct ExecuteTransactionResponse {
+pub struct ExecuteTransactionResult {
     pub sequence_info: TransactionSequenceInfo,
     pub execution_info: TransactionExecutionInfo,
     pub output: TransactionOutput,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
-pub struct ExecuteTransactionResponseView {
+pub struct ExecuteTransactionViewResult {
     pub sequence_info: TransactionSequenceInfoView,
     pub execution_info: TransactionExecutionInfoView,
     pub output: TransactionOutputView,
 }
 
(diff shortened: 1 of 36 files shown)


crates/rooch-rpc-api/src/jsonrpc_types/move_types.rs
```diff
@@ -384,6 +384,26 @@ impl From<H256View> for H256 {
     }
 }
 
+#[serde_as]
+#[derive(Default, Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, JsonSchema)]
+pub struct H256ViewResult(
```
jolestar, line 389: H256View is a String. I think we should replace it with StrView<H256>. We do not need to distinguish H256 argument and H256Result.

crates/rooch-rpc-api/src/jsonrpc_types/rooch_types.rs
```diff
@@ -50,7 +50,7 @@ impl From<TransactionType> for TransactionTypeView {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
-pub struct TransactionView {
+pub struct TransactionInfoView {
```
jolestar, line 53: This is not TransactionInfo. This is the Transaction self.

crates/rooch-rpc-api/src/jsonrpc_types/execute_tx_result.rs
```diff
@@ -159,21 +159,21 @@ impl From<TransactionOutput> for TransactionOutputView {
...
-pub struct ExecuteTransactionResponse {
+pub struct ExecuteTransactionResult {
     pub sequence_info: TransactionSequenceInfo,
     pub execution_info: TransactionExecutionInfo,
     pub output: TransactionOutput,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
-pub struct ExecuteTransactionResponseView {
+pub struct ExecuteTransactionViewResult {
```
jolestar, line 169: Should be ExecuteTransactionResultView?

jolestar (changes requested):

PR #979, opened 2023-10-16
[RPC] Unify view types with StrView

### LINKED ISSUE #976: [Review] Use StrView<T> to replace H256View, H64View. etc
Part of #966 

If a struct should display as a string in json, it should use StrView<T>, not defining a new view struct.

jolestar:
## Summary

1. Define H256View = StrView<H256>, H160View = StrView<H160>, H64 = StrView<H64>, IdentifierView = StrView<Identifier>
2. Define BytesView = StrView<Vec<u8>>, and remove Bytes.
3. Change u64 and u128 in RPC method arguments to StrView<u64>, StrView<u128>
4. Remove `net_version` API and add `eth` namespace.
5. Add StrView test cases.

- Closes #976


baichuan3 (approved):

baichuan3:
Encounter an bad case after change u64 and u128 in RPC method arguments to StrView, StrView.

Either pass the integer type directly, or support both integer and string types?

`
curl --location 'http://localhost:50051' \
--header 'Content-Type: application/json' \
--data '{
    "id": 101,
    "jsonrpc": "2.0",
    "method": "rooch_getEventsByEventHandle",
    "params": [
        "0x4b1b5d2a62f3c9a19a5314167439945bf98fbc871e1402e36d5bad3382064c56::event_test::WithdrawEvent", null, 5
    ]
}'
```


```
{
    "jsonrpc": "2.0",
    "error": {
        "code": -32602,
        "message": "invalid type: integer `5`, expected a string at line 1 column 2"
    },
    "id": 101
}
```

jolestar:
Need to change 5 to "5"

Issue #970, opened 2023-10-15
[Review] Multi key query RPC should return Vec<Option<V>> or Vec<T>


jolestar:
Part of #966 

Multi key query RPC should return `Vec<Option<V>>` or `Vec<T>`

Such as `get_transactions_by_hash(Vec<Hash>)`, if one transaction of the hash does not exist, we have two solutions:

1. Raise an error, we should return `Vec<V>`.
2. Return Option, we should return `Vec<Option<T>>`

I prefer 2, because sometimes, the client is not sure if the tx hash is valid, it needs to query the API to check if the tx hash is valid.

If we raise an error for the first invalid tx hash, it can not check the tx hash batch.

baichuan3:
If we expect to check if the tx hash is valid at the same time, option 2 is better indeed.

baichuan3:
If we expect to check if the tx hash is valid at the same time, option 2 is better indeed.

jolestar:
Option 2

baichuan3:
```
curl --location 'http://localhost:50051' \
--header 'Content-Type: application/json' \
--data '{
 "id":101, 
 "jsonrpc":"2.0", 
 "method":"rooch_getTransactionsByHash", 
 "params":[["0x60185a08d4b77296a4103707676f5d8ccf5584f524cbdd6aa61bcfba5a7385ec","0x5456f5d14c356762661ca8933dcf6098e61fa408767f495b49338d0858f48dfd"]]
}'
```

```
{
    "jsonrpc": "2.0",
    "result": [
        null,
        {
            "transaction": {
                "transaction_type": "rooch",
                "sequence_number": 1,
                "sender": "0x4b1b5d2a62f3c9a19a5314167439945bf98fbc871e1402e36d5bad3382064c56",
                "action_type": "functioncall",
                "action": {
                    "function_call": {
                        "function_id": "4b1b5d2a62f3c9a19a5314167439945bf98fbc871e1402e36d5bad3382064c56::fixed_supply_coin::faucet",
                        "ty_args": [],
                        "args": []
                    }
                },
[...]

PR #980, opened 2023-10-16
Multi key query RPC should return Vec<Option<V>>

baichuan3:
resolve https://github.com/rooch-network/rooch/issues/970

If the tx hash is invalid, filled None when returned.

### DIFF
--- a/crates/rooch-rpc-server/src/service/aggregate_service.rs
+++ b/crates/rooch-rpc-server/src/service/aggregate_service.rs
@@ -204,7 +204,7 @@ impl AggregateService {
             .get_transaction_execution_infos_by_hash(tx_hashes.clone())
             .await?;
 
-        assert!(
+        debug_assert!(
             transactions.len() >= sequence_infos.len()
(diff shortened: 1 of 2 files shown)


crates/rooch-rpc-server/src/service/aggregate_service.rs
```diff
@@ -196,25 +205,43 @@ impl AggregateService {
             .await?;
 
         assert!(
-            transactions.len() == sequence_infos.len()
+            transactions.len() >= sequence_infos.len()
```
jolestar, line 208: The assert will panic the server. It should change to debug_assert, just for debug binary.

jolestar: Or raise an error?

jolestar (approved):

crates/rooch-rpc-server/src/service/aggregate_service.rs
baichuan3: change to debug_assert

PR #989, opened 2023-10-17
[Move Object] Public Object transfer, borrow, borrow_mut

### LINKED ISSUE #80: [Object] Proposal for an Owner method to directly operate on Objects
Currently, Rooch provides methods in the ObjectStorage module such as borrow, borrow_mut, remove, and add to operate on Objects via Modules containing the Objects' generic types. These methods are ensured to be called via `private_generics` for type safety. But sometimes, an Object needs can to be directly operated on by its Owner.

A possible solution to this issue is to provide a method similar to Sui's `public_transfer` that allows Owners to operate on Objects directly.

```move

module moveos_std::object_storage{
    #[private_generics(T)]
    /// Borrow Object from object store with object_id
    public fun borrow<T: key>(this: &ObjectStorage, object_id: ObjectID): &Object<T>;

    public fun public_borrow<T: key + store>(this: &ObjectStorage, object_id: ObjectID): &Object<T>;

    #[private_generics(T)]
    /// Borrow mut Object from object store with object_id
    public fun borrow_mut<T: key>(this: &mut ObjectStorage, object_id: ObjectID): &mut Object<T>;

    public fun public_borrow_mut<T: key+store>(this: &mut ObjectStorage, object_id: ObjectID): &mut Object<T>;

    #[private_generics(T)]
    /// Remove object from object store
    public fun remove<T: key>(this: &mut ObjectStorage, object_id: ObjectID): Object<T>;
   
    public fun public_remove<T: key+store>(this: &mut ObjectStorage, object_id: ObjectID): Object<T>;

    #[private_generics(T)]
    /// Add object to object store
    public fun add<T: key>(this: &mut ObjectStorage, obj: Object<T>);
    
    public fun public_add<T: key+store>(this: &mut ObjectStorage, obj: Object<T>);

}

```
 
Overall, we propose further discussion and research on this topic to determine the need for such a method, its impact on security, and possible implementation details.

WGB5445:
## Summary

Object should be freely borrowable or transferable
Therefore, the borrow and transfer of Object with Key + store are added

Public_add and public_remove have not been added for the time being because they may destabilize the life cycle of Object, so these two functions will not be added for the time being.

- Closes #80

### DIFF
--- a/moveos/moveos-stdlib/moveos-stdlib/sources/object.move
+++ b/moveos/moveos-stdlib/moveos-stdlib/sources/object.move
@@ -50,6 +50,10 @@ module moveos_std::object {
         &self.value
     }
 
+    public fun public_borrow<T: key + store>(self: &Object<T>): &T {
+        &self.value
+    }
+
     public(friend) fun internal_borrow<T>(self: &Object<T>): &T {
         &self.value
     }
@@ -60,6 +64,10 @@ module moveos_std::object {
         &mut self.value
     }
 
+    public fun public_borrow_mut<T: key + store>(self: &mut Object<T>): &mut T {
+        &mut self.value
+    }
+
     public(friend) fun internal_borrow_mut<T>(self: &mut Object<T>): &mut T {
         &mut self.value
     }
@@ -67,6 +75,10 @@ module moveos_std::object {
     #[private_generics(T)]
     /// Transfer object to recipient
     public fun transfer<T: key>(self: &mut Object<T>, recipient: address) {
+        self.owner = recipient;
+    }
+
+    public fun public_transfer<T: key + store>(self: &mut Object<T>, recipient: address) {
         self.owner = recipient;
     }
 
--- a/moveos/moveos-stdlib/moveos-stdlib/doc/object.md
+++ b/moveos/moveos-stdlib/moveos-stdlib/doc/object.md
@@ -15,10 +15,13 @@ The differents with the Object in [Sui](https://github.com/MystenLabs/sui/blob/5
 -  [Function `address_to_object_id`](#0x2_object_address_to_object_id)
 -  [Function `new`](#0x2_object_new)
 -  [Function `borrow`](#0x2_object_borrow)
+-  [Function `public_borrow`](#0x2_object_public_borrow)
 -  [Function `internal_borrow`](#0x2_object_internal_borrow)
 -  [Function `borrow_mut`](#0x2_object_borrow_mut)
+-  [Function `public_borrow_mut`](#0x2_object_public_borrow_mut)
 -  [Function `internal_borrow_mut`](#0x2_object_internal_borrow_mut)
 -  [Function `transfer`](#0x2_object_transfer)
+-  [Function `public_transfer`](#0x2_object_public_transfer)
 -  [Function `id`](#0x2_object_id)
 -  [Function `owner`](#0x2_object_owner)
 -  [Function `unpack`](#0x2_object_unpack)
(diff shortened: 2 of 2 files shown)


moveos/moveos-stdlib/moveos-stdlib/sources/object.move
```diff
@@ -50,6 +50,10 @@ module moveos_std::object {
         &self.value
     }
 
+    public fun public_borrow<T: key + store>(self: &Object<T>): &T {
```
jolestar, line 53: How about we remove the `private_generics` from `borrow` and `borrow_mut`?  Because `Context::borrow_object` and `Context::borrow_object_mut` has the `private_generics`.

@stevenlaw123 @pause125 @baichuan3

moveos/moveos-stdlib/moveos-stdlib/sources/object.move
```diff
@@ -70,6 +78,10 @@ module moveos_std::object {
         self.owner = recipient;
     }
 
+    public fun public_transfer<T: key + store>(self: &mut Object<T>, recipient: address) {
```
jolestar, line 81: Make the `transfer` to `key + store` and remove `#[private_generics(T)]`, `public_transfer` to `transfer_extend` ? Keep consistency with `coin` and `account_coin_store`

moveos/moveos-stdlib/moveos-stdlib/sources/object.move
jolestar: And should we support borrowing Object<T: key+store> from Context without `private_generics` ?

WGB5445: Wait a minute, I feel there is a problem here. If we let go of Borrow, this means that T can be obtained by any module at will, which may cause security problems.

So I think it’s better to delete them and add them if necessary.

moveos/moveos-stdlib/moveos-stdlib/sources/object.move
baichuan3: Does `xxx_extend` always have #[private_generics(T)] check? Is it more appropriate to put the object `public_transfer` function in the transfer module?

moveos/moveos-stdlib/moveos-stdlib/sources/object.move
WGB5445: > And should we support borrowing Object<T: key+store> from Context without `private_generics` ?

Maybe this shouldn't be the case, it would mean that other modules could borrow arbitrary Object<T>
This may cause security issues

WGB5445: > How about we remove the `private_generics` from `borrow` and `borrow_mut`? Because `Context::borrow_object` and `Context::borrow_object_mut` has the `private_generics`.
> 
> @stevenlaw123 @pause125 @baichuan3
I think it's reasonable
Because there is no way to get `&Object<T>` or `Object<T>` except context and current module.

baichuan3: > borrowing Object<T: key+store>

The reference of Object can only be obtained through context. I think it can be removed and there is no security issue.

baichuan3: In my understanding, borrow Object<T: key> and borrow Object<T: key + store> satisfy two types of scenarios respectively?
1. If Object<T: key> is only a key capability, it can only be accessed through context::borrow_object (with private_generics(T) restriction) or object:borrow (need to pass Object<T> reference);
2. If Object<T: key+store> also has store capability, in addition to option 1, it can also be called through context::public_borrow_object* (there is no private_generics(T) restriction, and there is no need to pass the Object<T> reference).

moveos/moveos-stdlib/moveos-stdlib/sources/object.move
jolestar: The transfer function in the transfer module needs to use object_id as the argument.

```move
public entry transfer_object<T:key+store>(ctx, object_id, to){
  let object = context::borrow_object_mut<T>(ctx, object_id);
  object::transfer(object, to);
}
```

So, it depends on how to support the public borrow object from context.

PR #1023, opened 2023-10-22
SDK  0.0.2 preview

wow-sven:
## Summary

1. Uniform provider derived name
2. Readme Add use doc

- Closes #issue

### DIFF
--- a/sdk/typescript/README.md
+++ b/sdk/typescript/README.md
@@ -5,18 +5,21 @@ This is the Rooch TypeScript SDK built on the Rooch [JSON RPC API](https://githu
 WARNING: Note that we are still iterating on the RPC and SDK API before TestNet, therefore please expect frequent breaking changes in the short-term. We expect the API to stabilize after the upcoming TestNet launch.
 
 ## Installation
-``` shell
+
+```shell
 npm i @roochnetwork/sdk
 ```
+
 ## Connecting to Rooch Network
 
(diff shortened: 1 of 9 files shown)


sdk/typescript/README.md
```diff
@@ -4,6 +4,154 @@ This is the Rooch TypeScript SDK built on the Rooch [JSON RPC API](https://githu
...
+
+### Get States
+
+Refer to [this storage guide](https://rooch.network/zh-CN/docs/dive-into-rooch/storage-abstraction) for more information.
+
+```typescript
+import { JsonRpcProvider, DevChain } from '@rooch/sdk'
+const provider = new JsonRpcProvider(DevChain)
+
+const states = provider.getStates('object/0x1', null, 10)
```
jolestar, line 152: Here is `getStates`

sdk/typescript/src/account/account.ts
```diff
@@ -197,13 +197,13 @@ export class Account implements IAccount {
 
   async querySessionKeys(cursor: Bytes | null, limit: number): Promise<IPage<ISessionKey>> {
     const accessPath = `/resource/${this.address}/0x3::session_key::SessionKeys`
-    const state = await this.provider.getStates(accessPath)
+    const state = await this.provider.getState(accessPath)
```
jolestar, line 200: But `getState` here. We should unify them. I suggest use `getStates`, keep consistent with RPC

jolestar (approved):

Issue #1002, opened 2023-10-19
[MoveosStd] Bytecode module template


jolestar:
# Motivation

In Move, there are two challenges to implementing the dynamic deployment of a contract to an account, similar to the `create2` effect in EVM:

1. The address in the Move module is determined at compile time, so it is not possible to deploy a compiled bytecode directly to another account. This will resolved via #781. 
2. Dynamic deployment may need to replace the module name or struct name. For example, to dynamically deploy a contract that issues `Coin<T>`, the type of `T` needs to be defined in the module, and the struct name of `T `needs to be dynamically replaced. So, we designed the current solution.

## Solution

1. We define a template module, such as Coin issue module template.

```move
module my_coin::__COIN_MODULE_NAME_PLACEHOLDER__{
   
   struct __COIN_STRUCT_NAME_PLACEHOLDE__ has key,store{
   }

   fun init(ctx: &mut Context){
      coin::register_extend<>(ctx,
            string::utf8(b"__COIN_NAME_PLACEHOLDER__"),
            string::utf8(b"__COIN_SYMBOL_PLACEHOLDER__"),
            18)
   }
}
```

2. We compiled the module template and got the bytecode.
3. Replace the place holder in bytecode when deploying the module.

Part of #626

PR #1029, opened 2023-10-24
Remapping address when publish modules

### LINKED ISSUE #781: [MoveVM] Binding module address when module publishing
Binding module address when module publishing.

Suppose we want to deploy a prepared bytecode module, like a CoinType module, to a user or resource account within the Move contract. In that case, it is not feasible because the address in the module does not match the user's address.

So, we need to find a solution to change the module address when publishing the module.

1. Remapping module address to another address, like https://github.com/aptos-labs/aptos-core/issues/8379
2. Use a fixed placeholder address, like `0xB1ACE40DE2`; if the module address is the placeholder address, do not check the module address with the current sender's address; directly replace the module address with the sender's address.

pause125:
## Summary

Remapping module address to another address in Move code.

- Closes #781


moveos/moveos-stdlib/src/natives/moveos_stdlib/move_module.rs
```diff
@@ -265,6 +268,123 @@ fn check_compatibililty_inner(
...
+    address_mapping: &HashMap<AccountAddress, AccountAddress>,
+) -> PartialVMResult<()> {
+    // replace addresses in address identifiers.
+    for addr in module.address_identifiers.iter_mut() {
+        if let Some(new_addr) = address_mapping.get(addr) {
+            *addr = *new_addr;
+        }
+    }
+    // replace addresses in constant.
+    for constant in module.constant_pool.iter_mut() {
```
jolestar, line 367: Should the address in constant_pool be replaced? I think remap only changes the module address and should not change the constant.

jolestar: Or we provide low-level API to the developer, such as `replace_constant` and `replace_address_identifier`, `replace_identifier`, let the developer decide based on the use-case. And `replace_identifier` also can resolve the #1002

jolestar (approved):
I will merge this PR first.

PR #1042, opened 2023-10-26
Support replacing identifier, address and string constant in module bytecodes

### LINKED ISSUE #1002: [MoveosStd] Bytecode module template

pause125:
## Summary

Support replacing identifier, address and string constant in module bytecodes.

Now we can replace the module address, module name, struct name and string constant of the given module bytecodes in Move contracts. So devs can dynamic deployment contracts.

- Refactor implementation of #781 
- Resolve #1002


moveos/moveos-stdlib/moveos-stdlib/sources/move_module.move
```diff
@@ -55,15 +55,45 @@ module moveos_std::move_module {
...
+                byte_codes: vector::pop_back(&mut rebinded_bytes),
+            });
+            i = i + 1;
+        };
+        vector::destroy_empty(rebinded_bytes);
+        rebinded_modules
+    }
+
+    /// Replace given module's name to the new name
+    public fun replace_module_name(
```
jolestar, line 90: The name may be some other identifier, so should we rename this function to `replace_module_identiner` to avoid misuse?

moveos/moveos-stdlib/src/natives/moveos_stdlib/move_module.rs
```diff
@@ -270,22 +260,80 @@ fn check_compatibililty_inner(
...
+ **************************************************************************************************/
+#[derive(Debug, Clone)]
+pub struct ReplaceAddressConstantGasParameters {
     pub base: InternalGas,
     pub per_byte: InternalGasPerByte,
 }
 
-fn remap_module_addresses_inner(
-    gas_params: &RemapAddressesGasParameters,
+fn replace_addresses_constant(
```
baichuan3, line 335: What is the different between `replace_addresses_constant ` and `replace_address_identifiers `?

moveos/moveos-stdlib/moveos-stdlib/sources/move_module.move
```diff
@@ -101,14 +140,36 @@ module moveos_std::move_module {
...
+    /// Native function to replace constant addresses in module binary where the length of
+    /// `old_addresses` must equal to that of `new_addresses`.    
+    native public(friend) fun replace_addresses_constant(
         bytes: vector<vector<u8>>,
         old_addresses: vector<address>,
         new_addresses: vector<address>,
     ): vector<vector<u8>>;
 
+    /// Native function to replace constant bytes in module binary where the length of
+    /// `old_addresses` must equal to that of `new_addresses`.
```
baichuan3, line 167: typo in comment : `old_addresses` -> `old_bytes`, `new_addresses` -> `new_bytes`

jolestar (approved):

moveos/moveos-stdlib/src/natives/moveos_stdlib/move_module.rs
baichuan3: The native implementations of `replace_addresses_constant` and `replace_address_identifiers` are almost the same. Perhaps abstraction and code reuse can be considered in the next PR.

baichuan3 (approved):

PR #886, opened 2023-09-28
Check the Gas Free function during the compilation of the Move module.

stevenlaw123:
This is part of https://github.com/rooch-network/rooch/issues/839

Check the Gas Free function during the compilation of the Move module.

TODO:
1. Verify the correctness of the Gas Free function information in the module's metadata during module publishing.
2. Deduct gas fees by reading and executing the Gas Fee function information from the metadata.

PR #1068, opened 2023-10-29
Changing the execution stage of the gas_validate_function.

stevenlaw123:
Fix for the PR https://github.com/rooch-network/rooch/pull/886.


moveos/moveos/src/moveos.rs
```diff
@@ -243,12 +268,75 @@ impl MoveOS {
         }
     }
 
-    fn execute_gas_validate(
-        &self,
-        session: &mut MoveOSSession<'_, '_, MoveOSResolverProxy<MoveOSStore>, MoveOSGasMeter>,
-        action: &VerifiedMoveAction,
-    ) -> VMResult<Option<bool>> {
-        match action {
+    pub fn get_address_balance(&self, tx: &MoveOSTransaction) -> VMResult<u128> {
```
jolestar, line 271: MoveOs do not know the balance and do not know the `0x3::gas_coin`, so we should migrate this logic to the executor.

moveos/moveos/src/moveos.rs
```diff
@@ -243,12 +268,75 @@ impl MoveOS {
...
+                .expect("the return value of gas validate function should be u128");
+
+                Ok(balance.unchecked_as_u128())
+            }
+            _ => Ok(0),
+        }
+    }
+
+    pub fn execute_gas_validate(&self, tx: &MoveOSTransaction) -> VMResult<Option<bool>> {
+        let MoveOSTransaction {
```
jolestar, line 322: Same as above, we can directly call the validate function like the `transaction_validator::validate` in the executor. https://github.com/rooch-network/rooch/blob/6265cd36d2f19e6c42955d53f115dd2dd5af1c73/crates/rooch-types/src/framework/transaction_validator.rs#L29-L57

moveos/moveos/src/moveos.rs
stevenlaw123: OK, it's fixed.

moveos/moveos/src/moveos.rs
stevenlaw123: OK, it's fixed.

jolestar (approved):
This implementation is still not quite as refined as it could be, so I'll merge it first and refactor it myself.

Issue #1066, opened 2023-10-29
[Framework] Refactor RoochFramework with Object style 


jolestar:
- [x] #1067
- [x] ChainID to Singleton Object
- [x] Timestamp to Singleton Object
- [x] AddressMapping to Singleton Object
- [x] CoinInfo to Object or make it to CoinInfo<T> and become Singleton object.

jolestar:
#1138

Issue #1056, opened 2023-10-28
[MoveosStd] Simplify the Event API, remove the requirement for argument `&mut Context`


jolestar:
## Motivation

Our recommended development pattern is that developers borrow the required data from the `Context` at the outer tier and then call the base API to assemble it. This way the Object model and the AccountStorage model can reuse code.

However, this pattern encounters a problem: if you borrow the data at the outer tier, there is no way to pass the `&mut Context` and `&T` or `&Object<T>` together because the life cycle of `&T` or `&Object<T>` is bound to the `&mut Context`.

But `event::emit` requires the `&mut Context` argument, the developer needs to pass the `&mut Context` through the tiers.

So, I suggest to remove the `&mut Context` argument from the event API.

## Solution

1. Remove the EventHandle structure from the state, including the counter, so that `event::emit` is a function that does not change state. This will also help performance and parallel execution.
2. Store handle and counter information in Event Storage.
3. since `event::emit` does not change state, the view function can also emit events. Our view function calls no longer check the output of the event. But the event of the view function call is not recorded in the event storage.

PR #1077, opened 2023-10-30
[MoveosStd] Refactor singleton and display

jolestar:
## Summary

Part of #1066 

Refactor singleton and display.

1. Make singleton and display protected by `private_generics<T>`.
2. Make the display object default permanent and support Display<Object<T>>
3. Rename `context::borrow_object_mut` to `context::borrow_mut_object`.
4. Add more tests to the object.

### Refactor Display

Currently, Display has an issue:

We can create an `Object<Display<T>>` to display the `T`, But if we want to display the `Object<T>`, the display template can not get the `ObjectID` and `owner`.

So the Display object should be `Object<Display<Object<T>>>`, and the template can be:

```move
struct NFT has key,store {
        name: String,
        collection: ObjectID,
        creator: address,
}
```

```move
let display_obj: Object<Display<Object<T>>> = display::object_display<NFT>(ctx);
display::set_value(display_obj, string::utf8(b"url"), string::utf8(b"https://baseurl/{value.collection}/{id}")) 
```

So, I provide `display::resource_display<T>` for `T`, and `display::object_display<T>` for `Object<T>`.

- Closes #issue


moveos/moveos-stdlib/moveos-stdlib/sources/context.move
```diff
@@ -94,18 +94,44 @@ module moveos_std::context {
...
+    public fun borrow_singleton<T: key>(self: &Context): &Object<T> {
+        let object_id = object::singleton_object_id<T>();
+        borrow_object(self, object_id)
+    }
+
     /// Borrow mut Object from object store with object_id
-    /// If the object is not shared, only the owner can borrow an &mut Object from the global object storage
-    public fun borrow_object_mut<T: key>(_self: &mut Context, owner: &signer, object_id: ObjectID): &mut Object<T> {
+    /// If the object is not shared, only the owner can borrow an `&mut Object<T>` from the global object storage
+    public fun borrow_mut_object<T: key>(_self: &mut Context, owner: &signer, object_id: ObjectID): &mut Object<T> {
```
baichuan3, line 134: Since the `Context` reference is not used, can it be removed from the parameter?

jolestar: We require the `&mut Context` for ensure that if a transaction change the state, it must include `&mut Context` in the entry function's arguments.

moveos/moveos-stdlib/moveos-stdlib/sources/event.move
```diff
@@ -47,7 +47,7 @@ module moveos_std::event {
     /// Borrow a mut event handle from the object storage
     fun borrow_event_handle_mut<T>(ctx: &mut Context): &mut EventHandle {
         let event_handle_id = derive_event_handle_id<T>();
```
baichuan3, line 49: `event_handle_id` can also be refactored to use singleton_object directly

moveos/moveos-stdlib/moveos-stdlib/sources/context.move
baichuan3, line 134: `borrow_xxx_mut` and `borrow_mut_xxx` can be unified in rooch framework

baichuan3: Is this just a convention, or is it mandatory (the verifier checks)?

moveos/moveos-stdlib/moveos-stdlib/sources/event.move
jolestar: I plan to remove the EventHandle from state. #1056

moveos/moveos-stdlib/moveos-stdlib/sources/context.move
jolestar: It ensure by verifier and vm. An `&mut T` only can borrow from `&mut Context` or `&mut Object<T>` entry function argument. If the entry function has not `&mut` argument, it is an readonly function. I record an issue #1078

jolestar: Only one `borrow_box_mut` in the raw_table is an private native function. We can refactor it in the future.

jolestar: @stevenlaw123

baichuan3 (approved):

PR #1083, opened 2023-10-31
[Object] Implement shared and frozen Object

### LINKED ISSUE #1075: [MoveosStd] Implement shared and frozen Object
Follow #1001 

1. Add a `flag` field to ObjectEntity
 
```
struct ObjectEntity<T> {
      // The object id
      id: ObjectID,
      // The owner of the object
      owner: address,
      /// A flag to indicate whether the object is shared or frozen
      flag: u8,
      // The value of the object
      // The value must be the last field
      value: T,
}
```
2. Anyone can get `&mut Object<T>` from shared objects.
3. None one can get `&mut Object<T>` from frozen objects.

jolestar:
## Summary

Implement shared and frozen Object.

TODO: Add more tests and examples.

- Closes #1075

### DIFF
--- a/moveos/moveos-stdlib/moveos-stdlib/sources/object.move
+++ b/moveos/moveos-stdlib/moveos-stdlib/sources/object.move
@@ -31,6 +31,9 @@ module moveos_std::object {
 
     const SYSTEM_OWNER_ADDRESS: address = @0x0;
     
+    const SHARED_OBJECT_FLAG_MASK: u8 = 1;
+    const FROZEN_OBJECT_FLAG_MASK: u8 = 1 << 1;
+
     /// ObjectEntity<T> is a box of the value of T
     /// It does not have any ability, so it can not be `drop`, `copy`, or `store`, and can only be handled by storage API after creation.
     struct ObjectEntity<T> {
@@ -39,7 +42,7 @@ module moveos_std::object {
         // The owner of the object
         owner: address,
         /// A flag to indicate whether the object is shared or frozen
-        //flag: u8,
+        flag: u8,
         // The value of the object
         // The value must be the last field
         value: T,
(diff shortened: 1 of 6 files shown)


moveos/moveos-stdlib/moveos-stdlib/sources/object.move
```diff
@@ -125,19 +132,27 @@ module moveos_std::object {
...
     public fun to_shared<T: key>(self: Object<T>) {
         let obj_entity = borrow_mut_from_global<T>(self.id);
-        // TODO set the flag
-        transfer_to_system(obj_entity); 
+        to_shared_internal(obj_entity);
         to_permanent(self);
     }
 
+    fun to_shared_internal<T: key>(self: &mut ObjectEntity<T>) {
+        self.flag = self.flag | SHARED_OBJECT_FLAG_MASK;
```
baichuan3, line 140: Should be reset directly to `SHARED_OBJECT_FLAG_MASK` here ? 
Shared objects can be set to a frozen state? If yes,  there will be problems when setting frozen objects to be shared.
Or if the flag is expanded later, there may appear a bug here?

moveos/moveos-types/src/moveos_std/object.rs
```diff
@@ -221,12 +221,28 @@ pub type AccountStorageObject = ObjectEntity<AccountStorage>;
...
+        Self {
+            id,
+            owner,
+            flag,
+            value,
+        }
+    }
+
+    pub fn is_shared(&self) -> bool {
+        self.flag & Self::SHARED_OBJECT_FLAG_MASK == Self::SHARED_OBJECT_FLAG_MASK
```
baichuan3, line 241: the same above

moveos/moveos-stdlib/moveos-stdlib/sources/object.move
jolestar: Shared Object and frozen Object can not be at the same time, but I added a test to ensure the flag can be set at the same time.

baichuan3 (approved):

Issue #1037, opened 2023-10-25
[Indexer] Implement event indexer


baichuan3:
1. event indexer entry event_store/mod module.

2. Indexer's created_at and updated_at cannot use indexer time. They need to use the time on the chain and support replay.

3. Table Schema

- event

| Field	|  Type	|  Description
| -------- | ------- | ------- |
|event_handle_id |	varchar |	event handle id |
|event_seq |	int |	the number of messages that have been emitted to the path previously |
|type_tag |	varchar |	the type of the event data |
|event_data |	blob |	the data payload of the event |
|event_index |	int |	event index in the transaction events |
|created_at |	timestamp |	when the row was created |
|updated_at |	timestamp |	when the row was updated |
|tx_order |	bigint |	the sequencer order of the transaction |
|tx_hash |	varchar |	the transaction hash of the transaction |
|sender_address |	varchar |	user address who emit the event |

Primary key: (event_handle_id, event_seq) union pk
Index:

Issue #1035, opened 2023-10-25
[Indexer] Implement transaction indexer


baichuan3:
1. Transaction sender and other types are normalized. The sender of the transaction from Ethereum has Ethereum address and Rooch address. At least the Rooch address must be stored.

2. Transactions can be merged into a single table, and transaction indexer is triggered in RpcService::execute_tx.

3. Indexer's created_at and updated_at cannot use indexer local time. They need to use the time on the chain and support replay.

4. Table Schema

- transaction

| Field	|  Type	|  Description
| -------- | ------- | ------- |
|tx_hash |	varchar |	the hash of the transaction |
|transaction_type |	varchar |	the type of the transaction |
|chain_id |	int |	the chain id |
|auth_validator_id |	int |	the auth validator id of the authenticator info  |
|payload |	blob |	the authenticator info payload |
|encode_data |	blob |	the transaction encode data |
|created_at |	timestamp |	when the row was created |
|updated_at |	timestamp |	when the row was updated |
|sender_address |	varchar |	user address who emit the event |

Primary key: tx_hash pk
Index:

- transaction_sequence_info

| Field	|  Type	|  Description
| -------- | ------- | ------- |
|tx_order |	bigint |	the sequencer order of the transaction |
|auth_validator_id |	int |	the auth validator id of the tx order signature  |
|payload |	blob |	tx order signature payload |
|tx_accumulator_root |	varchar |	the tx accumulator root after the tx is append to the accumulator |
|created_at |	timestamp |	when the row was created |
|updated_at |	timestamp |	when the row was updated |
|sender_address |	varchar |	user address who emit the event |

Primary key: tx_order pk
Index:  
    
- transaction_execute_info

| Field	|  Type	|  Description
| -------- | ------- | ------- |
|tx_hash |	varchar |	the hash of the transaction |
|state_root |	varchar |	the root hash of Sparse Merkle Tree describing the world state at the end of this transaction. |
|event_root |	varchar |	the root hash of Merkle Accumulator storing all events emitted during this transaction |
|gas_used |	int |	the amount of gas used |
|status |	varchar |	the vm status |
|created_at |	timestamp |	when the row was created |
|updated_at |	timestamp |	when the row was updated |
|sender_address |	varchar |	user address who emit the event |

Primary key: tx_hash pk
Index:  

**TODO**
The final transaction table schema after be merged into a single table.

Issue #1028, opened 2023-10-24
[Indexer] Full Data & Real-time Indexer


baichuan3:
### Typical scenario
1. Which coins are registered in the system, the total amount of coin supply, and the number of currency holding addresses
2. Transaction list of a certain address; list of all coins and coin balance
3. Query event list by transaction

### Goal
1. Automatic indexing state and event data on Rooch
2. Provide API to query Index data
3. Provide SQL to query the Index data
4. Provide the ability to customize Indexer?

### Indexer solution
1. Checkpoint regularly and then write in batches
2. Listen the database file, generate Transaction Stream, and trigger writing
3. Automatically parse SMT nodes, parse `Type` types, and automatically generate Table Schema

| Option    | Data Mode| Create Table Schema
| -------- | ------- | ------- |
| Option 1: Base on Checkpoint  | Offchain    | Manual
| Option 2: Listen Database And Genarate Transaction Stream  | Offchain     | Manual
| Option 3: Auto generate table schema    | Onchain    | Auto|

### Automatic table creation solution

**Architecture**
<img width="729" alt="image" src="https://github.com/rooch-network/rooch/assets/1904567/224ece33-183b-4d9f-bb75-822afd08eac0">

**Core Process**
1. Parse SMT+NodeStore and get the AnnotatedMoveValue corresponding to State and Event.
2. Parse AnnotatedMoveValue and generate Table Schema Metadata and Data Schema Metadata
3. Rely on Table Schema Template to convert Table Schema Metadata and Data Schema Metadata into Table Schema and Data Schema
4. Determine whether the Table Schema is created. If not, create the Table first, then create the index; then write the data.
5. Repeat the above process

**How to parse**
1. Automatically parse SMT leaf nodes and obtain the Type type. Problem: crate dependency problem, need to rely on statedb

Entry: statedb module
Call the method AnnotatedStateReader::view_value

```
fn view_value(&self, ty_tag: &TypeTag, blob: &[u8]) -> 
Result<AnnotatedMoveValue> {
    let annotator = MoveValueAnnotator::new(self);
    annotator.view_value(ty_tag, blob)
}
```

2. Parse TableChange in StateChangeSet at the storage layer. The data structure is <Vec<u8>, Op<State>> and get the State type.

```
pub struct State {
    /// the bytes of state
    pub value: Vec<u8>,
    /// the type of state
    pub value_type: TypeTag,
}
``` 

state entry statedb module
     distinguish:
     resource
     module
    
[...]

PR #1088, opened 2023-10-31
Provide indexer store base on SQLite

### LINKED ISSUE #384: [Indexer] Provide indexer store base on SQLite
PostgreSQL vs Sqlite

Sqlite is embedded and does not require additional installation of Server, which will reduce the installation cost of developers. However, Sqlite does not support data partitioning, and there is a scalability limit for large-capacity data.
PostgreSQL has relatively complete functions. Both Sui and Aptos Indexer choose PostgreSQL, which requires additional installation, which can be regarded as one of the later expansion options.
Among them, GraphQL supports both PostgreSQL and Sqlite as endpoints.

It was decided to choose Sqlite as the Indexer DB in the early stage, and later maybe upgrade to PostgreSQL depending on the Indexer needs.

Choose diesel as the ORM operation database.

baichuan3:
## Summary

part of https://github.com/rooch-network/rooch/issues/1028

1. Provide indexer store base on SQLite with indexer config
2. Define `transaction` and `event` models and types
3. Define `transaction` and `event` db schema 
4. Implement part of `transaction` indexer
5. Implement part of `event` indexer

Continue
#1035
#1037

- Closes #384


crates/rooch-indexer/migrations/2023-10-31-085226_transactions/up.sql
```diff
@@ -0,0 +1,31 @@
+CREATE TABLE transactions (
+    tx_order                            BIGINT      NOT NULL      PRIMARY KEY,
+    tx_hash                             VARCHAR      NOT NULL,
+    transaction_type                    VARCHAR      NOT NULL,
+    sequence_number                     BIGINT      NOT NULL,
+    multichain_id                       VARCHAR      NOT NULL,
+    multichain_raw_address              BLOB         NOT NULL,
```
jolestar, line 7: Store `multichain_raw_address` as hex?

jolestar (approved):

crates/rooch-indexer/migrations/2023-10-31-085226_transactions/up.sql
baichuan3: fixed

Issue #1055, opened 2023-10-28
[Framework] Make Rooch Move Framework stable and promise compatibility


jolestar:
Make Rooch Move Framework (RoochFramework & MoveosStd ) stable and promise compatibility in future releases.

- [x] #1001 
- [x] #1056
- [x] #1066
- [x] #1067

PR #1090, opened 2023-11-01
[moveos_std] Refactor object::transfer

jolestar:
## Summary

1. Simply the singleton Object API. Singleton Object is always SystemOwned and can not be removed or transferred. new_singleton<T>() return &mut Object<T>.
2. object::transfer take `Object<T>`, and provide a `context::take_object` for owner.
3. Provide `context::fresh_uid` and `object::new(UID,T)` to avoid the two &mut reference problem(The NFT example test).

There are two types of Objects:

1. SystemOwnedObject: Object<T> is held by code in the state, or it is shared or frozen.
2. UserOwnedObject: Object<T> is held by the owner.

Part of #1055  

TODO:  

1. AccountStorage is a UserOwnedObject. How do we prevent the user from taking it from the object storage?


moveos/moveos-stdlib/moveos-stdlib/sources/context.move
```diff
@@ -65,11 +66,16 @@ module moveos_std::context {
         tx_context::fresh_address(&mut self.tx_context)
     }
 
-    /// Generate a new unique object ID
+    /// Generate a new unique ObjectID
     public fun fresh_object_id(self: &mut Context): ObjectID {
```
WGB5445, line 70: Is this interface necessary because it destroys the life cycle of objectID and Object, and now they are not unified.
I can create ObjectID but not Object which feels weird

jolestar: I has removed it,but add it back.
Because 

1. Some places depend on it.
2. UID need to provide a public function to convert UID to ObjectID. 
3. Developers can pass arbitrary ObjectID by transaction arguments.

moveos/moveos-stdlib/moveos-stdlib/sources/object.move
```diff
@@ -74,18 +89,26 @@ module moveos_std::object {
...
         let entity = new_internal(id, value);
         add_to_global(entity);
         Object{id}
     }
 
-    /// Create a new singleton object, singleton object is always owned by `System` and is p
+    /// Create a new singleton object, singleton object is always owned by `System`
     /// Singleton object means the object of `T` is only one instance in the Object Storage.
-    public(friend) fun new_singleton<T: key>(value: T): Object<T> {
+    public(friend) fun new_singleton<T: key>(value: T): &mut Object<T> {
```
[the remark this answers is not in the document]
jolestar: `borrow_object` and `borrow_object_mut_extend` support singleton object, `object::singleton_object_id` is public.

Issue #1103, opened 2023-11-03
[CLI] move integration-test command error


pause125:
When use move integration-test, an error occurs.

```
rooch move integration-test -p examples/counter/ --named-addresses rooch_examples=default
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: "Invalid address literal. The numeric value is too large. The maximum size is 32 bytes"', crates/rooch/src/commands/move_cli/commands/integration_test.rs:155:26
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

PR #1106, opened 2023-11-04
fixing the address parsing in integration test

stevenlaw123:
resolve https://github.com/rooch-network/rooch/issues/1103


crates/rooch/src/commands/move_cli/commands/integration_test.rs
```diff
@@ -340,3 +339,22 @@ impl IntegrationTest {
         Ok(())
     }
 }
+
+fn parse_account_arg(account: String) -> Result<AccountAddress, RoochError> {
```
jolestar, line 343: We should use the `WalletContext::parse_account_arg` https://github.com/rooch-network/rooch/blob/main/crates/rooch-rpc-client/src/wallet_context.rs#L73-L75

jolestar (approved):

PR #1108, opened 2023-11-04
[event] Refactor event, make Event API stateless

### LINKED ISSUE #1056: [MoveosStd] Simplify the Event API, remove the requirement for argument `&mut Context`

jolestar:
Follow #1090 

## Summary

Refactor events and make Event API stateless.

1. Remove EventHandle from Move.
2. Introduce TransactionEvent, and the events of TransactionOutput are TransactionEvent.
3. Record EventHandle in the event store.
4. Removet `&mut Context` from `event::emit`
5. Add the `drop` ability requirement to event `emit<T: drop>(e:T)`
6. Refactor Framework & Examples
7. Rename `Event::type_tag` to `Event::event_type`, and change type to `struct_tag`.

- Closes #1056

### DIFF
--- a/crates/rooch-rpc-server/src/server/rooch_server.rs
+++ b/crates/rooch-rpc-server/src/server/rooch_server.rs
@@ -187,10 +187,11 @@ impl RoochAPIServer for RoochServer {
         };
 
         let has_next_page = (data.len() as u64) > limit_of;
-        data.truncate(limit_of as usize);
         let next_cursor = data
             .last()
             .map_or(cursor, |event| Some(event.event_id.event_seq));
+
+        data.truncate(limit_of as usize);
 
         Ok(EventPageView {
             data,
--- a/crates/rooch-rpc-server/src/server/rooch_server.rs
+++ b/crates/rooch-rpc-server/src/server/rooch_server.rs
@@ -8,7 +8,6 @@ use jsonrpsee::{
     RpcModule,
 };
 use moveos_types::h256::H256;
-use moveos_types::moveos_std::event::EventID;
 use rooch_rpc_api::jsonrpc_types::{account_view::BalanceInfoView, StateOptions};
 use rooch_rpc_api::jsonrpc_types::{
     transaction_view::TransactionWithInfoView, EventOptions, EventView,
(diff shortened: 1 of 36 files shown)


moveos/moveos-store/src/event_store/mod.rs
```diff
@@ -47,33 +58,77 @@ pub trait EventStore {
...
+                    .get_mut(&tx_event.event_type)
+                    .expect("Event handle must exist");
+                let event_id = EventID::new(handle.id, handle.count);
+                let event = Event::new(
+                    event_id,
+                    tx_event.event_type,
+                    tx_event.event_data,
+                    tx_event.event_index,
+                );
+                handle.count += 1;
```
baichuan3, line 119: If parallel optimization is done in the future, what impact will this have on parallelization and transaction rearrangement?

baichuan3 (approved):

crates/rooch-rpc-server/src/server/rooch_server.rs
```diff
@@ -163,54 +162,37 @@ impl RoochAPIServer for RoochServer {
...
                 .collect::<Vec<_>>()
         };
 
         let has_next_page = (data.len() as u64) > limit_of;
-        data.truncate(limit_of as usize);
         let next_cursor = data
             .last()
             .map_or(cursor, |event| Some(event.event_id.event_seq));
 
+        data.truncate(limit_of as usize);
```
baichuan3, line 194: In the previously implemented version, when the cursor is not None, the cursor is skipped the next time the page is turned and returned from the cursor+1 position. This semantic has changed here.

moveos/moveos-store/src/event_store/mod.rs
jolestar: I've thought about this and have come to a conclusion: parallel executing does not affect events. 

Because in Rooch, the transaction is ordered by the sequencer first, the parallel executing does change the order, so we need to ensure the events are written to store as the transaction order. The events stored in different nodes will guarantee consistency.

crates/rooch-rpc-server/src/server/rooch_server.rs
jolestar: As the testsuite case of this version:

1. cursor:0(None), limit:1, result: Event{event_seq:0}, next_cursor: 1
2. cursor:1, limit:1, result: Event{event_seq:1}, next_cursor: 2

If the previous version, should it be like the following?

1. cursor:-1(None), limit:1, result: Event{event_seq:0}, next_cursor: 0
2. cursor:0, limit:1, result: Event{event_seq:1}, next_cursor: 1

If the last item of the result is the same as the next_cursor, the next_cursor seems useless.

baichuan3: Now the semantics of `next_cursor` is exactly the cursor that currently returns the result "the last item of the current page".

The earliest version used your method, and there was a problem:

When `has_next_page` is true, the normal `next_cursor` is returned; when `has_next_page` is false, the cursor corresponding to the last item of data is returned, which requires special processing.

For example, the following event only has 2 items of data:

1. cursor:-1(None), limit:1, result: Event{event_seq:0}, next_cursor: 1
2. cursor:1, limit:1, result: Event{event_seq:1}, next_cursor: 1

The previous DC discussion is here, https://discord.com/channels/1078938449974935592/1117770319436984341/1118013965189001216

jolestar: We need to discuss this, and I will refactor this later.

PR #1121, opened 2023-11-08
Implement transaction indexer

### LINKED ISSUE #1035: [Indexer] Implement transaction indexer

baichuan3:
## Summary

part of https://github.com/rooch-network/rooch/issues/1028

1. Implement indexer actor
2. Implement transaction indexer write
3. Simplified Indexer store implementation, removing asynchronous and multi-threading
4. Optimized the `Server start` failure prompt and supported `Server clean` to give the corresponding network

**TODO**
1. Solve the problem of automatic creation of Sqlite schema when `Server start` and integration test. There are two option
   - Use shell script to call `diesel migrate run` to complete schema initialization, which means that `Server start` and integration test have pre-dependencies
   - Simulate `diesel migrate run` in rooch cli to eliminate pre-dependencies and the user experience will be smoother.

5. Add more testcase

- Closes #1035


crates/rooch-indexer/.env
```diff
@@ -0,0 +1 @@
+DATABASE_URL=/Users/Baichuan/.rooch/local/roochdb/indexer.sqlite
```
jolestar, line 1: Is the `DATABASE_URL` must an absolute path?

crates/rooch-indexer/src/lib.rs
```diff
@@ -32,31 +32,73 @@ pub type SqlitePoolConnection = diesel::r2d2::PooledConnection<ConnectionManager
...
+    pub fn new(db_url: &str) -> Result<Self> {
+        let sqlite_cp = new_sqlite_connection_pool(db_url)?;
+        let store = Self {
+            sqlite_store: SqliteIndexerStore::new(sqlite_cp),
+        };
+        Ok(store)
+    }
+
+    pub fn mock_indexer_store() -> Result<Self> {
+        let tmpdir = moveos_config::temp_dir();
```
jolestar, line 58: Should keep the `temp_dir()` returned `DataDirPath`. Otherwise, the temp dir will be removed when the DataDirPath instance is dropped.

crates/rooch-indexer/src/actor/indexer.rs
```diff
@@ -0,0 +1,59 @@
...
+        let indexed_transaction =
+            IndexedTransaction::new(transaction, sequence_info, execution_info, moveos_tx)?;
+        let _transactions = vec![indexed_transaction.clone()];
+        // self.indexer_store.persist_transactions(transactions)?;
+        Ok(indexed_transaction)
+    }
+}
+
+#[async_trait]
+impl Handler<QueryTransactionsByHashMessage> for IndexerActor {
```
jolestar, line 49: We can query transactions by hash from the transaction store, and does not need the index. Providing query transactions by the sender is more useful.

crates/rooch-indexer/.env
baichuan3: the `.env` is only for dev, and It should be removed from the crate.

crates/rooch-indexer/src/actor/indexer.rs
baichuan3: Yes, the `transactionFilter` query RPC will be implemented in the following pr later to support queries including sender and other dimensions.

crates/rooch-indexer/src/lib.rs
baichuan3: Fixed the problem of using temp path to release instance when server starts.
As for whether calling `mock_indexer_store` in the integration test will release it, I will write a test case to cover and verify it later.

jolestar:
I will merge this PR first.

jolestar (approved):

Issue #1117, opened 2023-11-06
[MoveosStd] Math modules


jolestar:
Introduce some math modules for calculating, such as `pow`,`sqrt`,`mul_div`.

### Reference

* https://github.com/starcoinorg/starcoin-framework/blob/345a3900a0064dc57a9560235bc72c12f03448b1/sources/Math.move
* https://github.com/aptos-labs/aptos-core/blob/15280ac4cbb430e81070d2df0732064631b2e23f/aptos-move/framework/aptos-stdlib/sources/math128.move
* https://github.com/MystenLabs/sui/blob/da72e73a9f4a1b42df863b76af7745a4eeb1d412/crates/sui-framework/packages/sui-framework/sources/math.move

feliciss:
I may have time to look at this issue after finishing #771.

PR #1132, opened 2023-11-12
Math module (#1117)

HakanVardarr:
## Summary

I implemented math module for rooch. If you have any suggestions I can work on them.


jolestar:
I think we need a Math module for Move but not for Rust. https://github.com/rooch-network/rooch/issues/1117

Issue #997, opened 2023-10-19
[Multi-chain] Bitcoin Relayer


jolestar:
Relayer the Bitcoin BlockHeader to Rooch

Issue #423, opened 2023-07-07
[Framework] Bitcoin light client with Move


jolestar:
Implement a Bitcoin light client with Move.

1. Receive a Bitcoin header from the entry function argument.
2. Decode the header.
3. Verify the header via Nakamoto Consensus (Difficulty and the Longest Chain).
4. Record the header to state.
5. Provide a function to verify the transaction Merkle proof.

feliciss:
I'm interested in this issue. However, I may complete issue #406 first.

jolestar:
Now, we implement a full Bitcoin client in Move. the verifier Header requirement tracked via #1330

PR #1169, opened 2023-11-20
[Bitcoin] PoC implementation of bitcoin relayer

jolestar:
## [Bitcoin] PoC implementation of bitcoin relayer

1. Upgrade bitcoin crate to 0.31.0
2. Implement bitcoin relayer.

Part of #423 and #997


crates/rooch-framework/sources/bitcoin_light_client.move
```diff
@@ -0,0 +1,76 @@
...
+
+        let block_store = context::borrow_mut_resource<BlockStore>(ctx, @rooch_framework);
+        if(table::contains(&block_store.blocks, block_header.hash)){
+            //TODO handle repeat block hash
+            return
+        };
+        table::add(&mut block_store.blocks, block_header.hash, block_header);
+
+        let timestamp_seconds = (block_header.time as u64);
+        timestamp::try_update_global_time(ctx, timestamp::seconds_to_milliseconds(timestamp_seconds));
```
baichuan3, line 55: Does different relayers can update the global timestamp?

crates/rooch-config/src/lib.rs
```diff
@@ -82,9 +82,28 @@ pub struct RoochOpt {
...
-    #[clap(long)]
+    #[clap(long, env = "ETH_RPC_URL")]
     pub eth_rpc_url: Option<String>,
 
+    /// The Bitcoin RPC URL to connect to for relay L1 block and transaction to L2.
+    /// If not set, the relayer service will not start.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[clap(
+        long,
+        env = "BITCOIN_RPC_URL",
```
baichuan3, line 93: Is env configured through `.env` file?

crates/rooch-framework/sources/bitcoin_light_client.move
jolestar: Maybe we just run one relayer.

crates/rooch-config/src/lib.rs
jolestar: From the env variable. Some users do not wants expose the password to the cli command argument, can use the env.

baichuan3 (approved):

feliciss (commented):
ACK.

Issue #993, opened 2023-10-18
[UseCase] Rooch as Bitcoin Execution Layer


jolestar:
## Rooch as Bitcoin Execution Layer

<img width="1263" alt="Screen Shot 2023-11-25 at 11 45 14" src="https://github.com/rooch-network/rooch/assets/77268/b869b046-6be3-4e24-809a-546e5efc4583">

### 1. **Core Components & Interactions**:

- **Bitcoin**: The main chain. Its block information is synced to Rooch, providing verification data for Rooch.

- **Rooch**: As a Execution layer, it includes the following core components:
  - **Bitcoin Client**: Implemented Bitcoin logic using the Move smart contract. It's responsible for validating block header and progress transactions synced from Bitcoin.
  - **Relayer**: Periodically syncs the latest block from Bitcoin to Rooch's Bitcoin Client. Anyone can assume the role of a Relayer, ensuring that as long as there's at least one honest Relayer, the system remains secure. In the future, we can directly connect to Bitcoin P2P network.
  - **Rooch Sequencer**: Interacts with application contracts within Rooch, used for verifying transactions and handling relevant application logic.
  - **Rooch Full Node**: Stores the complete data of the Rooch chain and interacts with the Sequencer or DA to sync transaction information.

- **Data Availability Layer (DA)**: An external layer employed by the Rooch Rollup to ascertain the verifiability of Rooch transactions.
- **Client**: End-users or applications that engage with Bitcoin Network and Rooch's application contracts.

### 2. **Workflow**:

1. Relayers regularly sync Bitcoin's block information to the Bitcoin Client in Rooch.
2. The Bitcoin Client in Rooch validates and stores this block header and UTXO information.
3. Developers can create applications in Rooch using Move smart contracts. These applications can directly read Bitcoin state(BlockHeader and UTXO) or read&write Application state.
4. The client interacts with application contracts.

### 3. **Use Cases**:

- **Trigger-based Applications**: For instance, when a specific transaction is completed on the Bitcoin network, it automatically triggers a contract execution within Rooch.
  
- **Data Replication & Re-execution**: For example, replicating and re-executing standard definitions like Ordinals, BRC20, GRC20, etc., from Bitcoin in Rooch, positioning Rooch as a decentralized Indexer service. This way, anyone can deploy a Rooch node, synchronize, and re-execute transactions to establish their own Indexer service.

### Relative Issues

- [x] #992 
- [x] #423 
[...]

PR #1182, opened 2023-11-28
[Bitcoin] Implement bitcoin Inscription parse in Move

jolestar:
## Summary

Add a native function to parse Inscription.

Part of #423 #993

### DIFF
--- a/crates/rooch-framework/sources/ord.move
+++ b/crates/rooch-framework/sources/ord.move
@@ -1,0 +1,46 @@
+// Copyright (c) RoochNetwork
+// SPDX-License-Identifier: Apache-2.0
+
+module rooch_framework::ord {
+    use std::vector;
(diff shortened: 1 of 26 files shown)


crates/rooch-framework/sources/ord.move
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) RoochNetwork
+// SPDX-License-Identifier: Apache-2.0
+
+module rooch_framework::ord {
```
baichuan3, line 4: We can put the Bitcoin contracts in a separate directory.

jolestar: We have the plan to introduce a new `bitcoin_framwork.`

feliciss (commented):
ACK.

PR #1215, opened 2023-12-05
Deprecate `std::error` 

### LINKED ISSUE #1184: [Discussion] Deprecating the Move error code?
Currently, we use the original Move error code mechanism to generate the error code.

The error code includes a category and a code, and the developer needs to explain the error code to split the category and code.

How about we remove the category from the error code and abort with the code? The sui framework does not use the `std::error` module to generate error codes.

pause125:
## Summary

Deprecate the category from the error code, and remove `std::error` module. 

- Closes #1184


baichuan3:
After removing the category, can we consider whether move stdlib, moveos stdlib and rooch framework allocate different error code ranges in a conventional way, so that the use and error location will be more straightforward.

crates/rooch-framework/sources/account.move
```diff
@@ -287,7 +286,7 @@ module rooch_framework::account {
    }
 
    #[test(sender=@0x0)]
-   #[expected_failure(abort_code = 0x10005, location = Self)]
+   #[expected_failure(abort_code = 5, location = Self)]
```
jolestar, line 289: We can use `ErrorAddressReseved` to replace 5 here, which is clearer.

jolestar (approved):

jolestar:
> After removing the category, can we consider whether move stdlib, moveos stdlib and rooch framework allocate different error code ranges in a conventional way, so that the use and error location will be more straightforward.

The error code is located via module + error code. It is hard to maintain an error code range for every module. So, keep it simple.

baichuan3 (approved):

Issue #1193, opened 2023-11-30
[MoveosStd] Store the TypeTag of key in the TableInfo


jolestar:
relative issue #1179

Issue #1179, opened 2023-11-24
[RPC] how to use `rooch_listStates` to list table keys


jolestar:
`rooch_listStates` by `/table/$table_handle/` only can list value, how to list table keys?

The table is `Table<BTCAddress, u64>`

```bash
curl -H "Content-Type: application/json" -X POST \
--data '{"jsonrpc":"2.0","method":"rooch_listStates","params":["/table/0x05b585fab25e52849b5553ffb72803bf0b6768eca781a2101f3f054e3be94d65/", null ,null, {"decode":true}],"id":1}' \
http://localhost:50051
```

```json
[...]

PR #1223, opened 2023-12-12
[Moveosstd] Refactor tableinfo create and store typetag in table info

### LINKED ISSUE #1192: [MoveosStd] Create TableInfo object in Move module
Currently, the TableInfo object is auto-created when writing state to the state db.

There are some problems with this approach:

1. The TableInfo object does not in the transaction ChangeSet.
2. It is hard to auto-index for table changes. #1186 

So, we should create the TableInfo object in the Move module.

baichuan3:
## Summary

Create TableInfo object in Move module to make easy auto-index for table changes.

1. Add table new native function, create table data and table info, and store table info object in contract, excluding system reserve accounts, like 0x1, 0x2, 0x3...
2. Store type tag in TableInfo and TableChange, prepare for #1179
3.  For system reserve accounts (0x1, 0x2, 0x3...), the contract deployment precedes the table creation operation, so the module table cannot be created directly in the contract. The table data and table info of such accounts are completed in the VM, the TableInfo object is updated and maintenance (size and state root) is also handled in the VM to avoid having two sets of logic in the VM and the contract.

- Closes #1192 #1193


moveos/moveos-stdlib/src/natives/moveos_stdlib/raw_table/mod.rs
```diff
@@ -438,6 +490,69 @@ impl CommonGasParameters {
...
+    let mut table_data = table_context.table_data.write();
+
+    let handle = get_table_handle(&mut args)?;
+
+    let mut cost = gas_params.base;
+
+    let table = table_data.get_or_create_table(context, handle, &ty_args[0])?;
+
+    // Table SMT root
+    let state_root = table_context.resolver.resolve_state_root().map_err(|err| {
```
jolestar, line 521: new table's state_root should be the place holder hash.

baichuan3: fixed

jolestar (approved):

PR #1264, opened 2023-12-24
bump rust to 1.74

popcnt-subodhi:
## Summary

fix clippy lints warnings (1.74)


crates/rooch-indexer/src/types.rs
```diff
@@ -156,27 +158,27 @@ impl IndexedEvent {
 
 #[derive(Debug, Clone)]
 pub struct IndexedGlobalState {
-    /// The global state key
+    // The global state key
     pub object_id: ObjectID,
-    /// The owner of the object
```
jolestar, line 161: `///` on the field should be ok?

jolestar (approved):

crates/rooch-indexer/src/types.rs
popcnt-subodhi: @jolestar you're right. During the replacement, an inadvertent substitution occurred simultaneously.

PR #1273, opened 2023-12-27
Security enhancement for private_generics.

stevenlaw123:
1. During compilation, ensure that the type information in private_generics is complete and accurate.
2. During module deployment, check if the information in the current module's metadata is correct; otherwise, halt the deployment.

### DIFF
--- a/moveos/moveos-verifier/src/verifier.rs
+++ b/moveos/moveos-verifier/src/verifier.rs
@@ -392,7 +392,7 @@ where
                         .finish(Location::Module(module.self_id())));
                 }
 
-                let module_address = parts_vec.get(0).unwrap();
+                let module_address = parts_vec.first();
                 let module_name = parts_vec.get(1).unwrap();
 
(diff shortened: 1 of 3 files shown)


moveos/moveos-verifier/src/verifier.rs
```diff
@@ -380,6 +380,36 @@ where
...
+                let parts_vec = func_name_split.collect::<Vec<&str>>();
+                if (parts_vec.len() as u32) < 3 {
+                    return Err(PartialVMError::new(StatusCode::MALFORMED)
+                        .with_message(
+                            "incorrect format of the function name in metadata".to_string(),
+                        )
+                        .finish(Location::Module(module.self_id())));
+                }
+
+                let module_address = parts_vec.first().unwrap();
```
baichuan3, line 395: `unwrap` may panic verifier

jolestar (approved):

Issue #1173, opened 2023-11-23
Performance optimize


jolestar:
- [ ] Split executor to write executor and read executor
- [ ] Benchmark and Performance Bottleneck Analysis
- [ ] MoveVM performance optimize
- [ ] MoveVM Parallel execution

PR #1281, opened 2024-01-02
Split executor into writer executor and reader executor

baichuan3:
## Summary

Performance optimize and Part of #1173, Split executor to write executor and read executor

- Closes #issue


crates/rooch-executor/src/actor/reader_executor.rs
```diff
@@ -0,0 +1,626 @@
...
+        moveos_store: MoveOSStore,
+        rooch_store: RoochStore,
+    ) -> Result<Self> {
+        let moveos = MoveOS::new(
+            moveos_store,
+            genesis.all_natives(),
+            genesis.config.clone(),
+            system_pre_execute_functions(),
+            system_post_execute_functions(),
+        )?;
```
jolestar, line 82: The writer executor needs to send a message to the reader executor after the transaction is executed and let the read executor update the state root.

jolestar (approved):

crates/rooch-executor/src/actor/reader_executor.rs
jolestar: We can try and refactor it later.

PR #1290, opened 2024-01-10
Review moveos-stdlib and rooch-framework native functions, fix gas meters

pause125:
## Summary

Review moveos-stdlib and rooch-framework native functions, and fix gas meters


baichuan3 (approved):

feliciss (commented):
ACK.

PR #1401, opened 2024-03-01
Check the Object reference count runtime

jolestar:
## Summary

Check the Object reference count runtime

1. Count the Object reference count and forbid two references from the same object.
2. Refactor the `to_runtime_value` and `from_runtime_value` trait

TODO:
1. Increment the reference count when passing Object reference from the arguments.

### DIFF
--- a/moveos/moveos-stdlib/src/natives/moveos_stdlib/object.rs
+++ b/moveos/moveos-stdlib/src/natives/moveos_stdlib/object.rs
@@ -1,22 +1,21 @@
 // Copyright (c) RoochNetwork
 // SPDX-License-Identifier: Apache-2.0
 
-use move_binary_format::errors::PartialVMResult;
-use move_core_types::gas_algebra::InternalGas;
+use crate::natives::{helpers::make_module_natives, moveos_stdlib::raw_table::NativeTableContext};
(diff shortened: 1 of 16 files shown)


moveos/moveos-stdlib/src/natives/moveos_stdlib/object.rs
```diff
@@ -113,7 +110,20 @@ fn borrow_object_reference(
...
-    //TODO should we keep the GlobalValue in Context?
+    let object_id = ObjectID::from_runtime_value(object_id_value)
+        .map_err(|_e| partial_extension_error("Invalid object id argument"))?;
+    let table_context = context.extensions_mut().get_mut::<NativeTableContext>();
+
+    let data = table_context.table_data();
+    let mut table_data = data.write();
+    let gv = table_data.get_or_create_object_reference(object_id.clone())?;
+
+    if gv.reference_count() >= 2 {
```
baichuan3, line 121: Need to differentiate between `ref` and `mut ref`? Multiple `refs` should be reasonable

jolestar: The GlobalValue does not distinguish between ref and mut ref. I record an issue.

baichuan3 (approved):

PR #1414, opened 2024-03-05
[Bitcoin] Prepare Ord data verify tools and begin to verify

### LINKED ISSUE #1424: [RPC] Make TransactionOutput from `execute_raw_transaction` to optional
The TransactionOutput may verify big if the transaction changes to too many states; 
we can make it optional to improve the performance.
We can only return the output when the RPC argument requires it.

```rust
pub struct ExecuteTransactionResponseView {
    pub sequence_info: TransactionSequenceInfoView,
    pub execution_info: TransactionExecutionInfoView,
    pub output: TransactionOutputView,
}

```

to 

```rust
pub struct ExecuteTransactionResponseView {
    pub sequence_info: TransactionSequenceInfoView,
    pub execution_info: TransactionExecutionInfoView,
    pub output: Option<TransactionOutputView>,
}
```

baichuan3:
## Summary

Ord and Brc20 data verification tool. The Brc20 tool has not been implemented yet (TBD).

Ord verification uses two data sources, one is the full Indexer data exported by Ord cli, and the other is the Rooch Indexer data source.

The Ord verification process is that the Inscription data exported by Ord cli can be queried in the Rooch indexer based on txid+index, and then the owner address and body+content_type are verified.

1. Ord data verification tool.
2. Rooch Indexer supports querying Inscription by Bitcoin's `txid+index`
3. Rooch Indexer supports querying UTXO by Bitcoin's `txid+vout`
4. Added data verify mode. In this mode, UTXO Indexer writing and updating will be turned off to improve writing performance.
5. Make TransactionOutput from execute_raw_transaction to optional
6. Make the bitcoin client as actor in the bitcoin relayer to improve bitcoin blocks import performance

TODO
1. Relayer imported BTC data to the Rooch node, and it took 6 hours to import 176 blocks, averaging 29 blocks per hour, which was much slower than expected. Import performance needs to be determined.
2. Import 10,000 BTC blocks to the Rooch node. The Indexer file exceeds 100G and has a reading bottleneck that needs to be optimized.

- Closes #1424


jolestar (approved):

CryptoMasterTime:
no problem

baichuan3:
I will merge this first and continue to verify ord data

PR #1419, opened 2024-03-06
[framework] Refactor Object dynamic fields

jolestar:
## Summary

1. Remove `object_dynamic_field.move` and migrate the dynamic field functions to the object module.
2. Update Object::size field in Move: `add_fields` and `remove_fields`.
3. Introduce RootObject, and record the Root object's size in `TransactionExecutionInfo` and `StartupInfo`.
4. Refactor statedb, and simplify the logic.

TODO:

1. Refactor the `object::new(id, T)` => `object::new(T)`, auto-generate the ObjectID based on `hash(state_root|size)`


moveos/moveos-types/src/moveos_std/table.rs
```diff
@@ -0,0 +1,38 @@
...
+use move_core_types::{
+    account_address::AccountAddress, ident_str, identifier::IdentStr, language_storage::StructTag,
+    value::MoveStructLayout,
+};
+use serde::{Deserialize, Serialize};
+
+pub const MODULE_NAME: &IdentStr = ident_str!("table");
+
+#[derive(Eq, PartialEq, Debug, Clone, Deserialize, Serialize)]
+pub struct TablePlaceholder {}
```
baichuan3, line 17: `TypeTable.move` also needs to define `TablePlaceholder` in Rust? Will it be repeated?

baichuan3 (approved):
LGTM

moveos/moveos-types/src/moveos_std/table.rs
jolestar: Yes, but it is not used in Rust now.

Issue #1443, opened 2024-03-14
[test] Figure out why rooch-framework-tests is so slow


jolestar:
```
cargo nextest run -p rooch-framework-tests -p rooch-integration-test-runner -v -j 1
```

<img width="1207" alt="Screen Shot 2024-03-14 at 09 30 39" src="https://github.com/rooch-network/rooch/assets/77268/f6c753a6-4c41-435d-b933-b4f26eb90f14">

jolestar:
> It seems normal with command `cargo test -p rooch-framework-tests -p rooch-integration-test-runner -j 1`, maybe we can use `cargo test` to do these tests.

Help to update the GitHub action file. 

cc @shelmesky

steelgeek091:
> > It seems normal with command `cargo test -p rooch-framework-tests -p rooch-integration-test-runner -j 1`, maybe we can use `cargo test` to do these tests.
> 
> Help to update the GitHub action file.
> 
> cc @steelgeek091

OK.

jolestar:
I think the reason is that the nextest does not share the static variable between tests, so it needs to rebuild Genesis for every test. 

https://nexte.st/book/how-it-works.html

PR #1456, opened 2024-03-18
[moveos] support object field

jolestar:
## Summary

1. Make ObjectID to a path. ObjectID{address} => ObjectID{vector<address>}.
2. Add `add_object_field`,`borrow_object_field`,`borrow_mut_object_field`,`remove_object_field` to object module.
3. Refactor account module.
4. `object::borrow_object,` and `object::borrow_mut_object` also support child objects, but `object::remove` does not support removing child objects.

TODO: add more tests and examples. 

Part of #1423

PR #1458, opened 2024-03-19
use cargo test instead of cargo nextest

steelgeek091:
resolve https://github.com/rooch-network/rooch/issues/1443


jolestar (approved):

jolestar:
included in #1456

PR #1489, opened 2024-03-27
feat(moveos-store): buf write nodes when apply change_set

popcnt-subodhi:
## Summary

helping to reduce I/O cost when change_set is large


popcnt-subodhi:
done here:

https://github.com/rooch-network/rooch/pull/1549

PR #1544, opened 2024-04-09
feat(types): using bsc::serialized_size return tx_size

popcnt-subodhi:
## Summary

l1_block_encode_siz: [29.338 ns 29.367 ns 29.409 ns]

l1_block_serialized_size: [11.943 ns 11.973 ns 12.011 ns]


crates/rooch-types/src/transaction/ledger_transaction.rs
```diff
@@ -24,8 +24,7 @@ impl L1Block {
     }
 
     pub fn tx_size(&self) -> u64 {
-        //TODO optimize the size calculation
-        self.encode().len() as u64
+        bcs::serialized_size(self).expect("serialize size should success") as u64
```
jolestar, line 27: RoochTransaction also has this function

PR #1548, opened 2024-04-10
feat(tx): cache data_hash for RoochTransaction

popcnt-subodhi:
### pros.

1. saving redundant calculation

### cons.

1. change struct of RoochTransaction, the result of serilization changed
2. mut tx

### other approach

1. outside cache with limit @jolestar

### DIFF
--- a/crates/rooch-types/src/transaction/rooch.rs
+++ b/crates/rooch-types/src/transaction/rooch.rs
@@ -75,6 +75,8 @@ impl RoochTransactionData {
 pub struct RoochTransaction {
     data: RoochTransactionData,
     authenticator: Authenticator,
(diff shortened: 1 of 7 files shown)


crates/rooch-types/src/transaction/rooch.rs
```diff
@@ -75,13 +75,16 @@ impl RoochTransactionData {
 pub struct RoochTransaction {
     data: RoochTransactionData,
     authenticator: Authenticator,
+
+    data_hash: Option<H256>,
```
jolestar, line 79: Need to skip from serializing

jolestar (approved):

PR #1551, opened 2024-04-11
[Bitcoin] Implement one Input while Inscribing multiple inscriptions via Pointer

baichuan3:
## Summary

Case: https://ordinals.com/inscription/6ea3bf728b34c8c01ba4703e00ad688be100599b92fbdac71e6aea6ad8355552i0

- Closes #issue


crates/rooch-types/src/into_address.rs
```diff
@@ -93,4 +93,19 @@ mod tests {
...
+            "0x525535d86aea6a1ec7dafb929b5900e18b68ad003e70a41bc0c8348b72bfa36e",
+        )
+        .unwrap();
+        //The txid hex string use reverse order
+        let txid = bitcoin::Txid::from_byte_array(addr.into());
+        println!("{}", txid);
+        // assert_eq!(
+        //     "6ea3bf728b34c8c01ba4703e00ad688be100599b92fbdac71e6aea6ad8355552",
+        //     txid.to_string()
+        // );
```
jolestar, line 109: Why remove this check?

jolestar (approved):

jolestar:
Should we remove https://github.com/rooch-network/rooch/pull/1551/files#diff-c46533574b67286155c790a07ffc1cd2a55965dd572c0c4f5319af22c395fbf9L370-L374 ?

baichuan3:
> https://github.com/rooch-network/rooch/pull/1551/files#diff-c46533574b67286155c790a07ffc1cd2a55965dd572c0c4f5319af22c395fbf9L370-L374

Still looking for such a case, which can only be determined after verification of the case: when batch inscription, the receiver address exceeds 1 address, and the number of inscriptions is greater than the number of receivers.

crates/rooch-types/src/into_address.rs
baichuan3: Will put it back in the next PR

PR #1556, opened 2024-04-12
[Bitcoin] Refine inscribe inscriptions flow

baichuan3:
## Summary

1. Allow transfer and inscribe at the same transaction.
2. Confirm batch inscribe logic is correct handling.
  There are three modes for Inscribe: SameSat, SeparateOutputs, SharedOutput
   SameSat and SharedOutput mode have only one output.
   When SeparateOutputs is used, the number of output and inscription is consistent.

- Closes #issue


baichuan3:
Case: Transfer and inscribe may happen at the same transaction
https://ordinals.com/tx/207322afdcca902cb36aeb674214dc5f80f9593f12c1de57830ad33adae46a0a

PR #1577, opened 2024-04-15
[Bitcoin] Make Inscription object as child object

baichuan3:
## Summary

Summary about this PR

- Closes #issue


jolestar (approved):

jolestar:
The inscription RPC also need to be updated.

https://github.com/rooch-network/rooch/blob/0af1abe1ec8fd38f514c92d4aec49eb1cf4a8eb4/crates/rooch-rpc-api/src/jsonrpc_types/btc/ord.rs#L60-L66

PR #1589, opened 2024-04-18
feat(store): refine rocksdb config

popcnt1:


crates/rooch-rpc-server/src/lib.rs
```diff
@@ -440,7 +440,7 @@ pub fn init_storage(
     let rooch_store = RoochStore::new(StoreInstance::new_db_instance(RocksDB::new(
         rooch_db_path,
         rooch_store::StoreMeta::get_column_family_names().to_vec(),
-        store_config.rocksdb_config(),
+        store_config.rocksdb_config(false),
```
jolestar, line 443: Maybe we should share the StoreInstance between RoochStore and MoveOSDB.

jolestar (approved):

crates/rooch-rpc-server/src/lib.rs
popcnt1: Yes, that's what I plan to do. Two instance is annoying

Issue #1214, opened 2023-12-05
[Example] Bitcoin Plants -- A Inscription fully on-chain Game 


jolestar:
# Bitcoin Plants

## Gameplay Description:

1. Player mint seed Inscription on Bitcoin.
2. The player plants the seeds on Rooch Layer 2.
3. Players need to water the plant on Rooch Layer 2 regularly. The seeds will grow, sprout, and bear fruit. The fruit can be harvested. The fruit is represented by Layer2's Coin or NFT.
4. If a player transfers the seed Inscription on Bitcoin, the plant produced by this seed will also be transferred.
5. It will wither if the plant is not watered too long. After watering again, it will start growing from the beginning.

## Tasks

1. Read the Inscription on the Rooch Move contract
2. Connect to Bitcoin wallet
3. Write the game logic with Move

## Relative issues or PR

* #993 
* #997 
* #1182
* #1191 
* #423

PR #1608, opened 2024-04-23
Add permanent area for Inscription.

pause125:
## Summary

1. Add permanent_area for `bitcoin_move::ord::Inscription`
2. Add an incomplete example `bitcoin_plants`(#1214 ) which use the permanent area.

### DIFF
--- a/frameworks/bitcoin-move/sources/ord.move
+++ b/frameworks/bitcoin-move/sources/ord.move
@@ -12,6 +12,8 @@ module bitcoin_move::ord {
     use moveos_std::simple_map::{Self, SimpleMap};
     use moveos_std::json;
     use moveos_std::table_vec::{Self, TableVec};
+    use moveos_std::type_info;
+    use moveos_std::bag;
     use rooch_framework::address_mapping;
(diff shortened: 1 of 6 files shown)


jolestar (approved):

jolestar:
Inscription also needs the temp state area.

jolestar:
Need @@steelgeek091 help

frameworks/bitcoin-move/sources/ord.move
```diff
@@ -227,6 +226,8 @@ module bitcoin_move::ord {
                 inscription.offset = new_sat_point.offset;
 
                 // TODO handle curse inscription
+                // drop the temporary area if inscription is transferred.
+                drop_temp_area(&mut inscription_obj);
```
jolestar, line 230: We should define a custom transfer function for `Object<Inscription>`.

jolestar: It is a private function; ensure the `transfer_extend` and clean temp state are the same simultaneously.

PR #1611, opened 2024-04-24
feat: ord add metaprotocol_validity

yubing744:
## Summary

Summary about this PR

- Closes #issue

### DIFF
--- a/frameworks/bitcoin-move/sources/ord.move
+++ b/frameworks/bitcoin-move/sources/ord.move
@@ -30,6 +30,8 @@ module bitcoin_move::ord {
 
     const PERMANENT_AREA: vector<u8> = b"permanent_area";
     const TEMPORARY_AREA: vector<u8> = b"temporary_area";
+    
+    const METAPROTOCOL_VALIDITY: vector<u8> = b"metaprotocol_validity";
 
     /// How many satoshis are in "one bitcoin".
     const COIN_VALUE: u64 = 100_000_000;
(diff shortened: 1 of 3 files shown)


frameworks/bitcoin-move/sources/ord.move
```diff
@@ -619,5 +628,52 @@ module bitcoin_move::ord {
...
+
+        let protocol_type = type_info::type_name<T>();
+        let validity = MetaprotocolValidity {
+            protocol_type,
+            is_valid,
+            invalid_reason,
+        };
+
+        let validity_id = MetaprotocolValidityID{};
+        object::add_object_field_with_id<Inscription, MetaprotocolValidityID, MetaprotocolValidity>(inscription_obj, validity_id, validity)
```
jolestar, line 650: Why is MetaprotocolValidity an Object field? it should be a normal field?

yubing744: I follow the Open-Closed Principle and try not to modify the previous logic with new features. Fixing the previous logic may cause a large number of previous test cases to fail. But it feels like it's over-designed here, so I'd better change it to a normal field to make the logic clearer.

jolestar: I mean, we should use `object::add_field`, not `object::add_object_field`.

I think changing the Inscription struct not a good idea.

yubing744:
@jolestar help review

jolestar (approved):

PR #1645, opened 2024-04-30
[Bitcoin] Eliminate data import mode

baichuan3:
## Summary

1. Eliminate data import mode to avoid genesis change.
2. Keep the data import flag, used to determine whether to update the indexer
3. For statedb utxo genesis tool, adjust println output to debug log.

- Closes #issue


frameworks/bitcoin-move/sources/light_client.move
```diff
@@ -326,6 +326,10 @@ module bitcoin_move::light_client{
         let btc_block_store_obj = borrow_block_store();
         let btc_block_store = object::borrow(btc_block_store_obj);
         btc_block_store.latest_block_height
-    } 
+    }
+
+    public fun need_process_oridinals(block_height: u64) : bool {
+        block_height >= ORDINAL_GENESIS_HEIGHT
```
jolestar, line 332: Also, need to check the Bitcoin network

jolestar (approved):

Issue #1623, opened 2024-04-25
[move stdlib] Remove the `store` ability requirement from account resource function 


jolestar:
## Motivation

To keep compatibility with the Move `borrow_global|move_to|move_from` instruction, we need to remove the `store` ability requirement from the account resource function. In the future, we can support auto convert `borrow_global` to `account::borrow_resource`.

```move
public fun borrow_resource<T: key + store>(account: address): &T
```

to

```move
public fun borrow_resource<T: key>(account: address): &T
```

jolestar:
> How can we remove `store` ability as `T` is stored in object field, which requires `store` ability?

Use the object module's internal functions.

```move
fun add_field_internal<T: key, K: copy + drop, V>(obj_id: ObjectID, key: K, val: V) 
```

make it `public(friend)`.

```move
public fun account_move_resource_to<T: key + store>(self: &mut Object<Account>, resource: T){
      assert!(!object::contains_field(self, key<T>()), ErrorResourceAlreadyExists);
      object::add_field(self, key<T>(), resource)
   }
```

to

```move
public fun account_move_resource_to<T: key>(self: &mut Object<Account>, resource: T){
      assert!(!object::contains_field(self, key<T>()), ErrorResourceAlreadyExists);
      object::add_field_internal(object::id(self), key<T>(), resource)
   }
```

jolestar:
The `V` of `FieldValue<V>` does not require any ability.

```move
struct FieldValue<V> has key, drop, store {
    val: V
}
```

It should be worked.

PR #1647, opened 2024-04-30
Disable create account from arbitrary address and remove store ability for account resource function 

### LINKED ISSUE #1624: [move stdlib] Disable create account from arbitrary address
## Motivation

Because an Account is an Object, and the address is the ObjectID of `Object<Account>`, if we allow creating an account from an arbitrary address, someone may take the ObjectID before an Object is created(such as the named object or custom id object).

## Solution

1. Remove the entry function:

https://github.com/rooch-network/rooch/blob/5a26585c604fc2752aca385a268572140e046053/frameworks/rooch-framework/sources/account.move#L23-L37

2. Remove auto-create account when transfer coin and object

https://github.com/rooch-network/rooch/blob/5a26585c604fc2752aca385a268572140e046053/frameworks/rooch-framework/sources/transfer.move#L16-L27

pause125:
## Summary

1. Disable create account from arbitrary address
2. Remove the store ability requirement from account resource function 

- Closes #1624 #1623


frameworks/rooch-framework/sources/transfer.move
```diff
@@ -39,7 +39,7 @@ module rooch_framework::transfer {
         let maddress = multichain_address::new(multichain_id, raw_address);
         let to = address_mapping::resolve_or_generate(maddress);
         if(!account::exists_at(to)) {
-            account_entry::create_account(to);
+            account_entry::create_account(to); // TODO: creating or checking account?
```
jolestar, line 42: Do not create_account, and do not bind the address mapping.

frameworks/rooch-framework/sources/transfer.move
```diff
@@ -49,9 +49,7 @@ module rooch_framework::transfer {
     /// TODO: Currently, we can not pass the `Object<T>` argument by value, so, we use `ObjectID` instead.
     /// After the `Object<T>` argument can be passed by value, we should change the argument type to `Object<T>`.
     public entry fun transfer_object<T: key + store>(from: &signer, to: address, object_id: ObjectID) {
-        if(!account::exists_at(to)) {
-            account_entry::create_account(to);
-        };
+        assert!(account::exists_at(to), ErrorAccountNotExists);
```
jolestar, line 52: We should allow the transfer of an object to a no-existing account.

frameworks/rooch-framework/sources/transfer.move
```diff
@@ -19,10 +22,7 @@ module rooch_framework::transfer {
         to: address,
         amount: u256,
     ) {
-        if(!account::exists_at(to)) {
-            account_entry::create_account(to);
-        };
-
+        assert!(account::exists_at(to), ErrorAccountNotExists);
```
jolestar, line 25: We should allow the transfer of coins to the no-existing account.

jolestar (approved):

frameworks/rooch-framework/sources/transfer.move
jolestar: Likewise, if an account is created, but no one holds the private key for that address, it is also lost.

Issue #792, opened 2023-09-13
[CLI] Support `account transfer` command


jolestar:
1. Support `account transfer --sender 0xa --to 0xb -c 0xabc::abc::ABC` command to transfer ABC coin to `0xb` 
2. Add test case to `testsuit`.

### Refs
1. #793

PR #1658, opened 2024-05-07
[CLI] Support account transfer command

vegetabledogdog:
```
rooch account transfer --help
Transfer coin

Usage: rooch account transfer [OPTIONS] --to <TO> --coin-type <COIN_TYPE> --amount <AMOUNT>

Options:
  -t, --to <TO>
          The account's address to receive coin, if account not exist, auto create account
  -c, --coin-type <COIN_TYPE>
          Struct name as `<ADDRESS>::<MODULE_ID>::<STRUCT_NAME><TypeParam>` Example: `0x3::gas_coin::GasCoin`, `0x123::Coin::Box<0x123::coin_box::FCoin>`
  -a, --amount <AMOUNT>
          The amount of coin to transfer
      --chain-id <CHAIN_ID>
          chain_id  [ROOCH: 20230101] [BITCOIN: 0] [ETHER: 60] [default: 20230101]
      --sender <SENDER>
          Sender account address [default: default]
      --max-gas-amount <MAX_GAS_AMOUNT>
          Custom the transaction's gas limit. [default: 1_000_000_000] [alias: "gas-limit"]
      --authenticator <AUTHENTICATOR>
          Custom the transaction's authenticator format: `auth_validator_id:payload`, auth validator id is u64, payload is hex string example: 123:0x2abc
      --session-key <SESSION_KEY>
          Sign the transaction via session key This option conflicts with `authenticator`
      --config-dir <CONFIG_DIR>
          rooch config path
  -h, --help
          Print help
```
test:
`cargo test -p testsuite --test integration -- --name account`

issue: #792

By the way, there seems like an error in the rooch bitcoin test.
```
Scenario: rooch bitcoin test
✘  Given a bitcoind server for rooch_bitcoin_test
      Step failed:
      Defined: features/cmd.feature:309:7
      Matched: crates/testsuite/tests/integration.rs:106:1
      Step panicked. Captured output: failed to list docker networks: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

### DIFF
--- a/docs/website/components/index/titleHero.tsx
+++ b/docs/website/components/index/titleHero.tsx
@@ -11,17 +11,20 @@ export default function TitleHero({
     <div className="mx-auto px-4 sm:px-6 lg:px-8 pt-60 pb-60">
       {/* <!-- Title --> */}
       <div className="mt-5 max-w-3xl text-center mx-auto">
-        <p className="block font-bold text-gray-800 text-4xl md:text-5xl lg:text-6xl dark:text-gray-200">
+        <p className="block font-bold text-gray-800 text-4xl md:text-5xl lg:text-5xl dark:text-gray-200">
           {title}
         </p>
-        <p className="block pt-2 font-bold text-gray-800 text-2xl md:text-3xl lg:text-4xl dark:text-gray-200">
-          {slogan}
-        </p>
+          {slogan ?
+              <p className="block pt-2 font-bold text-gray-800 text-2xl md:text-3xl lg:text-4xl dark:text-gray-200">
+                  {slogan}
+              </p>
+              : ''
+          }
       </div>
-      {/* <!-- End Title --> */}
+        {/* <!-- End Title --> */}
 
-      <div className="mt-5 max-w-2xl text-center mx-auto">
-        <p className="text-lg text-gray-600 dark:text-gray-400">
+        <div className="mt-5 max-w-2xl text-center mx-auto">
+            <p className="text-lg text-gray-600 dark:text-gray-400">
           {description}
         </p>
       </div>
--- a/docs/website/pages/index.en-US.mdx
+++ b/docs/website/pages/index.en-US.mdx
@@ -7,8 +7,8 @@ searchable: false
 import TitleHero from '../components/index/titleHero'
 
 <TitleHero
-  title="Unlocking"
-  slogan="Infinite Utility for the Bitcoin Economy"
+  title="Unlocking Infinite Utility for the Bitcoin Economy"
+  slogan=""
   description="Rooch Network is the Application Layer of Bitcoin, an Asset-Centric infrastructure that provides indexing, trading, and enables application scenarios for all Bitcoin assets"
   button="Join the community"
   buttonHref="/learn/introduction"
--- a/docs/website/pages/index.zh-CN.mdx
+++ b/docs/website/pages/index.zh-CN.mdx
@@ -7,8 +7,8 @@ searchable: false
 import TitleHero from '../components/index/titleHero'
 
 <TitleHero
-  title="开启"
-  slogan="比特币经济的无限可能"
+  title="开启比特币经济的无限可能"
+  slogan=""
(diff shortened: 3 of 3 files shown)


crates/rooch/src/commands/account/commands/transfer.rs
```diff
@@ -0,0 +1,138 @@
...
+    #[clap(short = 'c', long, value_parser=ParsedStructType::parse)]
+    coin_type: ParsedStructType,
+
+    /// The amount of coin to transfer
+    #[clap(short = 'a', long)]
+    amount: U256,
+
+    /// chain_id  [ROOCH: 20230101] [BITCOIN: 0] [ETHER: 60]
+    #[clap(long, default_value_t = multichain_id::ROOCH)]
+    chain_id: u64,
```
jolestar, line 39: Should we use the `multi_chain_id` here? avoid to conflict with `chain_id(local/dev/test/main)`

jolestar: Or, this command only supports rooch addresses. If we want to support Bitcoin, we should provide a new command that accepts the Bitcoin address as an argument.

jolestar:
rooch_bitcoin_test requires the docker cli.

crates/rooch/src/commands/account/commands/transfer.rs
vegetabledogdog: I guess it will generate a different rooch address from bitcoin or eth address. But I don't know what the rooch address is. The `balance` command only supports rooch addresses.

I did tests on my local server.
For example, to transfer coin on the Bitcoin chain:
`rooch account transfer --to bc1qxy2kgdygjrsqtzq2n0yrf2493p83kkfjhx0wlh -c 0xe952d2f26b773a791ef03604bd323184ec2587d9dd67297c1c598f8252d35256::fixed_supply_coin::FSC -a 1000000 --chain-id 0`
It was successful, because when I run `rooch account balance`, its balance is correctly reduced by 1000000.

But, it can't get the receiver's balance.
```
rooch account balance -a bc1qxy2kgdygjrsqtzq2n0yrf2493p83kkfjhx0wlh
Error: Unbound named address: 'bc1qxy2kgdygjrsqtzq2n0yrf2493p83kkfjhx0wlh'
```

When I conduct tests on the Ethereum chain, the `balance` command incorrectly identifies the Ethereum address as a rooch address.

vegetabledogdog: I think there is an error.
Here is the transfer logic in `transfer.move`.
```
public entry fun transfer_coin_to_multichain_address<CoinType: key + store>(

from: &signer,
multichain_id: u64,
raw_address: vector<u8>,
amount: u256,
) {
let maddress = multichain_address::new(multichain_id, raw_address);
let to = address_mapping::resolve(maddress);
assert!(option::is_some(&to), ErrorAccountNotExists);

let to = option::extract(&mut to);
if(!account::exists_at(to)) {
account_entry::create_account(to);
};
account_coin_store::transfer<CoinType>(from, to, amount)
}
```

 I'd like to change this.
```
-let to = address_mapping::resolve(maddress);
+let to = address_mapping::resolve_or_generate(maddress);
```
This whill generate a new rooch address using black2b256 based on the Multichainaddress.

vegetabledogdog: But I don't know why the original code works, so I didn't change the code.

vegetabledogdog: > Should we use the `multi_chain_id` here? avoid to conflict with `chain_id(local/dev/test/main)`

There won't be a conflict.

jolestar: The account create logic changed in #1624

We could auto-detect the chain ID via the address string. The address string formats are different.

jolestar: We can remove auto-create for account, in the future, we need to deprecated the `CoinStores` in account resource, and then we can support transfer coin to no exists account.

PR #1662, opened 2024-05-09
feat(rooch.statedb): improve import utxo performance by adjusting rocksdb and parallelism

popcnt1:
## Summary

1. utxo importer:
     1.  two thread model, one produce update set batch, one apply updates to smt and state db. 
     2.  print more detailed info
2. rocksdb config:
     1. 1KiB as default block size: 12% improvement may caused by faster block read (filter and index block cache work well with 1KiB block size)
     2. 512MB memtable for buffering enough writes. Increase WAL size and L1 size according to the L0 size changes caused by larger memtable
     3. in pratice, we should use more than 16GiB block cache size
     4. smaller row cache size, because the hit rate is so low in random access I/O model
     5. enable stats by default, because no significant degradation. maybe 1-2%
     6. disable compression on L0 & L1

it's a good start for rocksdb optimization, because it's the first time to have enough data


moveos/moveos-config/src/store_config.rs
```diff
@@ -69,14 +69,14 @@ impl Default for RocksdbConfig {
...
-            max_total_wal_size: 1u64 << 30,
+            max_total_wal_size: 1u64 << 31,
             bytes_per_sync: 1u64 << 20,
             max_background_jobs: 6, // rocksdb will decide compaction threads and flush threads based on this number
             wal_bytes_per_sync: 1u64 << 20,
-            row_cache_size: 1u64 << 30,
-            max_write_buffer_numer: 5,
-            block_cache_size: 1u64 << 32,
-            block_size: 1u64 << 12,
+            row_cache_size: 1u64 << 27,
```
baichuan3, line 76: Do we need to distinguish the DB configuration corresponding to the imported data and the normal server?

moveos/raw-store/src/rocks/mod.rs
```diff
@@ -138,9 +141,28 @@ impl RocksDB {
             column_families.iter().map(|cf_name| {
                 let mut cf_opts = Options::default();
                 cf_opts.set_level_compaction_dynamic_level_bytes(true);
-                cf_opts.set_compression_type(rocksdb::DBCompressionType::Lz4);
-                cf_opts.set_bottommost_compression_type(rocksdb::DBCompressionType::Zstd); // zstd for bottommost level, saving space
+                cf_opts.set_compression_per_level(&[
```
baichuan3, line 144: the same as above

baichuan3 (approved):

PR #1670, opened 2024-05-13
Show display fields of object states from indexer

pause125:
## Summary

1. Unify `QueryOptions` parameter for `query_object_states`, `query_transactions`, `query_events`, `query_field_states`. This may break clients which called these apis.
2. `query_object_states` support return display fields of objects.

@wow-sven

### DIFF
--- a/crates/rooch-indexer/src/indexer_reader.rs
+++ b/crates/rooch-indexer/src/indexer_reader.rs
@@ -375,7 +375,11 @@ impl IndexerReader {
                 format!("{STATE_OWNER_STR} = \"{}\"", owner.to_hex_literal())
             }
             ObjectStateFilter::ObjectId(object_ids) => {
-                let object_ids_str = object_ids.into_iter().map(|obj_id| format!("\"{}\"", obj_id)).collect::<String>().join(",");
+                let object_ids_str = object_ids
+                    .into_iter()
+                    .map(|obj_id| format!("\"{}\"", obj_id))
+                    .collect::<Vec<_>>()
+                    .join(",");
                 format!("{OBJECT_ID_STR} IN ({object_ids_str})")
             }
         };
--- a/crates/rooch-indexer/src/indexer_reader.rs
+++ b/crates/rooch-indexer/src/indexer_reader.rs
@@ -374,8 +374,9 @@ impl IndexerReader {
             ObjectStateFilter::Owner(owner) => {
                 format!("{STATE_OWNER_STR} = \"{}\"", owner.to_hex_literal())
             }
-            ObjectStateFilter::ObjectId(object_id) => {
-                format!("{OBJECT_ID_STR} = \"{}\"", object_id)
+            ObjectStateFilter::ObjectId(object_ids) => {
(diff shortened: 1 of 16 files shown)


moveos/moveos-types/src/moveos_std/display.rs
```diff
@@ -159,12 +172,14 @@ fn parse_template(template: &str, move_value: &AnnotatedMoveValue) -> Result<Str
...
 pub struct RawDisplay {
-    pub fields: SimpleMap<MoveString, MoveString>,
+    pub sample_map: SimpleMap<MoveString, MoveString>,
 }
 
 impl RawDisplay {
     /// Render the display with given MoveStruct instance.
     pub fn render(&self, annotated_obj: &AnnotatedMoveValue) -> BTreeMap<String, String> {
+        println!(">>>>>> AnnotatedMoveValue: {:?}", annotated_obj);
+        println!(">>>>> template: {:?}", self.to_btree_map());
```
jolestar, line 182: remove the debug println, or change to `debug!` or `trace!`

jolestar (commented):
We need to release a new version after merge this PR.

crates/rooch-indexer/src/indexer_reader.rs
```diff
@@ -374,8 +374,13 @@ impl IndexerReader {
...
             }
-            ObjectStateFilter::ObjectId(object_id) => {
-                format!("{OBJECT_ID_STR} = \"{}\"", object_id)
+            ObjectStateFilter::ObjectId(object_ids) => {
+                let object_ids_str = object_ids
+                    .into_iter()
+                    .map(|obj_id| format!("\"{}\"", obj_id))
+                    .collect::<Vec<_>>()
+                    .join(",");
+                format!("{OBJECT_ID_STR} IN ({object_ids_str})")
```
baichuan3, line 383: Need limit the length of object_ids to avoid malicious calls

jolestar: Should we remove the `ObjectStateFilter::ObjectId`? Because if the caller knows the ObjectIds, he should call the `getStates` API.

crates/rooch-rpc-server/src/server/rooch_server.rs
```diff
@@ -642,19 +645,49 @@ impl RoochAPIServer for RoochServer {
             .map(|m| m.object_id.clone())
             .collect::<Vec<_>>();
         let access_path = AccessPath::objects(object_ids.clone());
-        let mut data = self
-            .rpc_service
-            .get_annotated_states(access_path)
-            .await?
+        let annotated_states = self.rpc_service.get_annotated_states(access_path).await?;
```
jolestar, line 648: If the `ObjectStateFilterView` is `ObjectStateFilterView::ObjectID,` we should directly build the `AccessPath,` and do not need to call the `query_object_states.`

baichuan3: indeed

baichuan3: The Object returned from the states API does not have a timestamp. The timestamp can be returned from the Transaction API. Consider adding a timestamp to `Transaction`?

PR #1690, opened 2024-05-16
[framework] Fix session key create session key bug

### LINKED ISSUE #1625: [bug] SessionKey can create SessionKey
https://github.com/rooch-network/rooch/blob/417f226614f56caa4e8657a067d9bf952567530f/frameworks/rooch-framework/sources/session_key.move#L116-L117

The assert ensures users can not use a SessionKey to create a new SessionKey, but it does not work. 

## Reason

https://github.com/rooch-network/rooch/blob/417f226614f56caa4e8657a067d9bf952567530f/crates/rooch-types/src/framework/auth_validator.rs#L152-L164

the `transaction_validtor` of above should `transaction_validator`.

1. All the logic with `TxValidateResult` should be checked.
2. Write a test to ensure the assert.

jolestar:
## Summary

Fix session key create session key bug

- Closes #1625

### DIFF
--- a/crates/rooch-framework-tests/src/tests/transaction_validator_tests.rs
+++ b/crates/rooch-framework-tests/src/tests/transaction_validator_tests.rs
@@ -6,7 +6,7 @@ use move_core_types::account_address::AccountAddress;
 use move_core_types::ident_str;
 use move_core_types::language_storage::ModuleId;
 use move_core_types::value::MoveValue;
-use move_core_types::vm_status::{AbortLocation, VMStatus};
+use move_core_types::vm_status::{AbortLocation, KeptVMStatus, VMStatus};
 use moveos_types::module_binding::MoveFunctionCaller;
 use moveos_types::move_std::ascii::MoveAsciiString;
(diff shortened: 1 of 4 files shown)


crates/rooch-framework-tests/src/tests/transaction_validator_tests.rs
```diff
@@ -113,17 +108,55 @@ fn test_session_key_rooch() {
...
+        KeptVMStatus::MoveAbort(l, code) => {
+            match l {
+                AbortLocation::Module(module_id) => {
+                    assert_eq!(
+                        module_id,
+                        SessionKeyModule::module_id(),
+                        "expect session key module"
+                    );
+                }
+                _ => panic!("expect move abort in module"),
```
baichuan3, line 139: Directly panic? Or throw an Error?

jolestar: The test case no Result

baichuan3 (approved):

Issue #1669, opened 2024-05-13
[Framework] Deprecated the CoinStores in the account resource.


jolestar:
We currently record CoinStores in the account resource to retrieve the account's coin list. We can use the indexer to query the account's CoinStore and deprecate the CoinStores in the account resource.

### Relative issue
1. #1624
2. #1664  #1658

PR #1703, opened 2024-05-18
[genesis] Fix module store size bug

### LINKED ISSUE #1565: [object] ModuleStore object field size is incorrect after init genesis
Because the genesis transaction directly writes the module to storage, does not through the `move_module.move`, the  ModuleStore object field size does not change.

jolestar:
## Summary

1. Fix #1565 to support framework upgrade.
2. Auto-create account when transfer; after #1669, we can remove the auto-create logic.


jolestar:
I merge this PR first. @pause125  @baichuan3

Issue #1495, opened 2024-03-29
Random test failed


jolestar:
malloc(): unaligned tcache chunk detected
error: test failed, to rerun pass `-p rooch-framework-tests --test tests`

Caused by:
  process didn't exit successfully: `/runner/_work/rooch/rooch/target/debug/deps/tests-b2307daf3ae0fc05 run` (signal: 6, SIGABRT: process abort signal)
Error: Process completed with exit code 101.

https://github.com/rooch-network/rooch/actions/runs/8475914508/job/23226655831?pr=1493

jolestar:
realloc(): invalid next size
error: test failed, to rerun pass `-p rooch-framework-tests --test tests`

Caused by:
  process didn't exit successfully: `/runner/_work/rooch/rooch/target/release/deps/tests-c6130266fac3d7b7 run` (signal: 6, SIGABRT: process abort signal)

jolestar:
task 3 'run'. lines 56-69:
status EXECUTED
Error: error: the function "0x3::account_authentication::rotate_authentication_key" may not exists.
corrupted size vs. prev_size in fastbins

    ┌─ /runner/_work/rooch/rooch/frameworks/rooch-framework/sources/account_authentication.move:6:1
    │  
  6 │ ╭ module rooch_framework::account_authentication {
  7 │ │    
  8 │ │    use std::option::{Self, Option};
  9 │ │    use std::signer;
    · │
189 │ │    }
190 │ │ }
    │ ╰─^




Run with `env UB=1` (or `env UPDATE_BASELINE=1`) to save the current output as the new expected output


failures:
    run_test::cases/table/basic.move

test result: FAILED. 29 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 6.57s

error: test failed, to rerun pass `-p rooch-framework-tests --test tests`

Caused by:
  process didn't exit successfully: `/runner/_work/rooch/rooch/target/release/deps/tests-0f038c5a8a8d34ff run` (signal: 6, SIGABRT: process abort signal)


https://github.com/rooch-network/rooch/actions/runs/8787259029/job/24112048074?pr=1601

jolestar:
thread '<unnamed>' panicked at /rustc/7cf61ebde7b22796c69757901dd346d0fe70bd97/library/alloc/src/collections/btree/node.rs:1199:35:
range end index 18446744073709551615 out of range for slice of length 11
[...]

PR #1710, opened 2024-05-20
Fix the random test failed problem.

steelgeek091:
Closes [https://github.com/rooch-network/rooch/issues/1495](https://github.com/rooch-network/rooch/issues/1495)

### DIFF
--- a/moveos/moveos-verifier/src/metadata.rs
+++ b/moveos/moveos-verifier/src/metadata.rs
@@ -24,17 +24,20 @@ use once_cell::sync::Lazy;
 use serde::{Deserialize, Serialize};
 use std::collections::BTreeMap;
 use std::ops::Deref;
+use std::sync::{Arc, RwLock};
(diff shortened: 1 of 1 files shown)


moveos/moveos-verifier/src/metadata.rs
```diff
@@ -321,6 +324,8 @@ fn get_type_name_indices(
 
                     unsafe {
                         GLOBAL_PRIVATE_GENERICS
+                            .write()
```
jolestar, line 327: Can remove the `unsafe{}`

jolestar:
I merge this PR first.

jolestar (approved):

moveos/moveos-verifier/src/metadata.rs
steelgeek091: > Can remove the `unsafe{}`

Got it，I will remove it in a later PR.

Issue #1465, opened 2024-03-21
[moveos_std] Support all evm precompiled functions


jolestar:
Add an `evm.move` module, include all evm precompiled functions

https://www.evm.codes/precompiled?fork=cancun

<img width="1542" alt="Screen Shot 2024-03-21 at 11 19 26" src="https://github.com/rooch-network/rooch/assets/77268/ca5e5efc-ed57-4135-86f3-16a62c0429a9">

0x02,0x06,0x08 need to be implemented first.

0x02 already include in `std::hash`

PR #1739, opened 2024-05-26
[moveos_std] Add evm precompiled functions

vegetabledogdog:
implement `0x02 sha256`, `0x06 ecAdd`, `0x08 ecPairing`

- #1465

### DIFF
--- a/frameworks/moveos-stdlib/sources/evm.move
+++ b/frameworks/moveos-stdlib/sources/evm.move
@@ -1,0 +1,104 @@
+// Copyright (c) RoochNetwork
+// SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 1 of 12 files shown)


frameworks/moveos-stdlib/sources/evm.move
```diff
@@ -0,0 +1,104 @@
+// Copyright (c) RoochNetwork
+// SPDX-License-Identifier: Apache-2.0
+
+// Evm Precompiled Contracts https://www.evm.codes/precompiled?fork=cancun
+module moveos_std::evm {
+
+    const E_EC_ADD_FAILED: u64 = 6;
```
jolestar, line 7: Use `ErrorECAddFailed` to keep consistency with other modules.

jolestar (approved):

PR #1765, opened 2024-05-29
Refactor some parts of the verifier and add more mvir tests.

steelgeek091:


moveos/moveos-verifier/src/build.rs
```diff
@@ -429,6 +429,24 @@ pub fn compile_and_inject_metadata(
 
             rooch_metadata.data_struct_map = data_structs_map;
         }
+
+        if metadata_type == "data_struct_func" {
```
jolestar, line 433: Define some consts?

jolestar (approved):

moveos/moveos-verifier/src/build.rs
steelgeek091: OK, will fix it in a later PR.

PR #1852, opened 2024-06-07
basic ord import logic

popcnt1:
## Summary

done:
1. ord obj store
2. ord parent

todo:
1. ord seal for utxo import
2. update start info
3. modify struct(waiting changes in move side)


popcnt1:
move to new branch

Issue #1803, opened 2024-06-03
[CLI] Refactor the CLI result


jolestar:
Refactor the CLI result support output json to every command.

Currently, we support output json in some commands, such as `account list`. But we sometimes need to parse the command output, such as in the integration test(`cmd.feature`).

- [x] #1895

feliciss:
May I handle this issue?

jolestar:
This is finished by adding the `execute_serialized` function to the CommandAction. 

https://github.com/rooch-network/rooch/blob/2910febd4b9b6717e229a0bfe9b46e8bbe746b54/crates/rooch/src/cli_types.rs#L28-L40

If you have a better solution, you can promote a refactor suggestion. @feliciss

PR #1854, opened 2024-06-08
[CLI] `rooch account` supports output in json

vegetabledogdog:
-  #1803 

`rooch account` supports JSON output.
`rooch account switch` and `rooch account nullify` support bitcoin address as an argument.
Add a test case that uses a Bitcoin address and fix a transfer task in `cmd.feature`.


jolestar (approved):

crates/rooch/src/commands/account/commands/list.rs
```diff
@@ -50,15 +51,18 @@ impl LocalAccountView {
     }
 }
 
-#[derive(Serialize, Deserialize, Debug)]
+#[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct AccountView {
+    #[serde(flatten)]
```
jolestar, line 56: If we `flatten` the `local_account`, the json path in `cmd.feature` must also be updated.

PR #1868, opened 2024-06-11
[CLI] `rooch move` success json output

vegetabledogdog:
-  #1803
```
{
  "Result": "Success"
}
```
Just enable "success" in the JSON output.


jolestar:
1. All commands should add a `--json` flag.
2. Only output json if the `--json` flag is set.

vegetabledogdog:
1. Make `rooch move` commands return `RoochResult` and implement `CommandAction` for them.
2. It still needs improvement, as most of them simply return "Success" in JSON output.

jolestar (approved):

PR #1917, opened 2024-06-18
[test] Add testcase for testnet 2821527 block and refactor the real block tests

jolestar:
## Summary

Try to reproduce issue #1897, but can not reproduce it in the test case.


jolestar:
This PR only change the test case, I merge it first.

PR #1997, opened 2024-06-24
feat: check the validity of bitseed protocol splits and merges

yubing744:
1、Check the validity of bitseed protocol split
2、Check the validity of bitseed protocol merge


yubing744:
@jolestar I will close this PR

PR #2129, opened 2024-07-10
[cli] rooch session-key list

vegetabledogdog:
```
rooch session-key list -h
List all session keys by address

Usage: rooch session-key list [OPTIONS]

Options:
  -a, --address <ADDRESS>        The account's address to list session keys, if absent, show the default active account [default: default]
      --password <PASSWORD>      The key store password
      --config-dir <CONFIG_DIR>  rooch config path
  -h, --help                     Print help
```

### DIFF
--- a/crates/rooch/src/commands/session_key/commands/list.rs
+++ b/crates/rooch/src/commands/session_key/commands/list.rs
@@ -1,0 +1,92 @@
+// Copyright (c) RoochNetwork
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::cli_types::{CommandAction, WalletContextOptions};
+use async_trait::async_trait;
+use clap::Parser;
+use move_core_types::identifier::Identifier;
+use moveos_types::move_types::FunctionId;
+use moveos_types::moveos_std::object::ObjectID;
+use moveos_types::transaction::FunctionCall;
+use rooch_rpc_api::jsonrpc_types::{
+    AnnotatedFunctionResultView, AnnotatedMoveValueView, StateOptions, StatePageView,
+};
+use rooch_types::address::ParsedAddress;
+use rooch_types::error::{RoochError, RoochResult};
+use std::collections::BTreeMap;
+
+/// List all session keys by address
+#[derive(Debug, Parser)]
+pub struct ListCommand {
+    #[clap(short = 'a', long = "address", value_parser=ParsedAddress::parse, default_value = "default")]
+    /// The account's address to list session keys, if absent, show the default active account.
+    address: ParsedAddress,
+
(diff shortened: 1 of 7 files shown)


crates/testsuite/features/cmd.feature
```diff
@@ -54,7 +54,8 @@ Feature: Rooch CLI integration tests
 
       # session key
       Then cmd: "session-key create  --app-name test --app-url https:://test.rooch.network --scope 0x3::empty::empty"
-      Then cmd: "move run --function 0x3::empty::empty  --session-key {{$.session-key[-1].authentication_key}}"
+      Then cmd: "session-key list"
```
jolestar, line 57: Add an assert to ensure the command is executed successfully.

jolestar (approved):

crates/testsuite/features/cmd.feature
vegetabledogdog: The error message is like this.
```
rooch session-key list
Error: Session key not found
```
assert this
Then assert: "'{{$.session-key[-1]}}' not_contains Error"

frameworks/rooch-framework/sources/session_key.move
```diff
@@ -260,6 +261,14 @@ module rooch_framework::session_key {
         remove_session_key(sender, authentication_key);
     }
 
+    public fun handle(account_address: address) : Option<ObjectID> {
```
jolestar, line 264: Rename to `get_session_keys_handle`? More clear.

jolestar: renamed in #2133

crates/rooch/src/commands/session_key/commands/list.rs
```diff
@@ -0,0 +1,92 @@
...
+        let address_addr = self.address.into_account_address(&mapping)?;
+
+        let function_id = "0x3::session_key::handle".parse::<FunctionId>()?;
+        let function_call = FunctionCall::new(function_id, vec![], vec![address_addr.to_vec()]);
+        let client = context.get_client().await?;
+        let view_result = client
+            .rooch
+            .execute_view_function(function_call)
+            .await
+            .map_err(|e| RoochError::ViewFunctionError(e.to_string()))?;
```
jolestar, line 43: This function call can be defined in the mod https://github.com/rooch-network/rooch/blob/main/crates/rooch-types/src/framework/session_key.rs

Then call it via client, like

https://github.com/rooch-network/rooch/blob/3197fd97aedc9d1ac76e974d0e029d1c47529f15/crates/rooch/src/commands/session_key/commands/create.rs#L102-L104

PR #2133, opened 2024-07-10
[bitcoin] Refactor ord sequence_number to ID

jolestar:
## Summary

Remove the `TableVec<InscriptionID>` from `InscriptionStore` and record `sequence_number` to `InscriptionID` mapping in the InscriptionStore dynamic field.

Simply the data importing.

@popcnt1 

cc @yubing744


popcnt1 (approved):
InscriptionStore in rooch-types/src/bitcoin/ord.rs need to be updated too

PR #2170, opened 2024-07-13
ci: accelerate github workflow

popcnt1:
## Summary

1. remove redundant cargo build
2. remove redundant dep install (dev_setup.sh) by 
3. add sccache 

@Xuanwo now sure the usage of sccache, we have our own runner machine, will it cached on this machine's `~/.cache/sccache/`? So in the next time, workflow could reuse the cached files?


Xuanwo:
> @Xuanwo now sure the usage of sccache, we have our own runner machine, will it cached on this machine's `~/.cache/sccache/`? 

Yes, sccache default to cache data locally.

> So in the next time, workflow could reuse the cached files?

I'm not sure how your runner is configured and whether the data stored locally will be purged between different jobs.

popcnt1:
> > @Xuanwo now sure the usage of sccache, we have our own runner machine, will it cached on this machine's `~/.cache/sccache/`?
> 
> Yes, sccache default to cache data locally.
> 
> > So in the next time, workflow could reuse the cached files?
> 
> I'm not sure how your runner is configured and whether the data stored locally will be purged between different jobs.

@yourmoonlight could u answer this question for me? About the data store between jobs in our runner

yourmoonlight:
> ## Summary
> 1. remove redundant cargo build
> 2. remove redundant dep install (dev_setup.sh) by
> 3. add sccache
> 
> @Xuanwo now sure the usage of sccache, we have our own runner machine, will it cached on this machine's `~/.cache/sccache/`? So in the next time, workflow could reuse the cached files?

The current GitHub runner is self-hosted using a k8s cluster. If the cache is set to local, different jobs cannot reuse this cache. 
IMHO can we set it to a caching mode similar to node.js and cache it in GitHub? @popcnt1 @Xuanwo

yourmoonlight:
> > ## Summary
> > 
> > 1. remove redundant cargo build
> > 2. remove redundant dep install (dev_setup.sh) by
> > 3. add sccache
> > 
> > @Xuanwo now sure the usage of sccache, we have our own runner machine, will it cached on this machine's `~/.cache/sccache/`? So in the next time, workflow could reuse the cached files?
> 
> The current GitHub runner is self-hosted using a k8s cluster. If the cache is set to local, different jobs cannot reuse this cache. IMHO can we set it to a caching mode similar to node.js and cache it in GitHub? @popcnt1 @Xuanwo

Or, consider using external storage services (such as the object storage services) to store cache? Any advices?

Xuanwo:
> Or, consider using external storage services (such as the object storage services) to store cache? Any advices?

Yes, I recommend using object storage services like S3 or Azure Blob Storage. You can directly configure the secrets in the runner.

popcnt1:
#2320 
#2321

PR #2190, opened 2024-07-15
update sdk-kit: use enums for SupportChain and SupportWallet

birchwork:
## Summary

Summary about this PR

- Closes #issue


wow-sven:
We can leave this place alone for now

PR #2134, opened 2024-07-10
[gh-1888] add account export module.

feliciss:
## Summary

Add `account export` module and resolve #1888.

Testing various scenarios:

```zsh
> rooch account export       
Export succeeded with the mnemonic phrase [spot demise medal priority old rack lobster cover gasp just warfare essay]

> rooch account export -a rooch16469un6dceflr8026v0afv4t525d3y5s3u3g45r2kxxyj08twzdq4vtufw
Address to be exported: rooch16469un6dceflr8026v0afv4t525d3y5s3u3g45r2kxxyj08twzdq4vtufw(0xd5745e4f4dc653f19dead31fd4b2aba2a8d892908f228ad06ab18c493ceb709a)
Export succeeded with the encoded private key [roochsecretkey1qxmxtzkhftwa579390cj7v09e6xmsl6r4tgdy9q3rupwazaw065dj58kc75]

> rooch account export -a default
Address to be exported: rooch1xxzwqu47gsxf2s3max43t7tzneunljr7vpdy6d436tnw0e7zsp4q32jwdc(0x3184e072be440c95423be9ab15f9629e793fc87e605a4d36b1d2e6e7e7c2806a)
Export succeeded with the encoded private key [roochsecretkey1q8caw0k0m7uuydqectm9cums2ecz6fd4pccnut8vlkwwsyww6kxmxleuz8v]

> rooch account export -a ''     
Export succeeded with the mnemonic phrase [spot demise medal priority old rack lobster cover gasp just warfare essay]

> rooch account export -a ""
Export succeeded with the mnemonic phrase [spot demise medal priority old rack lobster cover gasp just warfare essay]

> rooch account export -a wwww
Invalid arguments: Invalid Rooch address String: Unbound named address: 'wwww'

> rooch account export -a rooch3xxzwqu47gsxf2s3max43t7tzneunljr7vpdy6d436tnw0e7zsp4q32jwdc
error: invalid value 'rooch3xxzwqu47gsxf2s3max43t7tzneunljr7vpdy6d436tnw0e7zsp4q32jwdc' for '--address <ADDRESS>': parsing failed

> rooch account export -a rooch1xwzwqu47gsxf2s3max43t7tzneunljr7vpdy6d436tnw0e7zsp4q32jwdc
error: invalid value 'rooch1xwzwqu47gsxf2s3max43t7tzneunljr7vpdy6d436tnw0e7zsp4q32jwdc' for '--address <ADDRESS>': no valid bech32 or bech32m checksum

[...]

PR #2200, opened 2024-07-16
[gh-1094] add account import module.

### LINKED ISSUE #1094: [Feature request] Import CLI Account via address and private key
Currently, there is a lack of a command to import accounts, which needs to be added under the `rooch account` command.

feliciss:
## Summary

Add `account import` module as a pair of #2134.

- Closes #1094

### DIFF
--- a/crates/rooch-key/src/keystore/account_keystore.rs
+++ b/crates/rooch-key/src/keystore/account_keystore.rs
@@ -2,9 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use super::types::LocalAccount;
-use crate::key_derive::{generate_derivation_path, generate_new_key_pair, ROOCH_SECRET_KEY_PREFIX};
-use bitcoin::bech32::{encode, Bech32, Hrp};
-use rooch_types::crypto::SignatureScheme;
+use crate::key_derive::{generate_derivation_path, generate_new_key_pair};
 use rooch_types::framework::session_key::SessionKey;
 use rooch_types::key_struct::{MnemonicData, MnemonicResult};
 use rooch_types::{
@@ -81,17 +79,16 @@ pub trait AccountKeystore {
         Ok(mnemonic_phrase)
     }
 
-    fn export_private_key(&mut self, sk_bytes: &[u8]) -> Result<String, anyhow::Error> {
-        // get 33 bytes flag and secret key
-        let mut priv_key_bytes = Vec::with_capacity(sk_bytes.len() + 1);
-        // supports secp256k1 signature scheme
-        priv_key_bytes.push(SignatureScheme::Secp256k1.flag());
-        priv_key_bytes.extend_from_slice(sk_bytes);
-        // init `roochsecretkey` as HRP
-        let hrp = Hrp::parse(ROOCH_SECRET_KEY_PREFIX)?;
-        // encode hrp and 33 bytes private key using bech32 method
-        let bech32_encoded = encode::<Bech32>(hrp, &priv_key_bytes)?;
-        Ok(bech32_encoded)
+    fn import_external_account(
+        &mut self,
+        address: RoochAddress,
+        kp: RoochKeyPair,
+        password: Option<String>,
+    ) -> Result<(), anyhow::Error> {
+        let private_key_encryption = EncryptionData::encrypt_with_type(&kp, password)?;
+        // TODO: whether to add address to mnemonic data? It's either associated with the same mnemonic or not. There needs a check if the imported account is from the same mnemonic phrase.
+        self.add_address_encryption_data_to_keys(address, private_key_encryption)?;
+        Ok(())
     }
 
     fn get_accounts(&self, password: Option<String>) -> Result<Vec<LocalAccount>, anyhow::Error>;
--- a/crates/rooch-key/src/keystore/account_keystore.rs
+++ b/crates/rooch-key/src/keystore/account_keystore.rs
@@ -86,7 +86,6 @@ pub trait AccountKeystore {
         password: Option<String>,
     ) -> Result<(), anyhow::Error> {
         let private_key_encryption = EncryptionData::encrypt_with_type(&kp, password)?;
-        // TODO: whether to add address to mnemonic data? It's either associated with the same mnemonic or not. There needs a check if the imported account is from the same mnemonic phrase.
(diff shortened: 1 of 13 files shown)


crates/rooch-types/src/crypto.rs
```diff
@@ -124,6 +126,18 @@ impl RoochKeyPair {
             RoochKeyPair::Secp256k1(kp) => RoochKeyPair::Secp256k1(kp.copy()),
         }
     }
+
+    // Export Private Key method exports a private key in bech32 format
+    pub fn export_private_key(&self) -> Result<String, EncodeError> {
+        // get 33 bytes flag and secret key
+        let mut priv_key_bytes = Vec::with_capacity(self.private().len() + 1);
+        // supports secp256k1 signature scheme
+        priv_key_bytes.push(SignatureScheme::Secp256k1.flag());
```
jolestar, line 135: The keypair can be Ed25519; use `self.public_key.flag()` to get the flag.

crates/rooch/src/commands/account/commands/import.rs
```diff
@@ -0,0 +1,65 @@
...
+#[async_trait]
+impl CommandAction<Option<String>> for ImportCommand {
+    async fn execute(self) -> RoochResult<Option<String>> {
+        let mut context = self.context_options.build_require_password()?;
+        let password = context.get_password();
+        let mapping = context.address_mapping();
+        let rooch_address: RoochAddress =
+            self.address.into_rooch_address(&mapping).map_err(|e| {
+                RoochError::CommandArgumentError(format!("Invalid Rooch address String: {}", e))
+            })?;
```
jolestar, line 44: We should generate the rooch_address from the KeyPair.

crates/rooch-key/src/keystore/account_keystore.rs
```diff
@@ -81,17 +79,16 @@ pub trait AccountKeystore {
...
-        let bech32_encoded = encode::<Bech32>(hrp, &priv_key_bytes)?;
-        Ok(bech32_encoded)
+    fn import_external_account(
+        &mut self,
+        address: RoochAddress,
+        kp: RoochKeyPair,
+        password: Option<String>,
+    ) -> Result<(), anyhow::Error> {
+        let private_key_encryption = EncryptionData::encrypt_with_type(&kp, password)?;
+        // TODO: whether to add address to mnemonic data? It's either associated with the same mnemonic or not. There needs a check if the imported account is from the same mnemonic phrase.
```
jolestar, line 89: The address in mnemonic data is derived from the mnemonic. So we should not add the address to `mnemonic data`.

The address not in mnemonic data is an external account address.

crates/rooch-types/src/crypto.rs
```diff
@@ -109,7 +111,7 @@ impl RoochKeyPair {
     pub fn private(&self) -> &[u8] {
         match self {
             RoochKeyPair::Ed25519(kp) => kp.as_bytes(),
-            RoochKeyPair::Secp256k1(kp) => kp.as_bytes(),
+            RoochKeyPair::Secp256k1(kp) => kp.secret.as_bytes(),
```
jolestar, line 114: Why use `.secret` here? Should we also change the `Ed25519` ?

feliciss: > Why use `.secret` here? Should we also change the `Ed25519` ?

The Ed25519's public key can not be publicly accessible, so it has to be remained to `kp.as_bytes()` but we can change the function name to `kp_bytes` instead of `private`.

feliciss: Use `as_ref()` should be more appropriate to get the private keys.

crates/rooch-types/src/crypto.rs
```diff
@@ -108,8 +110,8 @@ impl RoochKeyPair {
 
     pub fn private(&self) -> &[u8] {
         match self {
-            RoochKeyPair::Ed25519(kp) => kp.as_bytes(),
-            RoochKeyPair::Secp256k1(kp) => kp.as_bytes(),
+            RoochKeyPair::Ed25519(kp) => kp.as_ref(),
+            RoochKeyPair::Secp256k1(kp) => kp.secret.as_ref(),
```
jolestar, line 114: kp.as_bytes() == kp.as_ref() == kp.secret.as_ref()

It seems like we do not need to change this.

crates/testsuite/features/cmd.feature
```diff
@@ -40,6 +40,7 @@ Feature: Rooch CLI integration tests
       Then cmd: "account export -a {{$.account[-1].account0.address}}"
       # use bitcoin_address
       Then cmd: "account nullify -a {{$.account[-1].account0.bitcoin_address}}"
+      # TODO: account import with -a and -k flags
```
jolestar, line 43: We need a test case to import and remove the imported account.

crates/rooch-key/src/keystore/base_keystore.rs
```diff
@@ -150,6 +150,23 @@ impl AccountKeystore for BaseKeyStore {
...
+        let index = match mnemonic_data
+            .addresses
+            .iter()
+            .position(|&target_address| target_address == *address)
+        {
+            Some(index) => index,
+            None => {
+                return Err(anyhow::anyhow!(
+                    "Could not find the address in mnemonic data"
+                ))
```
jolestar, line 166: If we nullify the external account, will it raise an error here?

crates/rooch-types/src/crypto.rs
feliciss: > kp.as_bytes() == kp.as_ref() == kp.secret.as_ref()
> 
> It seems like we do not need to change this.

Yes. You're exactly right on this but when clicking on `as_bytes()` it will go to traits not the in-depth position to get insight of. However, I don't know if it's just me.

jolestar: The IDE can not locate the implemention.

feliciss: > The IDE can not locate the implemention.

I'm using VS-Code-like IDE and it will lead to the Private or Secret Key implementation to get the bytes.

crates/rooch-key/src/keystore/base_keystore.rs
feliciss: Resolved as `Ok(())`.

jolestar (approved):

Issue #2169, opened 2024-07-13
[vm] readonly tx max gas limit


jolestar:
I increased the max gas amount for the read-only tx to `5_000_000_000` in #2167 

To resolve the read-only call error:

```
Relayer bitcoin_relayer error: Failed to get ready pending txs: EXECUTION_FAILURE { status_code: OUT_OF_GAS, sub_status: None, location: 0000000000000000000000000000000000000000000000000000000000000001::string, function_definition: 16, code_offset: 1, message: None }
```

1. Figure out why the `pending_block::get_ready_pending_txs` uses so much gas.
2. Calculate a suitable default value for the read-only tx max gas amount.

PR #2207, opened 2024-07-17
Reduce the gas cost of Call/CallGeneric instructions and related instructions.

steelgeek091:
resolve https://github.com/rooch-network/rooch/issues/2169


jolestar:
Should we merge this PR before v0.7 release? @steelgeek091

PR #2220, opened 2024-07-18
[cli]Line break output after password output

shaflow01:
## Summary

- Add line break after entering password

before:

697e89bafea5616ce0313b873e4f1f1
after:
043ac3760c4fc8bca59254816e6469e

### DIFF
--- a/crates/rooch/src/commands/account/commands/list.rs
+++ b/crates/rooch/src/commands/account/commands/list.rs
@@ -69,9 +69,11 @@ impl CommandAction<Option<AccountsView>> for ListCommand {
         let password = if context.keystore.get_if_password_is_empty() {
             None
         } else {
-            Some(
+            let password = Some(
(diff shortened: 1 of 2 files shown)


crates/rooch/src/commands/account/commands/list.rs
```diff
@@ -69,9 +69,11 @@ impl CommandAction<Option<AccountsView>> for ListCommand {
         let password = if context.keystore.get_if_password_is_empty() {
             None
         } else {
-            Some(
+            let password = Some(
                 prompt_password("Enter the password to create a new key pair:").unwrap_or_default(),
-            )
+            );
+            println!();
```
jolestar, line 75: Please also help to change the `WalletContextOptions` prompt_password.

https://github.com/rooch-network/rooch/blob/8a16bd56a71a01c07677cd593aff602688619430/crates/rooch/src/cli_types.rs#L117-L135

jolestar (approved):

PR #2255, opened 2024-07-23
fix(genesis_utxo): fix owener of RoochToBitcoinAddressMapping

popcnt1:
## Summary

`SYSTEM_OWNER_ADDRESS` has been replaced with `ROOCH_FRAMEWORK_ADDRES`

p.s.

I wonder when I create dynamic_field of RoochToBitcoinAddressMapping, the field's owner will be `SYSTEM_OWNER_ADDRESS` , but it belongs to an object which owner is  `ROOCH_FRAMEWORK_ADDRES`, that's weird @jolestar @baichuan3


baichuan3:
If the object flag is `shared` or `frozen`, the corresponding owner will be `SYSTEM_OWNER_ADDRESS` (0x0). Otherwise, the object is UserOwnedObject, and its owner will be the user address, including the system reserved address (0x0 - 0xa).
,.
RoochToBitcoinAddressMapping is a UserOwnedObject, and its owner is the rooch framework (0x3).

baichuan3 (approved):

jolestar:
> However, `BitcoinUTXOStore` and `InscriptionStore` is `shared`, which should be `UserOwnedObject`, and its owner should be `bitcoin-move`. Otherwise, it means that any user can get its mut reference and modify it. @popcnt1 @jolestar

The shared is ok, only the utxo module can provide the modify function for BitcoinUTXOStore

PR #2258, opened 2024-07-23
[VM] Handle the Discard type of status code from native function execution

steelgeek091:


frameworks/moveos-stdlib/src/natives/helpers.rs
```diff
@@ -58,7 +62,22 @@ where
...
+            if let Err(err) = &result {
+                let status = err.major_status();
+                let vm_status = VMStatus::error(status, Some("".to_string()));
+                match vm_status.keep_or_discard() {
+                    Ok(_) => result,
+                    Err(_) => {
+                        log::error!("Native function execution failed {}", err.to_string());
+                        Ok(NativeResult::err(
+                            InternalGas::new(3000),
+                            E_NATIVE_FUNCTION_EXECUTING_PANIC,
```
steelgeek091, line 74: Return `NativeResult::err` in there appropriate or not?

frameworks/moveos-stdlib/src/natives/helpers.rs
```diff
@@ -58,7 +62,22 @@ where
                     },
                 }
             }
-            result
+            if let Err(err) = &result {
+                let status = err.major_status();
+                let vm_status = VMStatus::error(status, Some("".to_string()));
+                match vm_status.keep_or_discard() {
```
jolestar, line 68:
```suggestion
                let vm_status = VMStatus::error(status, None);
```

frameworks/moveos-stdlib/src/natives/helpers.rs
```diff
@@ -58,7 +62,22 @@ where
...
+                    Ok(_) => result,
+                    Err(_) => {
+                        log::error!("Native function execution failed {}", err.to_string());
+                        Ok(NativeResult::err(
```
jolestar, line 72: We should convert the discard VMError to a keept VMError, not NativeResult::err

steelgeek091: The return type of this function is `PartialVMResult`, but the struct `PartialVMError` does not have the `Location` information in it:

```rust
struct PartialVMError_ {
    major_status: StatusCode,
    sub_status: Option<u64>,
    message: Option<String>,
    exec_state: Option<ExecutionState>,
    indices: Vec<(IndexKind, TableIndex)>,
    offsets: Vec<(FunctionDefinitionIndex, CodeOffset)>,
}
```

jolestar (approved):

PR #2259, opened 2024-07-23
[gh-1117] add math modules.

### LINKED ISSUE #1117: [MoveosStd] Math modules

feliciss:
## Summary

Add move math modules `u8`, `u16`, `u32`, `u64`, `u128`, `u256`.

- Closes #1117


frameworks/move-stdlib/sources/macros.move
```diff
@@ -0,0 +1,122 @@
+// Copyright (c) RoochNetwork
+// SPDX-License-Identifier: Apache-2.0
+
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+/// This module holds shared implementation of macros used in `std`
+module std::macros {
+    public macro fun num_max($x: _, $y: _): _ {
```
baichuan3, line 9: Do macros support arguments of arbitrary types?

feliciss: See the above doc.

> Type inference holes _ on type annotations
> 
> With type directed programming, often you need to annotate a variable declaration or provide type arguments. But, sometimes you really only need to annotate on specific type, but the other types can be inferred. _ will be added to allow that type to still be inferred, even when other parts of the type are annotated. For example
> 
> dynamic_field::borrow_mut<address, Coin<SUI>>(&mut id, owner)
> 
> could be rewritten as
> 
> dynamic_field::borrow_mut<_, Coin<SUI>>(&mut id, owner)
> 
> where the _ would be inferred as address

jolestar:
Rooch move do not support macros yet.

feliciss:
> Rooch move do not support macros yet.

Is there a way to support the 2024 macros version or we back to the one without macros?

feliciss:
> > > Rooch move do not support macros yet.
> > 
> > 
> > Is there a way to support the 2024 macros version or we back to the one without macros?
> 
> Back to the one without macros.

I have submitted the ones without macros.

jolestar (commented):
Needs some testcase.

feliciss:
> Needs some testcase.

Added.

jolestar (approved):

PR #2307, opened 2024-07-29
🔨 Remove Pagination's default property herf

nlyrthiia:
## Summary

CustomPagination Adding a default herf will cause the page to make meaningless jumps as well as add # hash to the url, greatly affecting the user experience.

https://github.com/user-attachments/assets/ec519bcc-0a69-446c-9603-2fd873faf56f

https://github.com/user-attachments/assets/b85b51f5-05f3-4c2e-9324-1c817d7c420c

Affects components:
- BitcoinAssetsOrdi
- AssetsCoin
- AssetsNft
- UtxoView

The default herf=# in the test impact component is meaningless and affects the experience, if you need to jump links you can customize the herf separately in the prop.


jojoo-eth (approved):
lgtm

Issue #2382, opened 2024-08-07
[Gas] Reload all native functions when the gas upgrade is complete.


steelgeek091:
Automatically reload all native functions after the gas schedule upgrade transaction executes successfully.

PR #2398, opened 2024-08-09
[VM] Reload all the native functions after the gas schedule upgrade.

steelgeek091:
resolve https://github.com/rooch-network/rooch/issues/2382

### DIFF
--- a/crates/rooch-executor/src/actor/executor.rs
+++ b/crates/rooch-executor/src/actor/executor.rs
@@ -125,6 +125,20 @@ impl ExecutorActor {
             .moveos_store
             .handle_tx_output(tx_hash, output.clone())?;
 
+        // The cost table has been upgraded, we need to reload the native functions.
+        if self.moveos.cost_table.read().is_none() {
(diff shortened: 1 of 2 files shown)


crates/rooch-executor/src/actor/executor.rs
```diff
@@ -125,6 +125,20 @@ impl ExecutorActor {
             .moveos_store
             .handle_tx_output(tx_hash, output.clone())?;
 
+        // The cost table has been upgraded, we need to reload the native functions.
+        if self.moveos.cost_table.read().is_none() {
```
jolestar, line 129: there are another moveos instance in ReadExecutor

jolestar (approved):

crates/rooch-executor/src/actor/executor.rs
steelgeek091: There should be a formal solution to provide a global, asynchronous, and thread-safe subscription and notification mechanism for notifying successful gas upgrades.

PR #2462, opened 2024-08-18
[gh-2406] support rpc requests via proxies.

### LINKED ISSUE #2406: [CLI] support https_proxy,all_proxy env to request RPC
Support https_proxy,all_proxy env to request RPC

ghpZ54K8ZRwU62zGVSePPs97yAv9swuAY0mVDR4:
## Summary

- Support `proxy` env in rooch config files.

- Closes #2406

### DIFF
--- a/crates/rooch-rpc-client/src/lib.rs
+++ b/crates/rooch-rpc-client/src/lib.rs
@@ -2,8 +2,11 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use anyhow::Result;
+use fast_socks5::client::Socks5Stream;
+use fast_socks5::server::{Config, Socks5Socket};
 use jsonrpsee::core::client::ClientT;
 use jsonrpsee::http_client::{HttpClient, HttpClientBuilder};
+use jsonrpsee::ws_client::{WsClient, WsClientBuilder};
 use move_core_types::language_storage::ModuleId;
 use move_core_types::metadata::Metadata;
 use move_core_types::resolver::ModuleResolver;
@@ -15,9 +18,13 @@ use moveos_types::{
     function_return_value::FunctionResult, module_binding::MoveFunctionCaller,
     moveos_std::tx_context::TxContext, transaction::FunctionCall,
 };
+use pin_project::pin_project;
 use rooch_client::RoochRpcClient;
+use std::env;
+use std::net::SocketAddr;
 use std::sync::Arc;
 use std::time::Duration;
+use tokio::net::{TcpListener, TcpStream};
 use tokio::runtime::Handle;
 
 pub mod client_config;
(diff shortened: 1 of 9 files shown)


crates/rooch-rpc-client/src/lib.rs
```diff
@@ -50,9 +58,24 @@ impl ClientBuilder {
...
+            let reqwest_client = Some(Arc::new(
+                reqwest::Client::builder()
+                    .proxy(Proxy::https(proxy_url.clone())?)
+                    .proxy(Proxy::all(proxy_url)?)
+                    .build()?,
+            ));
+            return Ok(Client {
+                http: http_client.clone(),
+                rooch: RoochRpcClient::new(http_client.clone()),
+                reqwest: reqwest_client,
```
jolestar, line 71: We should find method to add proxy to HTTP client and replace the HTTP client in RoochRpcClient, not add a new client.

crates/rooch/src/commands/env/commands/add.rs
```diff
@@ -18,14 +18,23 @@ pub struct AddCommand {
     pub rpc: String,
     #[clap(long, value_hint = ValueHint::Url)]
     pub ws: Option<String>,
+    #[clap(long, value_hint = ValueHint::Url)]
+    pub proxy: Option<String>,
```
jolestar, line 22: The proxy is global, not a configuration option for Rooch Env.

crates/rooch-rpc-client/src/lib.rs
feliciss: > We should find method to add proxy to HTTP client and replace the HTTP client in RoochRpcClient, not add a new client.

From my research, the HTTP client doesn't support proxied request. However, I'll find a way to add a proxy to the http client.

crates/rooch/src/commands/env/commands/add.rs
feliciss: > The proxy is global, not a configuration option for Rooch Env.

I think it should and shall be distinguished from the global proxy for each specific application. Global proxy in the system and the proxy in the application are two different elements.

crates/rooch-rpc-client/src/lib.rs
feliciss: From what I could find so far, the `JsonRpsee` lib only supports `socks5` proxy, meaning that there's no way to connect with `https_proxy` or `all_proxy` through using only the `JsonRpsee` lib. I'm sorry for that.

1. https://grants.web3.foundation/docs/RFPs/jsonrpsee-proxy-support
2. https://github.com/paritytech/jsonrpsee/issues/870
3. https://github.com/paritytech/jsonrpsee/pull/1168

feliciss:
I leave this issue untouched.

ghpZ54K8ZRwU62zGVSePPs97yAv9swuAY0mVDR4:
Is this feature still requested? If not , I will leave it closed.

jolestar:
> Is this feature still requested? If not, I will leave it closed.

Close it first.

Issue #2402, opened 2024-08-10
[genesis] Break change issues


jolestar:
- [x] #2401 
- [x] #2395 
- [x] #2403 
- [x] #2419
- [x] #2474
- [x] #2541 
- [x] #2542 
- [x] #2463

PR #2484, opened 2024-08-21
[moveos_std] Improve the event queue, trigger OffChainEvent sametime

jolestar:
## Summary

follow #2476 

1. Only create the event queue when the first subscriber subscribes to it to reduce the number of EventQueue objects.
2. Trigger an OffChainEvent when OnChainEvent emits; the developer can merge OnChain and OffChain events.

Leave a break change to #2402

### DIFF
--- a/frameworks/moveos-stdlib/sources/event_queue.move
+++ b/frameworks/moveos-stdlib/sources/event_queue.move
@@ -6,7 +6,8 @@ module moveos_std::event_queue {
     use std::option::{Self, Option};
     use std::string::String;
     use moveos_std::timestamp;
-    use moveos_std::object::{Self, Object};
+    use moveos_std::event;
+    use moveos_std::object::{Self, Object, ObjectID};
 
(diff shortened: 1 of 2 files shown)


frameworks/moveos-stdlib/sources/event_queue.move
```diff
@@ -27,16 +29,31 @@ module moveos_std::event_queue {
         emit_time: u64,
     }
 
+    /// The off-chain event
+    /// Every on-chain event also trigger an off-chain event
+    struct OffChainEvent<E> has copy, drop, store {
```
baichuan3, line 34: What is the different between OnChainEvent and OffChainEvent?

jolestar: OffChainEvent emit via `event::emit`, OnChainEvent store on chain.

baichuan3 (approved):

PR #2495, opened 2024-08-23
[framework] Implement compare and sort

jolestar:
## Summary

1. [moveos_std] Implement compare::compare
2. [moveos_std] Implement sort::sort
3. [move_std] Sync vector::slice from aptos-move

### DIFF
--- a/frameworks/move-stdlib/sources/vector.move
+++ b/frameworks/move-stdlib/sources/vector.move
@@ -604,6 +604,20 @@ module std::vector {
         for_each_reverse(v, |e| d(e))
     }
 
+    /// Extracts a sub-vector from `v` starting from `start` index to `end` index (exclusive).
+    public fun sub_vector<T: copy + drop>(v: &vector<T>, start: u64, end: u64): vector<T> {
+        assert!(start <= end, EINVALID_RANGE); 
+        assert!(end <= length(v), EINDEX_OUT_OF_BOUNDS); 
+
+        let result = empty<T>();
+        let i = start;
+        while (i < end) {
+            push_back(&mut result, *borrow(v, i));
+            i = i + 1;
+        };
+        result
+    }
+
     // =================================================================
     // Module Specification
 
--- a/frameworks/move-stdlib/sources/vector.move
+++ b/frameworks/move-stdlib/sources/vector.move
@@ -18,6 +18,12 @@ module std::vector {
     /// The length of the vectors are not equal.
     const EVECTORS_LENGTH_MISMATCH: u64 = 0x20002;
 
+     /// The step provided in `range` is invalid, must be greater than zero.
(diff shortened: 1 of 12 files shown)


frameworks/moveos-stdlib/sources/sort.move
```diff
@@ -0,0 +1,75 @@
+/// Utility functions for sorting vector.
+module moveos_std::sort {
+
+    use std::vector;
+
+    use moveos_std::compare;
+    
+    /// Sorts a vector using quick sort algorithm.
+    public fun quick_sort<T: copy + drop>(data: vector<T>): vector<T> {
```
mx819812523, line 9: 这里拷贝整个数组成本会不会有点大，能不能改成`data: &mut vector<T>`

jolestar: change to &mut vector<T>

mx819812523: `copy` trait for type `T` can be removed

jolestar: We cannot remove the `copy` because we can not borrow two items from a vector.

frameworks/move-stdlib/sources/vector.move
```diff
@@ -604,6 +610,36 @@ module std::vector {
         for_each_reverse(v, |e| d(e))
     }
 
+    public fun range(start: u64, end: u64): vector<u64> {
+        range_with_step(start, end, 1)
+    }
+
+    public fun range_with_step(start: u64, end: u64, step: u64): vector<u64> {
```
baichuan3, line 617: Need to check the size relationship between start, end and step parameters.

baichuan3 (approved):

frameworks/move-stdlib/sources/vector.move
jolestar: There is an assert `step > 0`. If end > start, the result is empty.

frameworks/moveos-stdlib/sources/sort.move
mx819812523: As far as I know, we can `borrow` two items from a vector, but can not `borrow_mut` two items from a vector. If we want to remove `copy` and `drop`, we can change `partition` function
```
fun partition<T>(data: &mut vector<T>, low: u64, high: u64): u64 {
        let i = low;
        let j = low;
        while (j < high) {
            let value = vector::borrow(data, j);
            let cmp = compare::compare(value, vector::borrow(data, high));
            if (cmp == compare::result_less_than()) {
                vector::swap(data, i, j);
                i = i + 1;
            };
            j = j + 1;
        };
        vector::swap(data, i, high);
        i
    }
```

jolestar: fixed at #2503

PR #2499, opened 2024-08-24
[moveos_std] Support compare function

### LINKED ISSUE #2418: [moveos_std] Support general compare function in `compare.move`
In #2373, we introduce a `compare` module in moveos_std and provide `compare_vector_u8` function.

We also need to provide a general `compare<T>(v1: &T, v2: &T)` function.

However, if we directly compare the `bcs::to_bytes` result, we cannot get the correct result when the T is a number.

https://github.com/aptos-labs/aptos-core/blob/f073f58af972428dce7a398d2bdc9d80495b601a/aptos-move/framework/aptos-stdlib/sources/comparator.move#L26-L35

One solution is to detect the type of T and use different compare strategies.

https://github.com/MystenLabs/sui/blob/58fd1d1d2cadec6fb2cc6b85df61727db56a2ad2/crates/sui-oracle/move/oracle/sources/meta_oracle.move#L79-L135

Dependents:
* #2417

vegetabledogdog:
- Closes #2418

### DIFF
--- a/frameworks/moveos-stdlib/doc/compare.md
+++ b/frameworks/moveos-stdlib/doc/compare.md
@@ -10,11 +10,14 @@ Utilities for comparing Move values
 -  [Function `result_equal`](#0x2_compare_result_equal)
 -  [Function `result_less_than`](#0x2_compare_result_less_than)
 -  [Function `result_greater_than`](#0x2_compare_result_greater_than)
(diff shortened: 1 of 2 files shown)


jolestar:
Sorry, this feature is repeated with #2495

PR #2516, opened 2024-08-27
fix: current session may be invalid

### LINKED ISSUE #2481: typescript sdk bug
### Rooch version

rooch 0.6.8 (git commit 82899480bd5fe65b0d7c5502ccf62f010e878b16)

### What did you do?

When I did clear session by ts sdk, sometimes comes out error. When I checking the localstorage , it seems I have 2 sessions. At that time when I clear session it will tell error : 

image

### What did you see happen?

image

### What did you expect to see?

I want it clean well. And the session should have only one.

newraina:
## Summary

Closes #2481 

- remove invalid session from cache storage when receive expired error
- ignore expired sessions when SDK init. (also added `isExpired` method to `Session` class)


jolestar:
@wow-sven @jojoo-eth help to review

sdk/typescript/rooch-sdk-kit/src/provider/useRoochClient.ts
wow-sven, line 1: The hooks directory already exists useRoochClient，This code should still be on the clientProvider

sdk/typescript/rooch-sdk/src/session/session.ts
```diff
@@ -138,6 +138,10 @@ export class Session extends Signer {
     return this.localCreateSessionTime
   }
 
+  isExpired() {
```
wow-sven, line 141: isExpired fun maybe is invalid. After lastActiveTime is updated, the browser does not cache the latest value. Therefore, remove invalid sessions during sdk initialization may not work.

wow-sven:
sdk test currently stuck. wait https://github.com/rooch-network/rooch/pull/2520
You can run tests locally

sdk/typescript/rooch-sdk-kit/src/provider/useRoochClient.ts
newraina: putting it in a separate file is better for testing. I can change the name to avoid conflicts.

sdk/typescript/rooch-sdk/src/session/session.ts
newraina: Got. After offline discussion, the `isExpired` related logic in rooch-sdk will be temporarily removed.

wow-sven (approved):

Issue #2517, opened 2024-08-27
[framework] Refactor package upgrade


pause125:
### Proposal Details

1. Unify package upgrading workflow for both framework packages and normal packages.
2. Split framework packages 0x1/0x2/0x3/0x4, and upgrade each package independently.

Related PR: #2504

PR #2529, opened 2024-08-28
[framework][break] Refactor package publishing

pause125:
## Summary

1. new entrypoint `moveos_std::module_store::publish_package_entry`
2. Sender will get a `moveos_std::module_store::UpgradeCap` object when first pubshing packge. The `UpgradeCap` object is transferrable, and is the only certificate to upgrade the package. Each package(identified with package_id) has only one `UpgradeCap`
3. Each package will track a version, and every time a package is published, an event will be triggered.

Part of #2517


moveos/moveos-types/src/moveos_std/module_store.rs
jolestar, line 1: The rust Package also needs the `version` field.

jolestar:
Need to reset the genesis file, because the Package struct is changed.

jolestar (approved):

PR #2548, opened 2024-08-31
[framework][break] cleanup deprecated functions of packages

### LINKED ISSUE #2505: proposal: A new FunctionArg type file:/local/path/to/file and read file binary as vector<u8>
### Proposal Details

Add a new FunctionArg type file:/local/path/to/file and read file binary as vector<u8>. This is useful for cases like publish module binary.

### LINKED ISSUE #2517: [framework] Refactor package upgrade

pause125:
## Summary

continue of #2529 

resolve #2505
resolve #2517

break changes: @stevenlaw123 @wow-sven @jolestar @baichuan3 

1. The format of exported package binary by `rooch move build` is changed.
2. New publishing workflow: `rooch move framework-upgrade` is deprecated. Now there are two ways to publish modules:
    1). Run `rooch move publish` directly
    2). Run `rooch move bulid -p <path/to/package> --export` to export package binary. And then run `rooch move run --function 0x2::module_store::publish_package_entry --sender default --args 'file:path/to/package.blob'` to submit the transaction.
3. refactor stdlib artifacts.


jolestar (approved):

crates/rooch/src/commands/move_cli/commands/build.rs
```diff
@@ -88,8 +91,25 @@ impl CommandAction<Option<Value>> for BuildCommand {
...
+                )));
+            };
+            let package_data = PackageData::new(
+                MoveString::from(package.compiled_package_info.package_name.as_str()),
+                pkg_address,
+                blob,
+            );
             let mut file = BufWriter::new(File::create(export_path.clone())?);
-            bcs::serialize_into(&mut file, &blob)?;
+            bcs::serialize_into(&mut file, &package_data)?;
```
jolestar, line 112: we can make the default value of `export` to `true`.

jolestar: Or remove the `export` option?

crates/rooch/src/commands/move_cli/commands/publish.rs
```diff
@@ -131,29 +126,20 @@ impl CommandAction<ExecuteTransactionResponseView> for Publish {
...
-                (
-                    "publish_package_entry".to_owned(),
-                    bcs::to_bytes(&pkg_bytes).unwrap(),
-                )
-            };
+            let pkg_data = PackageData::new(
+                MoveString::from(package.compiled_package_info.package_name.as_str()),
+                pkg_address,
+                bundles,
+            );
```
jolestar, line 133: In the future, we can call `build` in publish and directly use the PackageData blob from build dir.

frameworks/moveos-stdlib/sources/module_store.move
```diff
@@ -137,38 +136,7 @@ module moveos_std::module_store {
...
-        // Store ModuleUpgradeFlag in tx_context which will be fetched in VM in Rust, 
-        // and then announce to the VM that the code loading cache should be considered outdated. 
-        tx_context::set_module_upgrade_flag(upgrade_flag);
-    }
-   
-    // Deprecated
-    // TODO: clean this function when reset genesis
-    /// Entry function to publish modules to the account's storage
-    /// The order of modules must be sorted by dependency order.
-    public entry fun publish_modules_entry(account: &signer, modules: vector<vector<u8>>) {
```
jolestar, line 158: Or keep this function for compatibility with the js SDK.

jolestar (approved):

crates/rooch/src/commands/move_cli/commands/build.rs
jolestar: And rename `package.blob` to `package.rpd`,  `rpd` for `Rooch Package Data`?

crates/rooch/src/commands/move_cli/commands/framework_upgrade.rs
jolestar, line 1: How do we upgrade the framework? Should we use `rooch move publish`? Maybe we need to split the `stdlib` file into multiple package files?

frameworks/moveos-stdlib/sources/module_store.move
jolestar: If we do not update the SDK at the same time. Maybe the SDK does not call the publish module directly.

wow-sven:
> @wow-sven `0x2::module_store::publish_modules_entry` is deprecated, the sdk need to be updated to use `0x2::module_store::publish_package_entry` instead.

sdk publish package  it has not been officially used, and we can enable this pr after it is passed

wow-sven (approved):

PR #2614, opened 2024-09-11
[State] Support state sync and provide RPC to sync state

### LINKED ISSUE #2596: [State] Support state sync and provide RPC to sync state
### Proposal Details

State sync usage scenarios
1. The client rebuilds the entire state based on state sync
2. When repairing data, it is used for indexer rebuilding. For example, when reorg is performed, the state is rolled back and the indexer is also rolled back.

baichuan3:
## Summary

1. Support state sync and provide RPC to sync state
2. Revert indexer based on state change set when rollback tx
3. Replace tx and event indexer field sender from RoochAddress to AccountAddress, for reduce sender store size in SQLite by using short_str_lossless
4. Refactor revert tx and rollback tx tool to reuse code

TODO
RocksDB uses compaction to remove expired keys, setting TTL directly in RocksDB may not be a good choice.
and it may also have a performance impact.
Cleaning up state change set data regularly may be an option

- Closes #2596

### DIFF
--- a/crates/rooch/src/commands/db/commands/rollback.rs
+++ b/crates/rooch/src/commands/db/commands/rollback.rs
@@ -163,7 +163,6 @@ impl RollbackCommand {
                 .rooch_store
                 .get_state_change_set(previous_tx_order)?;
             if previous_state_change_set_ext_opt.is_some() && state_change_set_ext_opt.is_some() {
-                // let previoud_state_root = previous_state_change_set_ext_opt.unwrap().state_change_set.state_root;
                 let previous_state_change_set_ext = previous_state_change_set_ext_opt.unwrap();
                 let state_change_set_ext = state_change_set_ext_opt.unwrap();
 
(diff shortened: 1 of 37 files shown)


crates/rooch/src/commands/db/commands/rollback.rs
```diff
@@ -123,29 +138,102 @@ impl RollbackCommand {
...
+            rooch_db
+                .moveos_store
                 .transaction_store
                 .remove_tx_execution_info(tx_hash)?;
             println!(
                 "remove tx succ, tx_hash: {:?}, tx_order {}",
                 tx_hash, tx_order
             );
+
+            // remove the state change set
```
jolestar, line 154: 这块逻辑是不是也封装到 RoochDB 中，因为 PipelineProcessorActor 中会自动 revert 交易。

https://github.com/rooch-network/rooch/blob/68c09053542426562dddedb9eb0f5e02e021ec88/crates/rooch-pipeline-processor/src/actor/processor.rs#L224-L237

cc @steelgeek091

jolestar (approved):

crates/rooch/src/commands/db/commands/rollback.rs
steelgeek091: 感觉封装到 rooch-db 没问题，基本上都是对 rooch-db 的操作。

baichuan3: already refactor revert tx and rollback tx tool to reuse code

PR #2616, opened 2024-09-11
Gas Profiling Implementation

steelgeek091:


crates/rooch-executor/src/actor/reader_executor.rs
```diff
@@ -146,7 +146,12 @@ impl Handler<StatesMessage> for ReaderExecutorActor {
         msg: StatesMessage,
         _ctx: &mut ActorContext,
     ) -> Result<Vec<Option<ObjectState>>, anyhow::Error> {
-        let resolver = RootObjectResolver::new(self.root.clone(), &self.moveos_store);
+        let resolver = if let Some(state_root) = msg.state_root {
+            let root_object_meta = ObjectMeta::root_metadata(state_root, 55);
```
steelgeek091, line 150: How should we handle the size parameter? 
Should it be passed from the client or retrieved from Node?

baichuan3: Another way to implement it is to modify the state_resolver StateReader trait to support get_states and list_states by state root, so there is no need to construct a new RootObject

jolestar: In the read-only case, the VM do not use the size param, it can be zero.

jolestar (approved):

crates/rooch-executor/src/actor/reader_executor.rs
baichuan3: It is workable to implement this by directly setting the RootObject size to 0, but is there any ambiguity in constructing the RootObject for any state root query, such as the state root of a child object tree?

jolestar: The size is used by the ObjectRuntime; if the contract removes an element, the size will reduce. We do not record all Global root objects. This is a workaround solution; we will need to refactor in the future.

baichuan3: Do not need to record all Global root objects , just modify the state_resolver StateReader trait to support get_states and list_states by state root , and it can elegantly solve the querying arbitrary state root scenarios. 
It can be put in the subsequent refactor.

jolestar: There is already a StatelessResolver, which supports get_states by root; StateResolver is for the pin on a state_root and get states many times.

PR #2635, opened 2024-09-14
[cli] Migrate Bitseed cli to rooch

jolestar:
## Summary

1. Rename BitcoinOutPointView to OutPointView
2. Remove BlockHash from Seed
3. Use local keystore to sign the tx.
4. Refactor UTXO selector.
5. Use `rooch bitseed` to replace the bitseed image in integration test.

- Closes #1745

PR #2658, opened 2024-09-19
[bitseed] Implement bitseed merge

jolestar:
## Summary

Following #2635, please review #2635 first.

1. [bitcoin-move] Do not freeze the Inscription Object when burned; transfer it to the bitcoin_move address.
2. [bitseed] Migrate the bitseed runner process into the inscribe_factory module and remove it from examples.
3. [bitseed] Refactor inscribe merge, put the burn inputs after the reveal inputs.
4. [bitseed] Implement validate logic for bitseed merge.

TODO:

1. Implement birdseed split validation.
2. Refactor the inscribe_factory, make code clear.


baichuan3 (approved):

yubing744:
image

Now the latest Merge data stream becomes like this?

yubing744:
It seems that if the total number of SFT inscriptions burned in a transaction is greater than or equal to the total number of SFT inscriptions generated, then this operation should be valid. If so, the implementation of Merge and Split is the same.

PR #2665, opened 2024-09-21
fix(da): rm open-da(gcs) for mainnet deploy workflow

popcnt1:
## Summary

1. rm open-da(gcs) for mainnet deploy workflow
2. switch to save last block instead of a map: Replaced BTreeMap with an Option to store only the last block in StateCommitmentChain. Updated methods for adding and retrieving the last block accordingly, streamlining state management.


jolestar (approved):

yourmoonlight (commented):
LGTM

PR #2780, opened 2024-10-17
fix(sdk): duplicate instance

### LINKED ISSUE #2779: [SDK] Duplicate SDK instance
### Rooch version

rooch 0.7.4 (git commit 196f2932ae86d0f8dc5e625a2207fd0c6284f56f)

### What did you do?

Use `useWallets` from `@roochnetwork/rooch-sdk-kit` in the code to get all supported wallets.

### What did you see happen?

useWallets will return duplicate wallet items:

CleanShot 2024-10-17 at 12 22 24@2x

### What did you expect to see?

useWallets should return a non-duplicate array of wallets.

newraina:
### Root Cause

The function `checkWallets` is called multiple times, which creates different wallet instances for the same wallet class. Because of this, even though `registered` is a Set, it still ends up adding multiple duplicate wallets.

https://github.com/rooch-network/rooch/blob/b98ce6fefb1433d861f9d292a900b681f62c3352/sdk/typescript/rooch-sdk-kit/src/wellet/wallets.ts#L6-L8

### How to Fix

At the SDK level, each wallet class should have only one instance.

- Closes #2779

### DIFF
--- a/sdk/typescript/rooch-sdk-kit/src/utils/walletUtils.ts
+++ b/sdk/typescript/rooch-sdk-kit/src/utils/walletUtils.ts
@@ -7,10 +7,10 @@ import { Wallet, UniSatWallet, OkxWallet, OnekeyWallet } from '../wellet/index.j
 const unisatWallet = new UniSatWallet()
 const okxWallet = new OkxWallet()
 const onekeyWallet = new OnekeyWallet()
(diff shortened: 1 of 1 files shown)


jolestar:
@wow-sven @jojoo-eth help review

PR #2817, opened 2024-10-24
feat(rooch-da): add server status get

popcnt1:
## Summary

Added the ability to get DA server status, including last block number, transaction order, and backend availability. Updated background submitter to improve block submission handling and error reporting.


baichuan3 (approved):

crates/rooch-types/src/da/state.rs
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) RoochNetwork
+// SPDX-License-Identifier: Apache-2.0
+
+/// DA server basic status
+pub struct ServerStatus {
```
jolestar, line 5: Maybe we should include the DA ServerStatus to `Status`

```
https://github.com/rooch-network/rooch/blob/3234181217bcef1181403e3c9c0acbb74e3bb532/crates/rooch-rpc-api/src/jsonrpc_types/status.rs#L104-L114
```

Issue #2864, opened 2024-11-07
[CLI] Use tabled as cli default output style


jolestar:
### Proposal Details

https://github.com/zhiburt/tabled/

ghpZ54K8ZRwU62zGVSePPs97yAv9swuAY0mVDR4:
I can replace `display_json_as_table` with it in `session_key` module together with #630 if no one assigned.

PR #2870, opened 2024-11-08
[CLI] Use tabled as cli default output style #

baicaiyihao:
## Summary

This update improves the `rooch` CLI output formatting using the `tabled` library. It aims to optimize the display of data tables to enhance readability, especially for commands that return long text fields. Additional table views across the CLI can be similarly optimized if provided.

### Examples of Optimized CLI Commands

1. **`rooch account list`**: The output for account lists is adjusted so that multiple wallet addresses are consolidated into a readable table format, preventing line wrapping due to overly long addresses.

   Account List Output

2. **`rooch account balance`**: Account balances are displayed with clearer column formatting, ensuring balance values align effectively.

   Account Balance Output

3. **`rooch env list`**: The environment list displays each configuration in a structured table format, improving readability.

   Env List Output

4. **`session-key list`**: Adds a new feature that converts JSON output into a table format when using the `--table` flag.

   - **Table Format**:
     ```bash
     rooch session-key list --table
     ```
     Session-Key Table Output

   - **JSON Format**:
     ```bash
     rooch session-key list
     ```
     Session-Key JSON Output

References #2864

### DIFF
--- a/crates/rooch/src/commands/session_key/mod.rs
+++ b/crates/rooch/src/commands/session_key/mod.rs
@@ -1,6 +1,5 @@
-// Copyright (c) RoochNetwork
-// SPDX-License-Identifier: Apache-2.0
-
+use serde_json::Value;
+use tabled::{builder::Builder, settings::Style};
 use crate::cli_types::CommandAction;
 use async_trait::async_trait;
 use clap::Parser;
@@ -10,7 +9,6 @@ use rooch_types::error::RoochResult;
 
 pub mod commands;
 
-/// Session key Commands
 #[derive(Parser)]
 pub struct SessionKey {
     #[clap(subcommand)]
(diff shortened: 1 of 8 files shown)


baicaiyihao:
Currently, only some CLI functions have been updated. If there are other functionalities that would benefit from similar improvements, please list them so they can be optimized together.

popcnt1 (approved):

crates/rooch/src/commands/session_key/mod.rs
```diff
@@ -35,3 +48,101 @@ pub enum SessionKeyCommand {
     Create(Box<CreateCommand>),
     List(ListCommand),
 }
+
+fn display_json_as_table(data: &Value) {
```
jolestar, line 52: We can make the `--json` and `--table` as an option to every command, and auto display json to table as default in the future.

jolestar (approved):

PR #2871, opened 2024-11-08
Check the ability of the type T in Object

steelgeek091:


examples/argument_resolver/sources/main.move
```diff
@@ -34,6 +35,13 @@ module argument_resolver::argument_resolver {
         object::transfer(user_object, user_address);
     }
 
+    entry public fun create_object_without_key_ability(account: &signer) {
```
jolestar, line 38: `without_store_ability`?

jolestar (approved):

examples/argument_resolver/sources/main.move
steelgeek091: > `without_store_ability`?

OK, it's fixed.

PR #2873, opened 2024-11-09
fix twitter bind bug

vegetabledogdog:
The tweet address must be split by the ADDRESS_SPLIT_CHARS


infra/rooch-portal-v2/src/sections/settings/view.tsx
```diff
@@ -128,13 +128,13 @@ export function SettingsView() {
                 </Stack>
                 {address && (
                   <CopyToClipboard
-                    text={`BTC:${address.toStr()}
+                    text={`BTC:${address.toStr()}.
```
jolestar, line 131: Use a space char maybe better than `.` ?

jolestar (approved):

PR #2936, opened 2024-11-21
[framework] Gas fee distribute

jolestar:
## Summary

1. Distribute the gas fee to the sequencer and contract address.
2. Use gas_schedule::max_gas_amount const in transaction_validator. @steelgeek091 

The benchmark result is regressed because we need to split the gas fee. I can not find a better solution to optimize it.

Screenshot 2024-11-26 at 16 47 02


mx819812523:
The contract must define the method for receiving gas fees, right？ because we need `moveos_std::signer::module_signer`

jolestar:
> The contract must define the method for receiving gas fees, right？ because we need `moveos_std::signer::module_signer`

Yes, the entry function module receives the RGas.

mx819812523 (approved):

PR #2954, opened 2024-11-24
refactor(rooch-da): put da info into rooch status

popcnt1:
## Summary

1. rename DAServerStatus to DAInfoView and 
2. use StrView for DAInfoView


jolestar (approved):
This change will break the RPC, we need to release a new version.

PR #2968, opened 2024-11-27
migrate the move repository into the third_party directory

steelgeek091:


steelgeek091:
We should exclude the 'third_party' from the license checking.

jolestar (approved):

steelgeek091:
> some CI checks should be skipped for the thrid_party directory.

Fixed.

Issue #2751, opened 2024-10-12
[Babylon]  Rooch node equip with babylon finality


baichuan3:
### Proposal Details

The rooch node processes the finality logic

PR #2992, opened 2024-12-01
[babylon]  POC implements rooch finality

baichuan3:
## Summary

POC implements rooch finality, and equip with babylon finality gadget sidecar, part of Integration babylon https://github.com/rooch-network/rooch/issues/2751

- Closes #issue


crates/rooch-types/src/block.rs
```diff
@@ -11,32 +15,64 @@ pub struct Block {
     pub block_number: u128,
     /// How many transactions in the block
     pub batch_size: u64,
-    /// The hash of the batch, made by DA
-    pub batch_hash: H256,
+    /// The hash of the block, made by DA
+    pub block_hash: H256,
```
popcnt1, line 19: I used batch here, because DA only care about tx list. Not state made by present. If we call it block hash, may hint there is execution inside it? Actually, this hash contains no result of tx execution at all.

Which proposer needs is a hash which combine state and tx itself(state_root & batch_hash)

baichuan3: Yes, the block hash in the proposer needs to be associated with the transaction (need to check the block based on tx hash or tx order). This is somewhat different from the batch hash in DA, but it would be best if they could be unified.

PR #3063, opened 2024-12-17
Use versioned gas configuration to modify or add gas entries on-chain

steelgeek091:


jolestar (approved):
Add a CI task to generate genesis.

steelgeek091:
> Add a CI task to generate genesis.

Fixed

crates/rooch-genesis/src/lib.rs
```diff
@@ -300,8 +300,10 @@ impl RoochGenesis {
                 || network.chain_id == BuiltinChainID::Local.chain_id()
             {
                 FrameworksGasParameters::latest()
-            } else {
+            } else if network.chain_id == BuiltinChainID::Main.chain_id() {
                 FrameworksGasParameters::initial()
+            } else {
+                FrameworksGasParameters::v1()
             }
```
jolestar, line 307: 这里可以换个写法

if main initial
else if test v1
else latest

steelgeek091: Fixed

PR #3072, opened 2024-12-20
feat: check exist before batch handle

lispking:
## Summary

Before executing the batch operation, it is necessary to determine whether there are records or not. If there are no records, skip it.


jolestar (commented):
batch_harvest also needs to be updated.

jolestar (approved):

Issue #2837, opened 2024-10-31
[rooch-framework] Calling simple_rng::seed() multiple times in the same transaction returns the same value


yubing744:
### Proposal Details

Problem description:
In the same transaction, multiple calls to simple_rng::rand_u64_range return the same result, resulting in multiple random coordinates generated in the game being the same.

Expected result:
Each call to simple_rng::rand_u64_range returns a different random number.

Recommended fix:
Add a counter to tx_context, add +1 each time it is called, and seed generation adds the current value of this counter.

Security risk:
It is necessary to prevent external modules from modifying this counter and implementing random number probing attacks.

PR #3076, opened 2024-12-20
feat: add rand_u64_with_count and rand_u128_with_count

yubing744:
## Summary

Fix Calling simple_rng::seed() multiple times in the same transaction returns the same value

- Closes #issue
https://github.com/rooch-network/rooch/issues/2837

### DIFF
--- a/frameworks/rooch-framework/sources/simple_rng.move
+++ b/frameworks/rooch-framework/sources/simple_rng.move
@@ -78,8 +78,13 @@ module rooch_framework::simple_rng {
     fun seed_with_count(count: u64): vector<u8> {
         let base_seed = seed();
         let count_bytes = bcs::to_bytes(&count);
-        vector::append(&mut base_seed, count_bytes);
-        hash::sha3_256(base_seed)
+        let i = 0;
+        while (i < 8) {
+            let value = (*vector::borrow(&base_seed, i)) ^ (*vector::borrow(&count_bytes, i));
+            vector::insert(&mut base_seed, i, value);
+            i = i + 1;
+        };
+        base_seed
     }
 
     /// Generate a random u64 from seed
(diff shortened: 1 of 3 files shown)


yubing744:
@jolestar help review.

yubing744:
@jolestar help review it again.

frameworks/rooch-framework/sources/simple_rng.move
```diff
@@ -75,32 +75,70 @@ module rooch_framework::simple_rng {
...
 
+    fun seed_with_count(count: u64): vector<u8> {
+        let base_seed = seed();
+        let count_bytes = bcs::to_bytes(&count);
+        let i = 0;
+        while (i < 8) {
+            let value = (*vector::borrow(&base_seed, i)) ^ (*vector::borrow(&count_bytes, i));
+            vector::insert(&mut base_seed, i, value);
+            i = i + 1;
+        };
```
jolestar, line 86: Use `vector::append`?

yubing744: I tested and found that bytes_to_u64 only takes the first 8 bytes of the output bytes, so appending directly has no effect.

jolestar: We should update the seed function and add the count to the data before hashing.

yubing744: Directly modifying the seed function won’t cause compatibility issues?

jolestar: The seed function is a private function; no compatibility issue

yubing744: OK

yubing744: @jolestar help review it again.

jolestar (approved):

PR #3098, opened 2024-12-24
feat(rooch-relayer): add Bitcoin reorg-aware block store

### LINKED ISSUE #3078: proposal: btc-relyaer: save recent btc block local
### Proposal Details

background:

when replay tx local, we may meet re-org btc block which has been invalid after new block made at same height. Som in history tx, we will have may than one btc block at same height.

But in our ledgerTx, we only have height and block_hash for l1block. So we need find a way to build l1blockwithbody.

approach:

1. save recent blocks local: block_hash: block kv pair and block height about this pair
2. wrap btc_client, make it could not only get btc_block from btc network but also get block locally
3. if there is no new block for same height, and this height is beyond re-org protection, which means there is no re-org. we could safely delete the historcal blocks, because we could download this block from btc network at any time
4. if there is a new block for the same height, mark origin block seal. we should not remove this block

popcnt1:
## Summary

1. Introduce BitcoinReorgAwareStore to handle potential reorg scenarios by storing and replacing block data. Includes configuration options for block storage directory and reorganization height.

2. feat(bitcoin-client): add local block store support for reorg handling:
    
    Enable reading blocks from a local block store as a fallback when RPC fetching fails. Adds an optional configuration for the block storage directory to improve resilience against reorganization scenarios.

- Closes #3078


crates/bitcoin-client/src/actor/client.rs
```diff
@@ -88,7 +99,39 @@ impl Handler<GetBlockMessage> for BitcoinClientActor {
         _ctx: &mut ActorContext,
     ) -> Result<bitcoin::Block> {
         let GetBlockMessage { hash } = msg;
-        Ok(self.retry(|| self.rpc_client.get_block(&hash)).await?)
+
+        let rpc_ret = self.retry(|| self.rpc_client.get_block(&hash)).await;
+        if let Ok(block) = rpc_ret {
+            return Ok(block);
+        }
```
jolestar, line 106: Maybe read the local store first, then get from rpc? the local store can be a cache.

crates/rooch-relayer/src/actor/bitcoin_relayer.rs
```diff
@@ -244,3 +260,52 @@ impl Handler<GetReadyL1TxsMessage> for BitcoinRelayer {
...
+    ) -> Self {
+        Self {
+            block_store_dir,
+            recent_blocks_map: IndexMap::with_capacity(aware_height),
+            aware_height,
+            rpc_client,
+        }
+    }
+
+    pub async fn insert_or_replace(
```
jolestar, line 285: Should we put the BitcoinReorgAwareStore logic into BitcoinClientActor? Make the read-and-write operation in the same mod.

crates/bitcoin-client/src/actor/client.rs
popcnt1: local block store will only helpful when:

1. for online server, may need reorg block in startup for executor catch up sequencer
2. for offline execution, we need to upload these reorg to public storage, users could downloads these block

For normal process, this cache cannot help in any way

crates/rooch-relayer/src/actor/bitcoin_relayer.rs
popcnt1: In present, client is read-only too (for connect bitcoin node). And it's hard for a client to know which block has been consumed and could regard as reorg for Rooch

crates/bitcoin-client/src/actor/client.rs
jolestar: I mean, the local store saves all the blocks.

crates/rooch-relayer/src/actor/bitcoin_relayer.rs
jolestar: If the BitcoinClientActor saves all blocks as cache data, maybe we can remove this function.

jolestar (approved):

PR #3125, opened 2024-12-30
bugfix: build failed by dependency log not exists issue

lispking:


.gitignore
```diff
@@ -39,3 +39,4 @@ sdk/typescript/src/generated/
 
 # pnpm store
 .pnpm-store/
+pnpm-lock.yaml
```
jolestar, line 42: Should the `pnpm-lock.yaml` be ignored? @wow-sven

jolestar (approved):

PR #3304, opened 2025-02-11
feat(rooch-sequencer): prevent duplicate transactions

popcnt1:
## Summary

Add a check in the sequencer to verify if a transaction already exists based on its hash. Return an error if a duplicate transaction is detected, ensuring better consistency and data integrity.

Fix old transactions with the same tx_hash being overwritten


popcnt1:
duplicate tx caused by unexpected error return has been fixed: https://github.com/rooch-network/rooch/pull/3316

some tx_outputs aren't handle in right way, so the sequence number won't be added as expect. That's why there is duplicate tx.

PR #3320, opened 2025-02-14
Print full tx hash for trouble shoot

baichuan3:
## Summary

The tx hash printed in the current log is truncated, which is not convenient for troubleshooting. All logs output the complete tx hash.

- Closes #issue


popcnt1 (approved):
very helpful

jolestar (approved):

PR #3361, opened 2025-02-25
ci(deploy): add Turbo DA support for mainnet deployment

ghost:
## Summary

Extend mainnet deployment script to include Turbo DA backend configuration. Updated secrets and workflow to accommodate new endpoint and API key inputs.


jolestar:
Do we need to wait for the new version to deploy to the mainnet?

PR #3418, opened 2025-03-12
[devops] Run integration test parallel

baichuan3:
## Summary

- Closes #issue


baichuan3:
Integration tests cannot be run in parallel. Port conflicts will occur when starting the server. It need to use a random port.

PR #3595, opened 2025-05-23
DID on Move

jolestar:
## Summary

1. SessionKey supports Secp256k1.
2. AuthValidator supports smart contract accounts to send transactions via SessionKey.
3. Multibase public encode/decode.
4. DID on Move framework.
5. DID command in cli.


frameworks/rooch-framework/sources/did.move
```diff
@@ -0,0 +1,1811 @@
...
+        let did_account_address = account::account_cap_address(&did_document_data.account_cap);
+        
+        // 1. Verify signer is the DID's associated account
+        assert!(sender == did_account_address, ErrorSignerNotDIDAccount);
+        
+        // 2. Get current transaction's session key (authentication_key)
+        let session_key_opt = auth_validator::get_session_key_from_ctx_option();
+        assert!(option::is_some(&session_key_opt), ErrorNoSessionKeyInContext);
+        
+        let session_key = option::extract(&mut session_key_opt);
```
mx819812523, line 1478: Do we need to verify that the session key has not expired here?

PR #3600, opened 2025-05-27
Added missing range checks to Sha256Partial and Sha256General templates to prevent comparator overflows

### LINKED ISSUE #3599: Missing Range Constraints in Sha256Partial and Sha256General Allow Forged Inputs to Pass Verification
I would like to report a security vulnerability related to missing range checks in the [`Sha256Partial`](https://github.com/rooch-network/rooch/blob/d684170a08658da7ae5c73af07a680ee9dcb59ef/sdk/circomlib/circuits/sha256partial.circom#L9) and [`Sha256General`](https://github.com/rooch-network/rooch/blob/d684170a08658da7ae5c73af07a680ee9dcb59ef/sdk/circomlib/circuits/sha256general.circom#L9) templates. I will explain the issue using `Sha256Partial` as the primary example, but note that the same vulnerability also applies to `Sha256General`.

## Issue Description

The `Sha256Partial` template implements partial SHA-256 hashing over padded input blocks. It enforces several constraints to ensure correct operation:

1. It checks that the input padded bit length `in_len_padded_bits` is less than or equal to `maxBitsPadded`, using `LessEqThan(n)`.
2. It checks that `in_len_padded_bits` is a multiple of 512, via an equality constraint with `inBlockIndex * 512`.
3. It uses the value `inBlockIndex - 1` as an index for the `QuinSelector` template (outputs one of the array values based on a given selector index), which uses `LessThan(n)` internally to verify that the index lies within bounds.

However, these comparator templates (`LessEqThan(n)`, `LessThan(n)`, etc.) of circomlib assume that the inputs they receive fit within `n` bits (the parameter value), and thus do not internally constrain their inputs (for performance reasons). This can be exploited if the inputs are not range-constrained using `Num2Bits(n)` before instantiating the template.

Since no `Num2Bits` constraint is applied to `in_len_padded_bits` or `inBlockIndex`, a malicious prover can craft forged values that pass all constraints and manipulate the final SHA-256 output.

For background on this type of issue, please see:

* [zk-bug-tracker #4: Mismatching Bit Lengths](https://github.com/0xPARC/zk-bug-tracker?tab=readme-ov-file#4-mismatching-bit-lengths)
* [BlakeMScurr/comparator-overflow](https://github.com/BlakeMScurr/comparator-overflow)

## Exploits

We demonstrate two exploits that highlight this vulnerability:

### 1. **Length Check Bypass (LessEqThan only)**

We hereby bypass the `LessEqThan` check with a value `> maxBitsPadded`, which is intended to enforce the constraint `in_len_padded_bits ≤ maxBitsPadded`. To set up this example, we instantiate the `Sha256Partial` template with:

```
maxBitsPadded = 3 * 512 = 1536
```

This corresponds to 3 full 512-bit SHA-256 blocks. From this value, the following internal parameters are derived in the template:

```
maxBitsPaddedBits = ceil(log2(1536)) = 11
```

This leads to the instantiation of:

```circom
component bitLengthVerifier = LessEqThan(11);
bitLengthVerifier.in[0] <== in_len_padded_bits;
bitLengthVerifier.in[1] <== 1536;
bitLengthVerifier.out === 1;
```

Notice that `LessEqThan(11)` assumes that `in[0]` is less than 2¹¹ = 2048, but this is not enforced anywhere! A malicious prover can thus supply a much larger field element, cause an overflow, and pass the check. Specifically, we set:

```
in_len_padded_bits = p − 1
```

where `p` is the BN254 scalar field modulus used by Circom:

```
p = 21888242871839275222246405745257275088548364400416034343698204186575808495617
```

We also satisfy the accompanying equality:

```
in_len_padded_bits === inBlockIndex * 512
```

by setting:

```
inBlockIndex = (p − 1) / 512 = 42750474359061084418450011221205615407321024219562567077535555051905875968
```

All remaining lines until the LessThan consist of `assert` statements and weak assignments using `<--`, which do not contribute any constraints that are enforced by the verifier. As a result, they can be ignored in the context of our attack.

The `LessEqThan(11)` check passes because it internally performs the check by computing:

```
z = in[0] + 2^11 − 1536 = in[0] + 2048 − 1536 = in[0] + 512
```
and then decomposes `z` into 12 bits using `Num2Bits(12)` (this check is for the difference calculated inside the template, not for the inputs!) . The check passes if:
```
0 ≤ z < 2^12 = 4096
```

It then verifies that the most significant bit of the 12-bit decomposition of `z` is zero. In other words, that:

```
0 ≤ z < 2^12 = 4096
```

In our case, we set:

```
in[0] = in_len_padded_bits = p − 1
```

So:

```
z = p − 1 + 512 ≡ 511 mod p
```

Since `511 < 4096`, the check passes, even though `in_len_padded_bits ≫ 1536`! This allows the constraint to be satisfied despite the value being well outside the intended range.

The following files are included to reproduce and validate the exploit:

* [`sha256_lessthan_attack.circom`](https://gist.github.com/ArmanKolozyan/720e9d7056ffacc5acfaab7b21fbc6c7): A template containing all the relevant constraints.
* [`input_lessthan.json`](https://gist.github.com/ArmanKolozyan/e90b3cdd993958886adf8bb6ca4638b1): The specific inputs that satisfy the constraints while triggering the overflow.

### 2. **Selector Bypass**

If we continue looking at what happens further in the `Sha256Partial` template with the same forged values used in Exploit 1, we find a second vulnerability, this time in the `QuinSelector` component.

After computing `inBlockIndex` (which we set to `(p − 1) / 512`), the template computes an index:

```
index = inBlockIndex − 1
```

This index is then passed into 256 parallel instances of `QuinSelector`, each of which selects one SHA-256 output word based on the provided index. Internally, each `QuinSelector` includes the following comparator:

```circom
component lt = LessThan(bits);
lt.in[0] <== index;
lt.in[1] <== choices;
lt.out === 1;
```

This is intended to check that `index < choices`, where:

```
choices = maxBlocks = maxBitsPadded / 512 = 3
bits = maxBlocksBits = ceil(log2(3)) = 2
```

However, just like `LessEqThan(n)`, the `LessThan(n)` comparator assumes that its inputs are less than $2^n$, and does not internally enforce this with a `Num2Bits` check.

We now exploit this by continuing from our earlier forged values, but adjusting `inBlockIndex` to be:

```
inBlockIndex = 0
index = inBlockIndex − 1 = p − 1
```

Now the internal value that `LessThan(2)` computes is:

```
z = index + 2^2 − 3 = (p − 1) + 4 − 3 = p ≡ 0 mod p
```

Since `z = 0`, the check passes because `0 < 2^2 = 4`, and the MSB of its 3-bit binary decomposition is zero. Thus, `lt.out === 1`, even though `index ≫ choices = 3`.

This allows the prover to silently bypass the intended range check on the selector index.

The attack continues as follows: each `QuinSelector` tries to compare the `index` to values `i ∈ [0, 1, 2]` using `IsEqual(i, index)`. But since `index = p − 1`, no equality will succeed. As a result:

```
eqs[i].out = 0  for all i
calcTotal.sum = 0
```

Each of the 256 selectors will output zero, and so `Sha256Partial` returns an all-zero SHA-256 output, even though no valid compression block was selected. All constraints still pass.

In practice, this means that a malicious prover can convince the verifier that a padded message hashes to a block of all-zero bits, while the verifier has no way of detecting that the output is completely detached from the real SHA-256 state.

We include the following files to reproduce and validate this exploit:

* [`sha256_partial_attack.circom`](https://gist.github.com/ArmanKolozyan/7c0078272293a55991408047cab070d6): A template containing all the relevant constraints.
* [`input.json`](https://gist.github.com/ArmanKolozyan/3be3f461b81c886111669cb783f1245b): The crafted witness values that satisfy all constraints while bypassing the selector range check.

## Recommendations

To fix this issue, we recommend adding the following constraint to `Sha256Partial` and `Sha256General`:

```circom
component lenBits = Num2Bits(maxBitsPaddedBits);
lenBits.in <== in_len_padded_bits;
```

This ensures that `in_len_padded_bits < 2^maxBitsPaddedBits`, making the `LessEqThan` comparator sound and indirectly constraining `inBlockIndex` and `index`.

Otherwise, if such constraints are intentionally omitted for performance reasons, this assumption must be clearly documented. For instance, the `zk-email-verify` project [correctly applies a `Num2Bits` constraint](https://github.com/zkemail/zk-email-verify/blob/b193cf0c760456b837b2bbcf7b2c72d5bb3f43c3/packages/circuits/email-verifier.circom#L58) before passing inputs to the hasher. However, I have identified multiple public usages of this template on GitHub that forget the necessary range check (which I am now reporting). To avoid such misuse in the future, it is safer to enforce the `Num2Bits` constraint directly within the template.

ArmanKolozyan:
## Summary

This pull request addresses a security issue related to missing range checks in the `Sha256Partial` and `Sha256General` templates:

* \#3599
  Both templates use comparator components (`LessEqThan`, `LessThan`) that assume their inputs are bounded to a specific bit length, but this assumption was not enforced. As a result, malicious inputs to `in_len_padded_bits` and the derived `index = inBlockIndex - 1` could bypass constraints, leading to incorrect SHA-256 outputs while still satisfying all circuit checks.

To fix this, we add an explicit `Num2Bits` constraint on `in_len_padded_bits` in both templates.

For full details, please refer to the linked issue.

- Closes #3599.


jolestar:
Thank you.

PR #3974, opened 2026-03-19
fix(statedb): correct rebase artifact boundaries

jolestar:
## Summary
- add explicit end-of-object markers to the rebase artifact so paged parent objects rebuild correctly
- capture `sequencer_info` at export start so artifact cutover metadata is not taken after traversal finishes
- strengthen the roundtrip test to force paging/chunking and exercise the object-boundary case
- keep the design note aligned with the in-tree rebase implementation status

## Testing
- `cargo fmt --all`
- `cargo test -p rooch test_rebase_export_and_build_roundtrip -- --nocapture`

## Notes
- This PR fixes correctness issues in the current artifact/export-build flow.
- It does **not** claim fresh-build continuation is complete yet; tx accumulator and minimal tx-history dependencies still need a follow-up pass.

### DIFF
--- a/crates/rooch/src/commands/statedb/commands/rebase.rs
+++ b/crates/rooch/src/commands/statedb/commands/rebase.rs
@@ -3,6 +3,7 @@
 
 use std::collections::{BTreeMap, HashSet};
 use std::fs::{self, File};
+use std::path::Component;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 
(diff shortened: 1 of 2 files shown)


jolestar:
Addressed the current Copilot comments:
- validated artifact paths from the manifest so `chunk_files`, `genesis_file`, and `sequencer_file` must remain under the expected `objects/` or `meta/` subdirectories
- refreshed the design note so section 6 reflects the current BCS chunk layout instead of the older NDJSON proposal
- reran `cargo fmt --all` and `cargo test -p rooch test_rebase_export_and_build_roundtrip -- --nocapture`

The remaining known gap is still fresh-build continuation dependencies (`tx_accumulator` + minimal tx-history metadata), which is intentionally left for follow-up.

PR #3990, opened 2026-04-07
Redesign mainnet replay output store build

jolestar:
## Summary
- replace replay output construction with a fresh RocksDB plus selective CF copy
- import snapshot state nodes into the fresh DB and rebuild startup/sequencer metadata during replay
- add regression tests for fresh output creation, selective CF copy, and sequencer rewrite behavior

## Testing
- cargo check -p rooch-pruner --tests (still running / not completed in this session)
- cargo test -p rooch-pruner incremental_replayer -- --nocapture (build started but not completed in this session)

### DIFF
--- a/crates/rooch-pruner/src/state_prune/incremental_replayer.rs
+++ b/crates/rooch-pruner/src/state_prune/incremental_replayer.rs
@@ -4,11 +4,11 @@
 use crate::state_prune::{ProgressTracker, StatePruneMetadata};
 use anyhow::Result;
 use move_core_types::effects::Op;
+use moveos_common::utils::to_bytes;
 use moveos_config::store_config::RocksdbConfig;
 use moveos_store::transaction_store::TransactionStore as MoveOSTransactionStore;
 use moveos_store::{
-    MoveOSStore, CONFIG_GENESIS_COLUMN_FAMILY_NAME, EVENT_COLUMN_FAMILY_NAME,
-    EVENT_HANDLE_COLUMN_FAMILY_NAME, STATE_NODE_COLUMN_FAMILY_NAME,
+    MoveOSStore, CONFIG_GENESIS_COLUMN_FAMILY_NAME, STATE_NODE_COLUMN_FAMILY_NAME,
     TRANSACTION_EXECUTION_INFO_COLUMN_FAMILY_NAME,
 };
 use moveos_types::h256::H256;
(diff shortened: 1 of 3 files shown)


jolestar (commented):
我本地 review 了这版 selective-CF replay，有 3 个 blocker 需要先修：

1. 这版去掉了 output checkpoint，但没有补新的“一致性快照”机制。现在 `source_last_order` 在复制前就抓取，而 `copy_required_cfs()` 是逐个 CF 迭代复制。在线源库继续写入时，不同 CF 看到的会是不同时间点的数据，`transaction` / `tx_sequence_info_mapping` / `transaction_acc_node` 之间可能错位。原来的 checkpoint 至少保证了 DB 级一致性，这版把这层保证去掉了，但命令没有强制停源库再跑。

2. 这版把 replay 的默认语义悄悄改成了“默认丢掉整类历史 CF”。`REQUIRED_REPLAY_COLUMN_FAMILIES` 只复制 8 个 CF，不包含 `state_change_set`、`event`、`event_handle`。fresh DB 从空库开始构建，所以即使用户没有开 `--history-prune`，输出库也会默认丢失 changeset/event 历史面。这是行为回归，至少应该是显式模式开关，不能作为默认行为。

3. 这条 PR 当前连自己新增的 `rooch-pruner` 测试都编不过。我本地执行：

`cargo test -p rooch-pruner incremental_replayer -- --nocapture`

直接失败在 `incremental_replayer.rs` 测试模块里未解析的 `accumulator` 引用（`use accumulator::accumulator_info::AccumulatorInfo;`）。至少在测试依赖修好之前，这条 PR 还不能合。

整体方向是对的，但这版还没到可合并状态。建议先补：
- 复制阶段的一致性边界
- “默认丢历史”改成显式模式
- 修掉测试编译问题

jolestar:
Updated this PR with the direction we agreed on in review/discussion.

Current replay semantics:
- `db state-prune replay` now builds a fresh output DB and defaults to windowed history output.
- The default lower bound is `from_order`. `history_retain_from` and `history_retain_window` still let us narrow that window, but there is no separate `--history-prune` mode anymore.
- The generic post-copy trim path is removed. We now construct the retained history window directly during copy, then only run the final metadata repair/rewrite pass.

Per-CF handling in the new path:
- Full copy: `transaction_acc_node`, `config_genesis`, `da_last_block_number`, `da_block_submit_state`, `proposer_last_block`
- Window copy `[retain_from..=to_order]`: `transaction`, `tx_sequence_info_mapping`, `transaction_execution_info`, `state_change_set`
- Rebuilt instead of copied: `state_node` from snapshot + replay, then `startup_info` and `sequencer_info`
- Skipped by default: `state_node_recycle`, `event`, `event_handle`

Why this replaces the earlier version:
- It matches the conclusion that replay should produce the minimal runnable DB by default, instead of silently mixing full-copy semantics with an optional history-prune phase.
- It avoids the bad intermediate state from the previous PR version where we copied a broader set first and then trimmed with leftover legacy logic.
- It also removes the stale `prune_history` implementation from the runtime path, which was the source of the `state_change_set` review concern.

Verification:
- `cargo test -p rooch-pruner incremental_replayer -- --nocapture`

This update is in commit `63a28a01b`.

